PR #38, opened 2025-04-11
fix: chg repo to adk-docs to help redirect users to repo for docs

polong-lin:
Users have been reporting documentation issues in adk-python, which is the repo for the SDK itself (not the documentation). This will make the docs page point to the docs repo (adk-repo) instead to so that users naturally report bugs directly in the docs repo.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -6,8 +6,8 @@ site_description: >-
 site_dir: site
 
 # Repository
-repo_name: adk-python
-repo_url: https://github.com/google/adk-python
+repo_name: adk-docs
+repo_url: https://github.com/google/adk-docs
 
 # Copyright
 copyright: Copyright Google 2025


polong-lin:
Please hold on merging this PR for now.

koverholt:
An alternative option to this might be to disable issues in the `adk-docs` repo, point users to file issues in `adk-python`, possibly add a docs issue template, then we can label and manage docs from that one repo, if it makes sense. But happy to discuss more and determine the best options for the community!

hangfei:
Agree. Let's hold this change.

I have updated adk Python repo template to redirect the traffic. I think the doc PR has gone gone.


I believe the community convention is stll to have the source code repo on the top right.

AshleyHenry15:
Bringing the docs into the SDK repo has many advantages, such as triaging (is this a docs-only issue or a feature refactor), a developer's ability to make a code contribution and update the docs in a single repo/PR if you have release notes—you can easily link back to the issues, and you can also auto-build code examples and snippets into the docs (less documentation maintenance). 

If you can hot release docs outside of "product releases," having the code and the docs under the same roof has many advantages.

lavinigam-gcp:
As these changes are not finalized and status quo is maintained, I am going to close this PR. If this changes in the future, we can raise a new PR to fix this. Closing the issue.

PR #31, opened 2025-04-10
fix: Added 'asynio.run()' alternative to using await to run conversations.

IAmNo1Special:
Added commented "asyncio.run()" to places where raw "await" is being used.
This fixes the "SyntaxError: 'await' outside function" error.

PR #46, opened 2025-04-11
Added asyncio.run method as alternative to using await to execute async conversations

IAmNo1Special:
Added commented "asyncio.run()" to places where raw "await" is being used.
This fixes the "SyntaxError: 'await' outside function" error.

Same as closed unmerged PR #31.

### DIFF
--- a/docs/get-started/tutorial.md
+++ b/docs/get-started/tutorial.md
@@ -365,6 +365,9 @@ async def run_conversation():
 
 # Execute the conversation using await in an async context (like Colab/Jupyter)
 await run_conversation()
+
+# Alternatively, if you are not using Jupyter/ Colab, execute the conversation using asyncio
+# asyncio.run(run_conversation())
 ```
 
 **Expected Output:**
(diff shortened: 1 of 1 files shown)


IAmNo1Special:
@lavinigam-gcp
Had to resubmit after fixing my own branches but this PR changes are the same as what was approved.

Sita04 (approved):
Thanks for contributing to the docs and making it better! @IAmNo1Special

polong-lin:
@lavinigam-gcp Did you want to double-check anything here because I think you're mirroring everything in `tutorial.md` with the Colab notebook therein? If not, feel free to merge.

lavinigam-gcp:
I have suggested few changes @IAmNo1Special  - I am making the same on the notebook. Kindly update the PR and let me know if you have questions. 

More importantly, while asyncio run in Colab, but when you try to call asyncio.run() inside a Colab/Jupyter cell, you are telling Python to create a new event loop while one is already running (the one started by Colab/Jupyter), so user will get error. 

So our suggestions should be to use await on Colab/Notebook and asyncio for .py, jsut to avoid DX issues.

PR #47, opened 2025-04-11
fix: update gemini 2.0 flash to latest ver for Live API

polong-lin:


kweinmeister (approved):

Sita04 (approved):

hangfei:
for most place, i suggest we use gemini-2.0-flash. Only for top-level readme or quickstart, we add a comment mention we should use the preview version for live streaming. And for live streaming tutuorial, we should chagne to the preview version for sure.

polong-lin:
Updated per @hangfei 's suggestion. @kazunori279 could you please scan through the changes and do a final check?

PR #54, opened 2025-04-12
Update CONTRIBUTING.md

DanielDekhtyar:
Added contribution guidelines to this repository.
I just copied it from [CONTRIBUTING.md](https://github.com/google/adk-python/blob/main/CONTRIBUTING.md) from [adk-python](https://github.com/google/adk-python/tree/main) repository

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,1 +1,33 @@
-# TODO
+# How to contribute
+
+We'd love to accept your patches and contributions to this project.
+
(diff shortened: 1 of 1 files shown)


koverholt (approved):
Thanks for adding that! We also added https://google.github.io/adk-docs/contributing-guide/ recently, but it's good to have this as well in the repository root, and we can sync them up in a future PR. Cheers! 🎉

PR #57, opened 2025-04-13
Update Agent instruction in quick-start example

functicons:
Use second-person (you) instead of first-person (I) in the agent instruction.


koverholt (approved):
LGTM, thanks for the improvement! 🎉

PR #69, opened 2025-04-14
fix: Update google/adk-docs repository link

xujiongda:
The original link was point to the site itself rather than the git repository.


polong-lin:
Thank you for fixing the link!

PR #76, opened 2025-04-14
Add local dev CLI & UI docs

### LINKED ISSUE #75: Add adk flags and env variables to docs
**Is your feature request related to a problem? Please describe.**

Unable to find CLI flags like `--trace_to_cloud` documented anywhere.

While this might be hard to keep up to date, we could include a few important ones and a link to get more from the source: 

https://github.com/google/adk-python/blob/577222509c0f7ff7bd1fa4317eb20fd871a37daa/src/google/adk/cli/cli_tools_click.py

zeroasterisk:
Add LiteLLM ref to quickstart

Add local dev CLI & UI docs, should fix #75

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![License](https://img.shields.io/badge/License-Apache_2.0-blue.svg)](LICENSE)
 
-<img src="docs/assets/agent-development-kit.png" alt="Agent Development Kit Logo" width="150">
+[<img src="docs/assets/agent-development-kit.png" alt="Agent Development Kit Logo" width="150">](https://github.com/google/adk-python)
 
 **An open-source, code-first Python toolkit for building, evaluating, and
 deploying sophisticated AI agents with flexibility and control.**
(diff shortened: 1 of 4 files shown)


koverholt:
Thanks for the contribution! I'd like to break up this PR into a few different components, and then we can discuss the scope of #75:

- README changes we can handle in a different PR
- We have a large number of pages in getting started, so I'd like to refactor this to fit into one of the existing pages such as the quickstart, tutorial, or other page
- As we bring online various paths for different development environments, we want to be very intentional that we are starting with devs from a known initial state (Colab, Codespaces, etc.), otherwise we risk supporting open-ended issues with Python packages, Python versions, different Python distros, OSes, etc. So we want to hold off on a page for "local dev" at the moment.
- Finally, for the scope of #75, let me check in with the ADK team to see if we can get better CLI discovery in the CLI tool (the `click` library has a lot of convenience features for this purpose, or perhaps we can look at `click` + `sphinx` docs that we use now for API reference docs) rather than trying to manually update and sync it in our own docs manually.

So I will close this PR for now, and let's continue discussing in #75!

PR #87, opened 2025-04-15
Update documentation for Application Integration Tools

salonijuneja21:
Fixes: https://github.com/google/adk-python/issues/32


koverholt (approved):
Thank you @salonijuneja21, this is a great contribution! And thanks for adding screenshots as well 🤩 . Everything LGTM.

IgorSiilva:
@salonijuneja21 @koverholt 

do you guys can help me with a question?


i'm trying to follow the doc to integrate with a bigquery connection, i already create the connection (first image) and setup on the integration (second image)

image
image

i'm stuck with this error Required parameter Execute Query Request is not present.

image
image


Scrolling down the integration config i see that has a field task input with custom query empty, i need to provide the query itself here? i was expecting that the agent could generate the query and run by itself


image

PR #89, opened 2025-04-15
Update tutorial.md

didier-durand:
Hi,
Just added a new line around line 40 (after </div>) to allow bold on **multi-agent** to work as well as 2 preceding links (which would not display as such) to work correctly.
Best,
Didier


koverholt (approved):
Awesome, thanks for the contribution! 🤘 Docs should be live in a minute.

didier-durand:
Hi, my pleasure! thanks for merging this PR.

PR #93, opened 2025-04-16
Update docs to add root agent instead of summary agent

adars124:
There was a mistake in the Agent-as-a-tool example in the documentation. There should be `root_agent` passed in the Runner instead of `summary_agent` which i fixed in this PR.


koverholt (approved):
Great catch, this was a tricky one! 😵 I reproduced and it was only interacting with the summary agent like you mentioned, but your fix did the trick. LGTM. And thanks for your contribution! ❤️

PR #98, opened 2025-04-16
Fixed instruction for root agent to avoid summarization before passing text to summary agent

AlexRAV:
It must fix this issue https://github.com/google/adk-python/issues/212

### DIFF
--- a/examples/python/snippets/tools/function-tools/summarizer.py
+++ b/examples/python/snippets/tools/function-tools/summarizer.py
@@ -18,7 +18,7 @@ summary_agent = Agent(
 root_agent = Agent(
     model='gemini-2.0-flash',
     name='root_agent',
-    instruction="""You are a helpful assistant. When the user provides a long text, use the 'summarize' tool to get a summary and then present it to the user.""",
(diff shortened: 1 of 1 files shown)


koverholt:
Thanks for the PR and pointing to the original issue! Just after https://github.com/google/adk-python/issues/212 was opened, I merged #93 here, which might fix the issue due to an error in how `root_agent` vs. `summary_agent` was specified.

If you try again with the latest docs at https://google.github.io/adk-docs/tools/function-tools/ with the changes from #93, do you still see the same behavior?

AlexRAV:
@koverholt hi! Yes, I've tested it with new, fixed version where root_agent is passed to runner.

AlexRAV:
Just to clarify, as my previous message may have been a bit confusing: the issue described in https://github.com/google/adk-python/issues/212 is reproducible in the latest version, where the root_agent is passed to the runner instead of the summary_agent.

After that, I updated the prompt, and now the root_agent correctly forwards the exact user message—without additional summarization—to the summary_agent.

koverholt:
Perfect, thanks for confirming. Y'all are super fast ⚡, and thanks for helping make ADK more awesome! 🏆

koverholt (approved):
LGTM, thanks for testing, and for your contribution!

PR #103, opened 2025-04-16
Typo in print output

LazaUK:
Typo in print output, as use uses "AGNET" instead of "AGENT"


polong-lin (approved):

polong-lin:
Great catch, thank you! @LazaUK

PR #107, opened 2025-04-16
Clarify documentation on multimodal model support

ianmckee-google:
Multimodal inputs require specific supportedGenerationMethods to be included in the model details, bidiGenerateContent for audio/video, generateContent for text.


kazunori279 (approved):
Confirmed that the `-exp` model is still available and works with this sample. We will revert to this model until the google-adk supports the newer models.

PR #114, opened 2025-04-17
docs: Fix init.py -> __init__.py

ftnext:
Thanks to great docs.

I first read https://google.github.io/adk-docs/get-started/quickstart/#__init__py ,
then read evaluation doc and I found that `init.py` is wired.

Refer to https://github.com/google/adk-python/blob/v0.1.0/src/google/adk/cli/cli_tools_click.py#L111

```
The path to the __init__.py file that contains a
```

### DIFF
--- a/docs/evaluate/index.md
+++ b/docs/evaluate/index.md
@@ -316,7 +316,7 @@ adk eval \
 
 Here are the details for each command line argument:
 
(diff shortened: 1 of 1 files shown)


koverholt (approved):
Good catch! You get the `Agentic Eagle Eye Award of the Day` 🏆 😁 . And thanks for contributing to ADK & open source! 🎉

PR #122, opened 2025-04-17
feat: add header-checker-lint.yml

msampathkumar:
Add config file for header-checker-lint.yml to limit checking for python files


koverholt (approved):
Thanks for the config! 🙏  Feel free to open an issue if this needs one or more follow up PRs.

PR #123, opened 2025-04-18
fix: Quickstart (streaming): fix the SSL error for audio and video, and add Vertex AI support.

kazunori279:
Adding Vertex AI support to Quickstart (streaming)


kazunori279:
Confirmed this fixes the SSL error for audio and video, and for both Gemini API Key and Vertex AI.

koverholt (approved):
Great stuff, thanks for the fix, testing, and getting the updates in the docs! ❤️  Happy streaming w/ Gemini! 🌊

PR #125, opened 2025-04-18
fix: variable name in custom agents documentation

MePriyanshuHoo:
Update the variable name in `docs/agents/custom-agents.md`.

* Change the variable name to `GEMINI_2_FLASH` on line 114 to maintain consistency with example's code


koverholt (approved):
Good catch. Thank you for helping us get things in order and consistent while we’re in the process of setting up CI/CD, docs test, etc. Cheers! 🙌

PR #127, opened 2025-04-18
added missing root agent intialization in example code

ghost:


lavinigam-gcp:
Hi @kevin-automation-architects Thank you for pointing out the issue. I have just merged a PR: https://github.com/google/adk-docs/pull/138

I have added instructions in the sample so that user would know that they need to change the main agent name to root_agent as per the interface they choose. Given, most users are going to use the samples within colab, we have prioritized keeping it with descriptive name. 

Let me know if this make sense.

PR #138, opened 2025-04-21
parallel-agent-code-fix

lavinigam-gcp:
— Adding detail instructions in the sample code to run the sample code with choice of interface. Users can now choose the method and follow instructions given in the sample .py file. 
— It clarifies that user would need to change the main agent name to `root_agent` if they are running it in `adk web`, `adk cli` or `adk api_server`. 
— Uses `InMemoryRunner` for notebooks/script.

### DIFF
--- a/docs/agents/workflow-agents/parallel-agents.md
+++ b/docs/agents/workflow-agents/parallel-agents.md
@@ -47,5 +47,8 @@ These research tasks are independent.  Using a `ParallelAgent` allows them to ru
 ???+ "Code"
 
     ```py
-    --8<-- "examples/python/snippets/agents/workflow-agents/parallel_agent_web_research.py"
+    --8<-- "examples/python/snippets/agents/workflow-agents/parallel_agent_web_research.py:init"
     ```
+**NOTE:** Runnable Code Sample:
+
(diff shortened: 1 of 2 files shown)


kazunori279 (approved):
approved!

ghost:
@lavinigam-gcp, thanks for merging my PR, It would also be appreciated and extremely useful if there was information on how to merge the results from the parallel agent's response. I have been trying to do it with another `LlmAgent` who is a 'merge agent', but the results are consistent, even if I heavily prompt engineer heavily. 

Apart from these docs there is no other examples from Google using the `ParallelAgent`, not even in the `adk-samples`, repo

Merging responses is a common use case, especially for the parallel agent. Even documentation for merge responses for loop agents would be useful to many people, as most AI Agents only need to output a single response, considering most AI Agents are conversational

lavinigam-gcp:
@kevin-automation-architects That is a great point Kevin. You are right, we don't have an example that shows how to merge the output-key or states from individual agents, such that they can be used in downstream. Would you mind opening an issue and assigning it to me? If not, I will add one tomorrow. 

I will work on the sample this week and update the docs. Thank you again, this is a great ask :)

kevinnkansah:
Done

PR #141, opened 2025-04-21
Feat: Update docs for using MCP Tools with "adk-web"

pwwpche:
Update introductions for adk web using MCP Tool. This should help with existing MCP issues like

* https://github.com/google/adk-python/issues/304
* https://github.com/google/adk-python/issues/233


docs/tools/mcp-tools.md
```diff
@@ -484,10 +360,111 @@ python3 ./mcp_agent/agent.py
 
 The script will print startup messages and then wait for an MCP client to connect via its standard input/output to your MCP Server in adk\_mcp\_server.py. Any MCP-compliant client (like Claude Desktop, or a custom client using the MCP libraries) can now connect to this process, discover the load\_web\_page tool, and invoke it. The server will print log messages indicating received requests and ADK tool execution. Refer to the [documentation](https://modelcontextprotocol.io/quickstart/server#core-mcp-concepts), to try it out with Claude Desktop.
 
+## Using MCP Tools in your own Agent implementation with Code
+
+This section is relevant to you if:
```
kanlanc, line 418: How is this different from the above examples?

Like what kind of situation would warrant this?

My brain is struggling to wrap around how could the third example be different from the second or the first one where the tools and agents are also async like the google maps api from my understanding is async

kanlanc:
Hey @pwwpche , 

First, thank you very much for this.

Everything made sense until the last couple of lines, and I added a comment to explain my confusion.

docs/tools/mcp-tools.md
pwwpche: The process for creating agent is the same.

However in `adk web`, exit stack is handled by the adk web fastapi runner itself - we don't have to manage it. If we are building your own agent app - by developing our application server and invokes the runner manually like in the example, we would need to manage the exit stack

Does that make sense to you?

kanlanc: I think I get it, if users are going to be making their own runner sessions then the third example makes sense but if using the default way of doing it which is just using the Agent class to export agents and not defining custom runner sessions and instead using adk web, then the first two examples are the way to go?

That was another thing I was wondering about when I was going through the docs, do I make my own runner session or by running adk web its handled, now its more clear.

Thank you! Let me know if my understanding makes sense.

kanlanc: Also, if we are using root_agent to export it from agent.py file, shouldnt the __init__ also import it as 

```python
from .agent import root_agent
```

kanlanc:
@pwwpche 

Also, if we are using root_agent to export it from agent.py file, shouldnt the __init__ also import it as 

```python
from .agent import root_agent
```


And secondly, doing it the way mentioned in the commit still runs me into co routine issues because its not being awaited.

<br>
Error

image

<br><br>
agent.py file

image

<br><br>

__init__ file

image

docs/tools/mcp-tools.md
```diff
@@ -43,6 +43,10 @@ The examples use the `MCPToolset` class in ADK which acts as the bridge to the M
 5. **Proxy Calls:** When the `LlmAgent` decides to use one of these tools, `MCPToolset` forwards the call (`call_tool` MCP method) to the MCP server and returns the result.  
 6. **Manage Connection:** Handle the lifecycle of the connection to the MCP server process, often requiring explicit cleanup.
 
+These examples assumes you interact with MCP Tools with `adk web`. If you are not using `adk web`, see "Using MCP Tools in your own Agent implementation with Code" section below.
+
+_Note: Using MCP tool requires a slightly different syntax to export the agent containing MCP Tools. A simpler interface for using MCP in ADK is currently WIP._
```
hangfei: expand WIP

docs/tools/mcp-tools.md
```diff
@@ -55,128 +59,59 @@ Create `agent.py` in `./adk_agent_samples/mcp_agent/` and use the following code
...
           args=["-y",    # Arguments for the command
-                "@modelcontextprotocol/server-filesystem",
-                # TODO: IMPORTANT! Change the path below to an ABSOLUTE path on your system.
-                "/path/to/your/folder"],
+            "@modelcontextprotocol/server-filesystem",
+            # TODO: IMPORTANT! Change the path below to an ABSOLUTE path on your system.
+            "/path/to/your/folder",
+          ],
       )
-      # For remote servers, you would use SseServerParams instead:
```
hangfei, line 84: could we add another example for this?

docs/tools/mcp-tools.md
```diff
@@ -55,128 +59,59 @@ Create `agent.py` in `./adk_agent_samples/mcp_agent/` and use the following code
...
 
-# --- Step 1: Import Tools from MCP Server ---
-async def get_tools_async():
-  """Gets tools from the File System MCP Server."""
-  print("Attempting to connect to MCP Filesystem server...")
+async def create_agent():
+  """Gets tools from MCP Server."""
   tools, exit_stack = await MCPToolset.from_server(
-      # Use StdioServerParameters for local process communication
       connection_params=StdioServerParameters(
```
hangfei, line 69: can we enhance the example so it works with two MCP tools.

docs/tools/mcp-tools.md
```diff
@@ -55,128 +59,59 @@ Create `agent.py` in `./adk_agent_samples/mcp_agent/` and use the following code
...
-
-# --- Step 3: Main Execution Logic ---
-async def async_main():
-  session_service = InMemorySessionService()
-  # Artifact service might not be needed for this example
-  artifacts_service = InMemoryArtifactService()
-
-  session = session_service.create_session(
-      state={}, app_name='mcp_filesystem_app', user_id='user_fs'
+  agent = LlmAgent(
```
hangfei, line 79: we should cover the following case: 
agent1(MCP1), agent2(MCP2)

(not sure how the exit_stack can work)

docs/tools/mcp-tools.md
```diff
@@ -484,10 +360,111 @@ python3 ./mcp_agent/agent.py
...
+  print("Cleanup complete.")
+
+if __name__ == '__main__':
+  try:
+    asyncio.run(async_main())
+  except Exception as e:
+    print(f"An error occurred: {e}")
+```
+
 ## MCP with adk web
```
hangfei: after above change, the following is not needed any more?

docs/tools/mcp-tools.md
pwwpche: Updated with the newest example - now it includes an agent with one local and another remote MCP Server

kanlanc:
This particular section looks good to me! 

A comment from me is Adk web is good for sure but before I became more familiar with adk the difference in examples of using a full main function with asyncio and some other examples just exposing sub agents and root agents confused me. I was stuck between the quick start and examples show the former but the full power of ask is displayed with the latter. (OPENAPI toolset is also misbehaving in the former but works in the latter)

In my opinion, it would be really cool to add a section for saying what exactly adk web abstracts from the developer, how to tailor the code for just sub and root agents architecture and a section for when to follow either architecture

kanlanc (approved):
Chaining multiple MCP servers together is a much needed feature and the explanations are spot on.

Thank you very much for this

hangfei:
@koverholt @polong-lin could you help review and merge? Thanks!

PR #135, opened 2025-04-20
fix(tutorial): add MODEL_GEMINI_2_5_PRO and fix agent model call

Izekc:
Add variable MODEL_GEMINI_2_5_PRO and fix the agent model value to MODEL_GEMINI_2_0_FLASH, so the tutorial can be runnable

Issue #118, opened 2025-04-17
Tutorial guide: call_agent_async update


nicastronaut:
**Describe the bug**

In the Tutorial guide, the `call_agent_async` function is defined in Step 1.4. 

Then, in Step 2, it says "We'll use the call_agent_async function (defined earlier, which now accepts runner, user_id, and session_id)", but does not clearly show the updated code or instruct the user to update that function anywhere and how those arguments are handled.

During Step 3.4. (the agent team), the function receives unexpected arguments. I skipped the multi-modal step as I was going through it, but I suspect the same thing would happen.

If you remove those arguments from the function calls in Step 3.4., the agent conversation runs, but it does not specify calling the sub-agent tools as expected:
```
>>> User Query: Hello there!
<<< Agent Response: Hi! How can I help you?


>>> User Query: What is the weather in New York?
--- Tool: get_weather called for city: New York ---
<<< Agent Response: The weather in New York is sunny with a temperature of 25°C.


>>> User Query: Thanks, bye!
<<< Agent Response: You're welcome! Goodbye!
```

**To Reproduce**

1. Follow the code exactly as it is written in the Tutorial guide (copy paste for posterity rather than the premade colab)
2. Get to either the multi-model step, or skip to Step 3.4.

**Expected behavior**

I expect the output would specify the sub-agents being called due to the proper handling of the runner/session with the `call_agent_async` function, and thus "confirms successful automatic delegation"

**Versions**

 - OS: Linux
 - ADK version: 0.1.0
 - Python version: 3.13.3

lavinigam-gcp:
Hi @nicastronaut Thank you for raising the bug, and you were absolutely right. There seems to be some mismatch between the docs page and the notebook. I have updated them now. They both reflect the same text and code: 

https://google.github.io/adk-docs/tutorials/agent-team/

Pondering more on your raised issue, I also added a separate folder for anyone to run this tutorial using `adk web` : 

https://github.com/google/adk-docs/tree/main/examples/python/tutorial/agent_team/adk-tutorial

This means, you don't have to necessarily follow notebook and can run the same example with different steps using other interfaces of ADK: web, CLI and api_server. 

Let us know if this helps. Happy to addresses any further issue you are facing.

lavinigam-gcp:
This is fixed in #334. Closing the issue. If there's still something that's not working, please raise a new issue and thank you again! :)

PR #143, opened 2025-04-22
fix: tutorial section overhaul and bug fixes

lavinigam-gcp:
- Tutorial section is moved out of getting started and is a seperate section
- Fixed a sentence in getting-started page that still showed “google search”
- Given we will have more tutorials as part of docs, now we have sub-sections with the tutorial name
- For weather agent team - fixed issues pointed out in #118 #135
    - It now includes code for asynio in the notebook
    - It also now has .py file to run it through adk web
    - Step 2 is now optional and not connected to other steps. So user can follow other steps if they choose to
    - There is disclaimer at the top - for different method to run the code
    - The text on the page is now synced with notebook
    - Collab and social link are updated with new changes in folder.


koverholt (changes requested):
This is great! Thanks for the overhaul!

Be sure to add redirects like https://github.com/google/adk-docs/blob/main/mkdocs.yml#L91 so that the existing tutorial doesn't result in a 404.

Otherwise LGTM!

kazunori279 (approved):

koverholt (commented):
Thanks for the updates, all LGTM!

PR #144, opened 2025-04-22
Add note on policy when using the built-in Google Search tool

polong-lin:


docs/tools/built-in-tools.md
```diff
@@ -23,6 +23,10 @@ The `google_search` tool allows the agent to perform web searches using Google
 Search. It is compatible with Gemini 2 models, and you can add this tool to the
 agent's tools list.
 
+!!! note "Additional requirements when using the `google_search` tool"
```
lavinigam-gcp: Can we also add a note here that currently `google_search` is only supported through Gemini models and not other models (through LiteLLM). This was pointed out by a user on Reddit: https://www.reddit.com/r/agentdevelopmentkit/comments/1k2wgrt/comment/moe86yx/?context=3

polong-lin: Good call out. Just fixed this.

lavinigam-gcp (approved):

koverholt (approved):
LGTM, thanks!

PR #146, opened 2025-04-22
Remove reference to  pip install google-adk[database]

wietsevenema:
sqlalchemy is already installed and this package does not exist per https://github.com/google/adk-python/pull/43

Fixes https://github.com/google/adk-python/issues/25


koverholt (approved):
Thanks for the catch and for your contribution. LGTM! ❤️

PR #162, opened 2025-04-24
Update small temp error in sample agent.py

alokpattani:


moficodes (approved):

koverholt (approved):
This is the greatest two-character PR ever in ADK history! Thank you, Alok. ❤️ 😂 

It's interesting that the agent, or LLM, when it was doing this calculation to convert °C to °F, it used 25 °C times 1.8, which is 45 °F, which is close to 41 °F. And it forgot the last step was to add 32 (or some hallucination like that perhaps), which is the same reason that I got a B on one of my engineering exams in 2006. And that is when I learned the difference between converting temperature **differences** with `F = C*(9/5)` and converting temperature **values** with `F = C*(9/5)+32` So, well done, Alok, you get the `math + stats` ADK contribution award of the day! 🏆 

Thank you for teaching our agents and docs that we need humans in the loop to make sure we don't make silly engineering or math calculations like that one when using language models to do math operations. Onward and forward with agents! 😁

PR #168, opened 2025-04-25
fix: add missing .txt within "Install dependencies" step

hsuyuming:
fix: add missing .txt within "Install dependencies" step


koverholt (approved):
Oops, great catch! 🕵️  Thanks for fixing this typo and helping others avoid this error! 🏆

Issue #169, opened 2025-04-25
Missing refer link in Model Context Protocol Tools docs section.


keli-wen:
**Describe the bug**

In docs section [Model Context Protocol Tools#Prerequisites](https://google.github.io/adk-docs/tools/mcp-tools/#prerequisites). There is a miss here, the link that needs to be pointed to is missing.

There is an additional issue: the documentation for MCP Tools lacks a reminder about the placement of the `.env` file. It might be necessary to add supplementary information to avoid some issue like:  https://github.com/google/adk-python/issues/305

Refer to [code](https://github.com/google/adk-docs/blob/main/docs/tools/mcp-tools.md?plain=1#L67-L69)

```python
# Load environment variables from .env file in the parent directory
# Place this near the top, before using env vars like API keys
load_dotenv('../.env')
```

**To Reproduce**

Skip

**Expected behavior**

- Complete the missing link.
- Add a reminder like **Important:** Place the `.env` file in the parent directory of `adk_agent_samples`.

**Screenshots**

<img width="553" alt="Image" src="https://github.com/user-attachments/assets/e97bc6ba-2b43-486f-8f70-a15da9537fa3" />

**Versions**

 - OS: MacOS
 - ADK version: `0.3.0`
 - Python version: `3.11`

**Additional context**

tpryan:
Issue seems to be resolved.

PR #170, opened 2025-04-25
fix: update ADK setup link and add .env file location note in MCPTools.

keli-wen:
### Description

Refer to issue #169 

Fix the issue according to the method mentioned in the issue.


lavinigam-gcp:
@keli-wen Thank you for raising this. Due to merge conflicts, this is addressed in #342 and will be merged.

PR #178, opened 2025-04-25
fix: Instruction Templating Documentation

matt-landers:
Added additional documentation for for LLM instructions. 

The instruction is a template string where the variables are replaced from state. This is in some of the examples, but it was missing in the documentation.


koverholt (approved):
This is great! Thanks for your contribution! 🚀 And great catch that the template syntax was used elsewhere but not defined here! I made a few formatting fixes (unrelated to your changes!).

If you have a chance in a followup PR, consider adding a few examples in the code block that demonstrate some of all of these templated vars. If not, no worries! Thanks again! 🙌

PR #179, opened 2025-04-25
Add Windows Encoding Note for LiteLLM

designcomputer:
When using ADK agents with `LiteLlm` on Windows, users might encounter a `UnicodeDecodeError: 'charmap' codec can't decode byte...` error. This often happens when `litellm` (used by `LiteLlm`) tries to read cached files (like model pricing information) using the default Windows encoding (`cp1252`) instead of UTF-8.

**Suggested Improvement:**

Add a note to the documentation, specifically on the Models page or where `LiteLlm` is discussed, recommending that Windows users set the `PYTHONUTF8` environment variable to `1`. This forces Python to use UTF-8 globally and can prevent this decoding error.

**Example (PowerShell):**

```powershell
# Set for current session
$env:PYTHONUTF8 = "1"

# Set persistently for the user
[System.Environment]::SetEnvironmentVariable('PYTHONUTF8', '1', [System.EnvironmentVariableTarget]::User)

### DIFF
--- a/docs/agents/models.md
+++ b/docs/agents/models.md
@@ -176,6 +176,23 @@ layer, providing a standardized, OpenAI-compatible interface to over 100+ LLMs.
         )
         ```
 
+## Avoiding LiteLLM UnicodeDecodeError on Windows
+
+When using ADK agents with LiteLlm on Windows, users might encounter the following error:
+```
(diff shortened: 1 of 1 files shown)


lavinigam-gcp:
@designcomputer Thank you for raising this. Due to merge conflicts, this is addressed in #342 and will be merged.

PR #181, opened 2025-04-26
docs: Update quickstart.md

wyf7107:
expose the option session_db_url for custom session db store

### DIFF
--- a/docs/get-started/quickstart.md
+++ b/docs/get-started/quickstart.md
@@ -155,6 +155,11 @@ There are multiple ways to interact with your agent:
     ```shell
     adk web
     ```
+    !!!note "User custom database url for session service"
+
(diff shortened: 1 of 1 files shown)


koverholt:
Thanks for the PR! ❤️ 

We aim to keep the information in the quickstart as minimal as possible for the sake of repeatability for new users. The note about a custom database URL is too complex to add in to the quickstart. The location of the note is a bit out of place in the `Run Your Agent` section and might cause confusion for users just trying to run the simple/sample agent. We do want to point users to the right place to extend their agents and add components such as a persistent DB.

For those reasons and to keep the quickstart page simple, I'll close this PR for now. Feel free to open an issue with more details of the underlying problem(s) you are trying to solve for. Thanks again!

PR #184, opened 2025-04-26
fix: update authentication doc to correct some error and keep it consistent with latest codes

seanzhou1023:


seanzhou1023:
@koverholt @lavinigam-gcp

docs/tools/authentication.md
```diff
@@ -174,25 +173,40 @@ from google.adk.tools.google_api_tool import calendar_tool_set
...
+Below sections focus on the first case and you should be able to map it to the second case very straightforward. We will also describe some differences to handle for the second case if necessary.
 
 Here's the step-by-step process for your client application:
 
 **Step 1: Run Agent & Detect Auth Request**
 
 * Initiate the agent interaction using `runner.run_async`.  
 * Iterate through the yielded events.  
-* Look for a specific event where the agent calls the special function `adk_request_credential`. This event signals that user interaction is needed. Use helper functions to identify this event and extract necessary information.
+* Look for a specific function call event whose function call has a special name: `adk_request_credential`. This event signals that user interaction is needed. You can use helper functions to identify this event and extract necessary information. (For the second case, the logic is similar. You deserialize the event from the http response
```
Sita04:
```suggestion
* Look for a specific function call event whose function call has a special name: `adk_request_credential`. This event signals that user interaction is needed. You can use helper functions to identify this event and extract necessary information. (For the second case, the logic is similar. You deserialize the event from the http response).
```

PR #190, opened 2025-04-29
fix: the ParallelAgent execute run_async instead of run() 

hsuyuming:
fix: the ParallelAgent execute run_async instead of run() base on this line [1].
[1]https://github.com/google/adk-python/blob/main/src/google/adk/agents/parallel_agent.py#L94


koverholt (approved):
Nice detective work, thanks for including the link to the ADK source. Thanks for the catch and for your contribution! 😍 🎉

PR #195, opened 2025-04-30
feat: add MCP section

kazunori279:
Adding new MCP section. Is this the right way to add new section?

### DIFF
--- a/docs/mcp/index.md
+++ b/docs/mcp/index.md
@@ -14,6 +14,6 @@ MCP follows a client-server architecture, defining how **data** (resources), **i
 
     This guide walks you through two ways of integrating Model Context Protocol (MCP) with ADK.
 
-    [:octicons-arrow-right-24: More information](installation.md)
+    [:octicons-arrow-right-24: More information](../tools/mcp-tools.md)
 
 </div>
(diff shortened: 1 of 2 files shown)


koverholt (approved):
Yes, everything here LGTM. Thanks for your contributions and setting this up! ❤️

koverholt:
My only question is about moving MCP out of the Tools section. Do you think users will also expect to see it in the Tools section? If so, it's possible to add the navigation entry to the `mkdocs.yml` file both at the top level AND under the Tools menu. But only if it makes sense! 😁 

Do you have a preference between 1) top-level nav link only for MCP, 2) top-level nav link AND link under Tools?

koverholt (changes requested):
One more thing is that whenever you are moving an existing page, you should add a redirect from the old page path to the new page path in `mkdocs.yml`, as in:

https://github.com/google/adk-docs/blob/3794147f77e8a2303d52f394335b177aaaae17cc/mkdocs.yml#L86-L93

koverholt (approved):
Discussed more with @kazunori279, and this MCP page is a NEW top-level nav page. And the MCP tools page will remain in place. Sorry for the confusion! 😁 

So no need for duplicate links or redirects. Thanks again! ❤️

PR #215, opened 2025-05-04
doc: add streaming tools guide.

hangfei:
add streaming tool guide


koverholt (approved):
Thanks @hangfei for the writeup! The content LGTM. I see a few syntax errors, but I'd like to merge this for now to get it into the hands of users! Then we can fix those issues iteratively in future PRs and CI tests. ❤️

PR #218, opened 2025-05-05
chore: update google-cloud-tools.md

jackwotherspoon:
Clearing up that the tools for `ApplicationIntegrationToolset` should be added
to `tools.py` as this is not explicitly mentioned.

Can only tell this is where to put them by the import in the subsequent
section having the import as `from .tools import ...`

Also fixing rendering of `__init__.py`

image

### DIFF
--- a/docs/tools/google-cloud-tools.md
+++ b/docs/tools/google-cloud-tools.md
@@ -115,7 +115,7 @@ you only need to follow a subset of these steps.
     )
     ```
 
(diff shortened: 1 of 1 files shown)


koverholt (approved):
Awesome, thanks for contributing these fixes! ❤️ Really appreciate your attention to detail, we have so many items like this on the TODO list since launch, so everyone in OSS-land like you @jackwotherspoon is helping us faster than we can! Thank you! 😁

PR #219, opened 2025-05-05
fix: google_search import issue

kavinkumar807:
This PR fixes #216 

fixed the google search import from `built_in_google_search` to  `google_search`

PR #226, opened 2025-05-07
fix: Fixed the google search tool name and model name

kazunori279:
- With 0.4.0, the tool name has changed to google_search
- As of today, gemini-2.0-flash-exp is the workable model with Streaming for ADK that is soon to be replaced with gemini-2.0-flash-live-001.


koverholt (approved):
Thanks for the PR and for your contribution! ❤️ I was catching up with the releases, and the `google_search` tool is fixed as of #219. And thanks for the info on the streaming model version string!

PR #229, opened 2025-05-07
fix: fix long running tool docs

seanzhou1023:


seanzhou1023:
@polong-lin @koverholt

koverholt (approved):
This is fantastic, thank you for contributing! As a first pass, all docs and code samples LGTM, will merge this now based on user impact, and we can continue testing, updating, and iterating in future PRs. Thanks a bunch! ❤️

koverholt:
Thanks for the PR! If you happen to know an issue in adk-docs or adk-python to mention in a comment here, please do. If not, no worries!

dincal:
The newly uploaded sample is also incorrect. 

- It seems like ticket_id should be changed to ticket-id.
- Also, if the last event is not a function response (which I observed when I ran it), long_running_function_response will likely be None. The code related to this also needs to be revised.

PR #236, opened 2025-05-09
feat: Add shop agent notebook

kazunori279:
This is a supplemental notebook sample for upcoming video created with Sita.


Sita04 (approved):

XinyueZ:
@kazunori279 When should I use AgentTool, a pure tool, or sub-agents? 
In theory, these three can replace each other.

kazunori279:
Hi @XinyueZ , thanks for the great question! A pure tool doesn't have any reasoning capability itself unless you explicitly use Gemini API/else inside it. With sub-agents, each agent has responsibility to have smooth communication with the user so you need to write their instructions for it. In this sample I used AgentTool since I wanted to have the Shop Agent to take the responsibility for all user communication, whereas the Research Agent can just focus on the research. Does this answer to your question?

XinyueZ:
@kazunori279 nice, I get it.
The next question is, why don't we use Supervisor to control both as sub-agents?
I don't think there's anything wrong with the current implementation; I just feel that the step of having users input "yes" should be decided by the supervisor in the middle.

kazunori279:
@XinyueZ ah I just haven't tried Supervisor yet. This notebook is intended to be the minimum code to prove the concept (using Google Search for a smart recommendation), but in production it sounds like a great design pattern for the best control on the UX. Thanks for the advice.

XinyueZ:
@kazunori279 We are still looking for the source code of the audio demo you showed before on YouTube.

kazunori279:
For now I put higher priority to publish shorter snippets like this notebook and [the custom audio streaming sample](https://google.github.io/adk-docs/streaming/custom-streaming/) based on the code and knowledge I got by building the demo, rather than publishing its 3K lines of code without much explanations of the basics :) But I hope to publish a whole demo code in the future after cleaning it up!

XinyueZ:
3k lines, boom @kazunori279 
rock

PR #239, opened 2025-05-11
docs(examples): update commands and links in ADK tutorial README

hnliuzesen:
Correct `adk run` and `adk api_server` commands by removing script-specific references. 
The old command will return error via latest google-adk.

Adjust documentation link titles and URLs to point to updated sections for clarity and accuracy. 
The old pages are not found.


koverholt (approved):
Thanks for your contribution! ❤️ And thanks for catching these errors in the README, outside of the main docs build!

PR #240, opened 2025-05-11
fix: fix hunman_in_the_loop example

RobertIndie:
### Description

The `human_in_the_loop.py` script was failing with the following error:

```
ValueError: Session not found: session1234
```

The issue was caused by incorrect keys for the ticket ID and user ID.


koverholt (approved):
Oof, sorry you ran into those issues, but thank you 💯 x for your PR and contributions to fix it for the next user(s)! Appreciate the fixes and the detailed notes in the PR! This is a great tip for folks prepping their ADK agents to deploy to Cloud Run!

PR #245, opened 2025-05-12
Clarify multi-agent project structure for Cloud Run deployment

triangle-z:
* This update clarifies how to define and deploy multiple agents using Cloud Run.  
* It specifies that each agent should be placed in its own folder at the root level, with each folder containing a `root_agent`.  
* It also shows an example structure, in complement of the one in `Project Structure`.

Here's how it looks :
image


lavinigam-gcp (approved):
LGTM, merging

PR #249, opened 2025-05-12
fix: fix the image path

kazunori279:
fix the image path


koverholt (approved):
Thank you for the fixes and followup PR! ❤️

PR #254, opened 2025-05-13
example-addition: Added some examples to better understand different …

ygicp:
…part of session object's identification.

In the session object identification description, it was unclear (for first time readers), what exactly is id, and app_name and how they are different. Added some examples according to my own understanding. Please change accordingly if you feel so. Thanks.


lavinigam-gcp:
@ygicp Thank you for raising this. Due to merge conflicts, this is addressed in #342 and will be merged.

PR #286, opened 2025-05-20
fix: typo in built-in-tools.md

jsmeredith:


lavinigam-gcp (approved):
Thank you for catching and fixing it :) LGTM

PR #288, opened 2025-05-21
fix: gcloud deploy cloudrun

keithrozario:
Fixed the sample code in `main.py`

* changed `app: FastAPI = get_fast_api_app(` to `app = get_fast_api_app(`
* removed unnecessary import of FastAPI

I have tested this and it works.


lavinigam-gcp (approved):
LGTM. Thank you for fixing it.

PR #296, opened 2025-05-22
docs: update Python artifact docs to use async/await

jackwotherspoon:
As of ADK 1.0.0, the artifact service methods are now `async` and should use `await`

[ADK Python artifact service code showing async](https://github.com/google/adk-python/blob/a2263b18083fccd176b3ed61c368a65d9b7efde4/src/google/adk/artifacts/in_memory_artifact_service.py#L83)

## ADK 0.5.0
```python
# synchronous load artifact from ToolContext
artifact = tool_context.load_artifact(filename=filename)
```

## ADK 1.0.0 🎉 
```python
# async load artifact from ToolContext
artifact = await tool_context.load_artifact(filename=filename)
```

### DIFF
--- a/docs/artifacts/index.md
+++ b/docs/artifacts/index.md
@@ -358,7 +358,7 @@ The artifact interaction methods are available directly on instances of `Callbac
             filename = "generated_report.pdf"
 
             try:
-                version = context.save_artifact(filename=filename, artifact=report_artifact)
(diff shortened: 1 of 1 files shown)


jackwotherspoon:
@koverholt mind taking a look when you have a second 😄

koverholt (approved):
Thanks for the contribution! LGTM. Appreciate you helping keep things up to date for the 1.0.0 release! 😁

PR #301, opened 2025-05-24
Fix: await executing get_session

mochizuki875:
[get_session()](https://github.com/google/adk-python/blob/a380d70ac74bd458d283c5664527162e51e9807e/src/google/adk/sessions/in_memory_session_service.py#L109) is defined as async func, so I think we need to `await`.
Otherwise, it will return a coroutine.

```bash
✅ New InMemorySessionService created for state demonstration.
✅ Session 'session_state_demo_001' created for user 'user_state_demo'.

--- Initial Session State ---
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[1], line 35
     33 print("\n--- Initial Session State ---")
     34 if retrieved_session:
---> 35     print(retrieved_session.state)
     36 else:
     37     print("Error: Could not retrieve session.")

AttributeError: 'coroutine' object has no attribute 'state'
```

### DIFF
--- a/docs/tutorials/agent-team.md
+++ b/docs/tutorials/agent-team.md
@@ -981,7 +981,7 @@ session_stateful = await session_service_stateful.create_session(
 print(f"✅ Session '{SESSION_ID_STATEFUL}' created for user '{USER_ID_STATEFUL}'.")
 
 # Verify the initial state was set correctly
-retrieved_session = session_service_stateful.get_session(app_name=APP_NAME,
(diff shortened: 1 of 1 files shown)


jonatanblue:
This PR fixes https://github.com/google/adk-docs/issues/315

lavinigam-gcp:
@mochizuki875 Thank you for raising the PR. Really appreciate your effort. The changes have been merged here: https://github.com/google/adk-docs/pull/334. Closing the PR

PR #307, opened 2025-05-25
fix: remove invalid as_dict() usage from final session state logging

UtkarshVijayvergia:
This addresses a bug in the session inspection logic where `final_session.state.as_dict()` was being called. Since `final_session.state` is already a Python `dict`, calling `.as_dict()` results in:
```
AttributeError: 'dict' object has no attribute 'as_dict'
```

Changes Made:
- Removed the .as_dict() call and printed the dictionary directly using print(final_session.state).
- Added a comment `For detailed view`

Impact:
- Fixes the runtime error and ensures correct behavior when inspecting session state.


lavinigam-gcp:
@UtkarshVijayvergia Thank you for fixing this. Merging.

PR #345, opened 2025-05-30
fix: remove .as_dict() usage from final session state logging

UtkarshVijayvergia:
This addresses a bug in the session inspection logic where `final_session.state.as_dict()` was being called. Since `final_session.state` is already a Python dictionary, calling `.as_dict()` results in:
```
AttributeError: 'dict' object has no attribute 'as_dict'
```

Changes Made:
- Removed the `.as_dict()` call.

Impact:
- Fixes the runtime error and ensures correct behavior when inspecting session state.


lavinigam-gcp:
@UtkarshVijayvergia Thanks for the PR Utkarsh, merging!

PR #356, opened 2025-05-31
fix: removed `Await` keyword for session management using `InMemorySessionService()`

UtkarshVijayvergia:
This addresses a bug in the session creation and fetching logic where `InMemorySessionService()` is used. Since `InMemorySessionService().create_session(...)` and `InMemorySessionService().get_session(...)` is not asynchronous and using `await` keyword gives:
```
TypeError: object Session can't be used in 'await' expression
```

Changes Made:
- Removed the `await` keyword.

Impact:
- Fixes the TypeError and ensures correct behavior when inspecting session state.

### DIFF
--- a/docs/tutorials/agent-team.md
+++ b/docs/tutorials/agent-team.md
@@ -295,7 +295,7 @@ USER_ID = "user_1"
 SESSION_ID = "session_001" # Using a fixed ID for simplicity
 
 # Create the specific session where the conversation will happen
-session = await session_service.create_session(
+session = session_service.create_session(
     app_name=APP_NAME,
     user_id=USER_ID,
     session_id=SESSION_ID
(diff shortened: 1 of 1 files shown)


lavinigam-gcp:
@UtkarshVijayvergia Hi Utkarsh, are you using adk=<1.0.0? You will get this error in the older version. The tutorial is updated to work with >=1.0.0 where a lot of methods are async by default. 

I just tested the notebook, and it works with the latest version.

UtkarshVijayvergia:
I just checked and I am using ADK version 1.0.0. I will update the version. Thank you!

lavinigam-gcp:
@UtkarshVijayvergia Thanks Utkarsh for checking. But, we should mention this at the top of the tutorial the version it works, so that user knows and doesn't run into issue. Thank you for flagging the friction. I am happy to add that line, or if you want to take that PR? Either way its fine.

UtkarshVijayvergia:
@lavinigam-gcp Sounds good, I will go ahead and add the version at the top.

lavinigam-gcp:
@UtkarshVijayvergia I am closing this PR. you can raise the new one. Thanks again!

PR #357, opened 2025-06-01
Add internal documentation link to "Output Key" section of LLM Agent docs

Rohit10jr:
Updated the Markdown link to point directly to the relevant section (output_key) within the LLM Agent documentation, instead of linking to the top of the page.


lavinigam-gcp:
@Rohit10jr Thank you, Rohit, for fixing it. Approved and merged! :)

PR #358, opened 2025-06-01
Fix import path for CallbackContext in context section

Virtuoso633:
Updated the import path of CallbackContext in the Python example under the "CallbackContext" section. This reflects the actual module path: from google.adk.agents.callback_context.


lavinigam-gcp:
@Virtuoso633 Thank you for fixing it. Approved and merged! :)

PR #362, opened 2025-06-02
when interrupted the audio should stop playing

yunus:
When there is an interruption, audio keeps playing. To prevent that I added a short code to the app.js. Whenever an interrupt message is received it tell audio player to stop.


kazunori279 (approved):
Awesome, I wanted to add this. Thanks for the PR! :)

PR #369, opened 2025-06-04
docs: specify ADK version requirement for tutorial compatibility

UtkarshVijayvergia:
This PR adds a note at the beginning of the tutorial documentation to clarify that the tutorial requires `adk >= 1.0.0`. 

This helps prevent compatibility issues, as older versions may not support some async methods used in the examples.


lavinigam-gcp:
@UtkarshVijayvergia Thank for fixing this. Merging this! :)

PR #380, opened 2025-06-06
fix: Add missing name tag to schema annotations.

shukladivyansh:
Without this tag, we get errors like https://github.com/google/adk-java/issues/152

### DIFF
--- a/examples/java/cloud-run/src/main/java/agents/multitool/MultiToolAgent.java
+++ b/examples/java/cloud-run/src/main/java/agents/multitool/MultiToolAgent.java
@@ -43,7 +43,7 @@ public class MultiToolAgent {
     }
 
     public static Map<String, String> getCurrentTime(
-        @Schema(description = "The name of the city for which to retrieve the current time")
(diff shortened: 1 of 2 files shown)


vorburger (changes requested):
I disagree with this. This repetiion should not be required; see https://github.com/google/adk-java/issues/152#issuecomment-2950669586.

shukladivyansh:
cc @Sita04

glaforge:
Currently, you have to 1) either provide a `name` attribute, or indeed 2) use `-parameter`.

I didn't check, but if `name` is always required, regardless of whether the user specifies `-parameter` or not, then I would think it's a bug. 

IMHO the logic should be: if the `name` is specified, use it (as you may want to override the actual parameter name), or else if the name can be found by reflection (because `-parameter` is set) then use that, otherwise throw an error.

glaforge:
Historical notes: A bit before the release of 0.1.0, no error was thrown when the name attribute wasn't specified. But if `-parameter` wasn't set, a generic name like `arg0` was used instead.

joefernandez:
Closing this as obsolete.

Issue #152, opened 2025-04-23
Friction log for "Deploy to GKE" page and steps


koverholt:
Feedback from GKE team:

- Sample agent
  - Sample "capital agent" is confusing, users are using weather agent or quickstart agent
  - Suggest combining project structure and code files sections
- Deploying and testing
  - Need to fix root agent import, no local testing, only aware after
  - Had to add three roles before the Cloud Build command worked
- Testing the agent
  - Unable to see debug output after testing the agent deployed
  - Errors, permission denied

More details in a friction log document.

cc @moficodes, if you are interested in updates to the GKE docs based on some amazing feedback and friction log from two great folks on the GKE team: @tomergee and @syangx39 😄.

koverholt:
Thanks for taking a look at these updates @moficodes, appreciate the collaboration and your contributions! ❤

PR #381, opened 2025-06-06
fix: Add -parameters compiler arg in example POMs to avoid adk-java issue #152

vorburger:
See https://github.com/google/adk-java/issues/152.

This is a "counter proposal" to #380.


Poggecci (approved):
+1 to this over @Schema for a quickstart

PR #391, opened 2025-06-09
fix: integration-connector code sample unformatted

nag763:
Hey

Little issue that can be easily solved

So far :

image

As an extra space has somehow been added prior code block

Which is solved by removing that extra space

Reg,

LL


koverholt (approved):
Thanks for the fix! ❤️

koverholt:
Looks like this was fixed in #390. Thanks again for your contribution, looking forward to many more! 🙏

PR #392, opened 2025-06-10
fix: upgraded to 1.2.1, and added a troubleshooting tip

kazunori279:
fix: upgraded to 1.2.1, and added a troubleshooting tip


lavinigam-gcp (approved):
LGTM

PR #411, opened 2025-06-16
doc: Update quickstart-streaming.md

hangfei:
doc: remove the model name as it changes very often and it's causing confusion. instead, point them to the right source so they can figure it out.


hangfei:
@kazunori279 @polong-lin could you help review this change? Thanks.

kazunori279 (approved):

PR #414, opened 2025-06-17
fix: capital agent example asyncio run

keithrozario:
Resolved `SyntaxError` when trying to run the agent, by replacing `await main()` with `asyncio.run(main())`.

This allows the code snippet to run be run directly by the user. Otherwise the user will get the following error:

_SyntaxError: 'await' outside function_


joefernandez:
Fixed by: https://github.com/google/adk-docs/pull/528

PR #425, opened 2025-06-18
feat: Add link checker action

koverholt:
Adds a link checker to verify external links. Note that internal links are already checked as part of the mkdocs build.

Uses:

- https://github.com/byrnereese/linkchecker-mkdocs
- https://github.com/marketplace/actions/mkdocs-link-check


koverholt:
Will merge now and perform followup testing in subsequent PRs.

PR #441, opened 2025-06-23
fix: Add missing "name=" to @Schema parameter annotation to eliminate error message

tucosh:
The @Schema parameter annotation is missing the parameter name, leading to the following error message:

[agents.multitool.MultiToolAgent.main()] ERROR com.google.adk.tools.FunctionTool -- Functions used in tools must have their parameters annotated with @Schema or at least the code must be compiled with the -parameters flag as a fallback. Your function tool will likely not work as expected and exit at runtime.


Sita04 (approved):
Thanks for identifying the issue and sending the fix!

PR #450, opened 2025-06-25
fix: change call_agent to call_agent_async

guru-desh:
There's a bug in pages that use `call_agent` where this code exists:

```
session_service.get_session(app_name=APP_NAME,
                                                  user_id=USER_ID,
                                                  session_id=SESSION_ID)
```

The problem with this code is that this needs to be awaited, but it is not awaited, which causes the code to break.

To fix these code snippets, the following was done:
1. Create a function called `setup_runner_and_session`, which properly setups the session and runner.
2. Change `call_agent` to `call_agent_async`.

I have two questions before submitting this PR:
1. Are the `llms-txt` automatically updated or am I expected to update them based on these changes?
2. How can I automatically run all these samples to make sure that they pass? Is it part of the CI pipeline?
3. How can I fix the "invalid or missing license headers" pipeline?


lavinigam-gcp:
@guru-desh Thank you for raising the PR. I have fixed few more errors while review. Merging :)

lavinigam-gcp (approved):

PR #454, opened 2025-06-26
feat: add info for Memory Bank service

mmmarklu:
Do not merge until Vertex AI Memory Bank service has been launched in public preview, otherwise the links will 404.

feat: add new MemoryService implementation for Vertex AI Memory Bank as per request from the Agent Engine product team.

### DIFF
--- a/docs/sessions/memory.md
+++ b/docs/sessions/memory.md
@@ -32,7 +32,13 @@ ADK provides different ways to implement this long-term knowledge store:
     memory_service = InMemoryMemoryService()
     ```
 
-2. **`VertexAiRagMemoryService`**  
+2.  **`VertexAiMemoryBankService`**
+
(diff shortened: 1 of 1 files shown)


lavinigam-gcp:
@mmmarklu Hey Mark, do we know the timeline for this to be launched? I can see that the integration is already live in the [ADK v1.5.0](https://github.com/google/adk-python/releases/tag/v1.5.0). 

Also, do you think we should add the python quick-start directly in the docs, rather than a link?

mmmarklu:
@lavinigam-gcp Polong said "make a brief mention in ADK docs and link out to CGC docs as the source of truth, but open to suggestions"

Launch is on Monday.

Sita04:
@mmmarklu Kindly ping me once the changes for the docs are in and I can merge them.

PR #457, opened 2025-06-28
refactor: update MCP examples to use StdioConnectionParams

andrewginns:
- Replace StdioServerParameters with StdioConnectionParams wrapper as recommended
    - `StdioServerParameters is not recommended. Please use StdioConnectionParams.`
  - Fix import paths to prevent ImportError
    - `StdioServerParameters` can't be imported from `mcp_toolset` in testing with `google-adk>=1.5.0`


matt-buckley:
Would be great to get this one merged as this caused some confusion for me also, very basic fix!

andrewginns:
@matt-buckley it looks like the adk team made the same changes independently so the main branch shouldn't have this issue.

I just merged main into this branch and can't see any difference beyond minor linting/style changes. Closing this due to that.

matt-buckley:
@andrewginns I believe that's because the merge simply overwrote your changes - the version in `main` [here](https://github.com/google/adk-docs/blob/main/docs/tools/mcp-tools.md) still tries to import from `mcp_toolset` when this is no longer possible.

andrewginns:
@matt-buckley indeed you're right, I saw they updated the MCP params but the imports haven't been changed. Reopened and updated.

joefernandez (approved):

PR #458, opened 2025-06-29
chore: update name of adk package in python examples

marianocodes:
Update package name in some of the examples. 

Before `google_adk` . After `google-adk`


tpryan (approved):
Yep this seems like a typo. Thanks.

PR #499, opened 2025-07-09
chore: add meta info to doc (last update time)

Sita04:


docs_health_report.md
```diff
@@ -0,0 +1,112 @@
+<!-- BEGIN_DOCS_HEALTH_REPORT -->
+# Sample Documentation Health Report
```
joefernandez, line 2: Do we want to call this "Sample" since we are publishing this? Maybe date it? 

# Documentation Health Report - July 2025

docs/agents/index.md
```diff
@@ -1,3 +1,7 @@
+---
+version: 1.0.0
```
joefernandez, line 2: Is this going to be visible on the page? Also: because this number looks like a version number for ADK, I think at least some people are going to assume that it corresponds with a ADK version number, in which case this makes it look like the docs are way out of date. Maybe it's better to display a last updated date and keep the version as hidden metadata?

boyangsvl: +1, the version is a bit confusing.

joefernandez:
Closing this but OK to reopen if we want to revisit this approach.

PR #502, opened 2025-07-10
docs: add ADK logging documentation

xuanyang15:
We have some users requesting a public doc for logging. This doc is supposed to help users to understand how logging is setup in ADK and how they can use the logs to get the information they need.


polong-lin:
Since this is a new page, you'll also need to add it to mkdocs.yml so that it shows up on the left-nav:

Could you please add it as the first line here?

```
  - Observability:
    - Logging: observability/logging.md
    - Arize AX: observability/arize-ax.md
    - Phoenix: observability/phoenix.md 
```

File:
https://github.com/google/adk-docs/blob/ae18ef6d1311df45af12741983ab2dddda4e0c77/mkdocs.yml#L165

lavinigam-gcp:
@polong-lin Added the page in mkdocs.

PR #514, opened 2025-07-16
Local models with adk-java docs

shelajev:
- This pull request adds documentation on using adk-java wrapper for Lanchain4j for wiring models other than Gemini or Claude.
- It includes an example with Docker Model Runner for running open source models locally.
- It includes the necessary dependencies configuration: adk-java-contrib & langchain provider.


joefernandez:
Thanks for this contribution! Unfortunately, this change is now obsolete due to recent restructuring of the ADK documentation.

PR #544, opened 2025-07-29
feat: add mcp-tools-java.md

wenhaozhao:
Add mcp-tools-java.md to Introduce How to Configure MCP-Tools in ADK-Java

### DIFF
--- a/docs/tools/mcp-tools-java.md
+++ b/docs/tools/mcp-tools-java.md
@@ -1,0 +1,160 @@
+# Model Context Protocol Tools
+
+ This guide walks you through two ways of integrating Model Context Protocol (MCP) with ADK.
+
+## What is Model Context Protocol (MCP)?
+
+The Model Context Protocol (MCP) is an open standard designed to standardize how Large Language Models (LLMs) like Gemini and Claude communicate with external applications, data sources, and tools. Think of it as a universal connection mechanism that simplifies how LLMs obtain context, execute actions, and interact with various systems.
+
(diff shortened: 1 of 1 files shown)


vorburger (changes requested):
I don't think we can just merge this as-is ... because this appears to just copy/paste a fair bit of content from https://github.com/wenhaozhao/adk-docs/blob/f2e9128c7c7fd2287c007c7a0fb598dc7c7ef0ca/docs/tools/mcp-tools.md (which is https://google.github.io/adk-docs/tools/mcp-tools/), without even adding it to the navigation. My understanding of how adk-docs are managed is that we use the same MD for Python & Java, and then have language specific sections with those tabs on them? @wenhaozhao would you be willing to update this accordingly?

vorburger:
@wenhaozhao I just noticed that (meanwhile) on https://google.github.io/adk-docs/tools/mcp-tools/ there actually is a _"For Java, refer to the following sample to define an agent that initializes the MCPToolset"_ would you perhaps like to simply enhance that and add how to configure timeouts for MCP server, introduced in https://github.com/google/adk-java/pull/272?

wenhaozhao:
ok, i'll update this later

vorburger (approved):
LGTM. PS: @wenhaozhao you could [separately, also] contribute this to our (brand new) https://github.com/google/adk-java/tree/main/tutorials BTW, if you like ... just a thought.

wenhaozhao:
> LGTM. PS: @wenhaozhao you could [separately, also] contribute this to our (brand new) https://github.com/google/adk-java/tree/main/tutorials BTW, if you like ... just a thought.

cool!!!

joefernandez:
obsolete, due to: https://github.com/google/adk-docs/pull/1496

PR #560, opened 2025-08-05
docs: remove coming soon from gke deploy

moficodes:


moficodes:
@polong-lin 

Since adk 1.9 released `adk deploy gke` is now a valid command.

PR #595, opened 2025-08-13
fix: add code example for skip_summarization

slothwriter:


tpryan (commented):
I feel like this needs an additional comment that explains that you are skipping summarization because the content will come back from the summary_agent already summated therefore you should skip the automatic step of summarizing that summarization refers to. Or at least that is my read of the whole thing. If i misunderstand please push back.

slothwriter:
@tpryan I just realized the explanation is in the documentation already, see the section about Customization[https://google.github.io/adk-docs/tools/function-tools/#customization](url). Let's keep the code snippet clean and keep comments in the code minimal.

tpryan (approved):

PR #605, opened 2025-08-13
docs: call BigQuery tools native tools rather than built-in tools

shobsi:


joefernandez:
We should NOT fragment the documentation for BigQuery tool support because we don't like term "built-in". Lets change the build-in terminology instead of creating this doc fragmentation.

PR #627, opened 2025-08-19
Removed constraint from `output_schema`

serranoarevalo:
Since [this](https://github.com/google/adk-python/commit/af635674b5d3c128cf21737056e091646283aeb7) got merged it seems that the constraint does not exist anymore.


dylan-apex:
You should also update https://github.com/google/adk-docs/blob/main/examples/python/snippets/agents/llm-agent/capital_agent.py as it references the same limitation

PR #631, opened 2025-08-20
docs: standardize use of `Agent` in Python examples

### LINKED ISSUE #629: Doc has inconsistent use of LlmAgent and Agent even in the same page
Problem:

- The code examples in the documentation create a confusing experience by alternating between LlmAgent and Agent for initialization. A user might follow one example, and then in the next section, the required import changes without any context or explanation for the switch.

Proposed Solution:

- For better clarity and consistency, the documentation should be standardized to use a single class, such as LlmAgent, across all examples. An explicit note should be added to inform users that LlmAgent and Agent are functionally identical and can be used interchangeably.

While [a note exists](https://google.github.io/adk-docs/agents/llm-agents/) stating they are the same, its current wording is ambiguous. I initially interpreted it to mean LlmAgent is simply referred to as Agent conceptually, not that they were equivalent imports.

dylan-apex:
Fixes #629 

This PR standardizes the use of the `Agent` from `google.adk.agents` over `LlmAgent` for Python references and example code.
It is an opinionated change, but even with an explanation that `LlmAgent == Agent`, for some people this goes over their head, so we should not intermingle the two names in the same documentation.

I also standardized the import from `google.adk.agents.llm_agent` to `google.adk.agents` in various locations and added a `GEMINI_MODEL` export to `adk-docs/examples/python/agent-samples/youtube-shorts-assistant/util.py`


joefernandez:
Thanks for the contribution. While I appreciate the effort to standardize Agent syntax over LlmAgent, this change is too far reaching and too risky to be considered as a single documentation update. Consider breaking the PR up into smaller and well explained changes and keep the scope of each change more focused.

dylan-apex:
@joefernandez Would splitting out the `docs/` and `examples/` changes be acceptable, or how granular do you want the subsequent PRs to be?

PR #632, opened 2025-08-20
docs: Add Agent Config feature docs

joefernandez:
docs: Add Agent Config feature docs

PR #633, opened 2025-08-20
fix: add a stub index page for the Agent Config API docs

joefernandez:
fix: add a stub page for the Agent Config API docs


joefernandez:
Context: Creates the page that the failed link is pointing to, so we can clear the link check error in https://github.com/google/adk-docs/pull/632

seanzhou1023 (approved):

PR #648, opened 2025-08-25
feat: Update Safety and Security page to introduce ADK Plugins as a method for implementing agent-agnostic, global security policies

sammalik-lgtm:

PR #656, opened 2025-08-27
fix: Updating Safety PR to comply with style guidelines

joefernandez:


joefernandez:
Update for: https://github.com/google/adk-docs/pull/648

PR #658, opened 2025-08-28
docs(agentengine): update deployment instructions to use GA template

yeesian:
The GA template is available as of https://github.com/googleapis/python-aiplatform/releases/tag/v1.111.0


Sita04 (approved):
Thanks for the changes!

PR #680, opened 2025-09-09
Add TS examples for function tools and agent as a tool

slothwriter:
issue # 440671611


examples/typescript/snippets/tools/function-tools/agent-as-a-tool-example.ts
happyhuman, line 1: Would it make sense to put these two examples in separate subdirectories? Also, should not we include a `tsconfig.json` and a `package.json` for them to make sure the users can quickly run them?

examples/typescript/snippets/tools/function-tools/agent-as-a-tool-example.ts
```diff
@@ -0,0 +1,73 @@
...
+import {Part} from '@google/genai';
+
+/**
+ * This example demonstrates how to use an agent as a tool.
+ */
+async function main() {
+  // Define the summarization agent that will be used as a tool
+  const summaryAgent = new LlmAgent({
+    name: 'summary_agent',
+    model: 'gemini-2.5-flash',
```
happyhuman, line 15: Should we stay consistent with python and use 'gemini-2.0-flash'? I can go either way on this.

slothwriter: we should use 2.5 model because python library and code examples were created a while ago.

examples/typescript/snippets/tools/function-tools/agent-as-a-tool-example.ts
slothwriter: We should create a subdirectory for each feature md file i.e function-tools.md and then within this subdirectory, just the ts files for each example. Because I checked the path structure for function-tools in adk-docs/examples/python/snippets  - tools is a directory. function-tools is the subdirectory. No need to create one more layer of subdirectories for each example ts files, does it make sense?

slothwriter: And yes we agree we will add a tsconfig.json and a package.json for each feature so each sample code is self-contained.

PR #681, opened 2025-09-09
docs: Add Tool Confirmation feature

joefernandez:


tpryan (approved):

docs/tools/confirmation.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+This function than then be set as the parameter value for the
+`require_confirmation` parameter:
+
+```
+root_agent = Agent(
+   ...
+   tools=[
+        # Set require_confirmation to True to require user confirmation
```
DeanChensj, line 90: Update the comment here?  Something like: 

"Require confirmation when the amount exceed the confirmation threshold"

docs/tools/confirmation.md
```diff
@@ -0,0 +1,222 @@
...
+...
+```
+
+This implementation method requires minimal code, but is limited to simple
+approvals from the user or confirming system. For a complete example of this
+approach, see the
+[human_tool_confirmation](https://github.com/google/adk-python/blob/fc90ce968f114f84b14829f8117797a4c256d710/contributing/samples/human_tool_confirmation/agent.py)
+code sample.
+
+### Require confirmation function
```
DeanChensj, line 69: remove?

docs/tools/confirmation.md
DeanChensj, line 1: Somehow the code in this file does not have python highlight

docs/tools/confirmation.md
```diff
@@ -0,0 +1,222 @@
...
+
+-   The `id` in the `function_response` should match the `function_call_id`
+    from the `RequestConfirmation` `FunctionCall` event.
+-   The `name` should be `adk_request_confirmation`.
+-   The `response` object contains the confirmation status and any
+    additional payload data required by the tool.
+
+## Known limitations {#known-limitations}
+
+The tool confirmation feature has the following limitations:
```
DeanChensj, line 212: Live is also not supported for now

PR #733, opened 2025-10-01
feat: Add workflow to run Go program and create PR

tpryan:
Introduces a GitHub Actions workflow that checks out two repositories, sets up Go, installs dependencies, runs a Go program, and creates a pull request with generated documentation updates.


.github/workflows/feature-matrix-updater.yaml
```diff
@@ -0,0 +1,47 @@
+name: Run Go Program and Create PR
+on:
+  push:
+    branches:
+      - main
```
koverholt, line 5: Do we want this action to run on each push to main, or something else like manually from the GitHub UI, on a schedule, etc?

koverholt (approved):
Thanks for the PR! LGTM other than one comment about the action's trigger.

Optional sidenote: And for testing purposes perhaps you could temporarily point the action to this branch to see it run within this PR?

.github/workflows/feature-matrix-updater.yaml
tpryan: For now manual. But in the future, on updates.

koverholt: Resolved in #734.

PR #738, opened 2025-10-02
docs: add agent observability with opik

vincentkoc:
Opik is one of the leading open-source telemetry, evaluation and optimization platform. A large % of our open-source users are Google ADK users and our integration has been stable for sometime (we shipped day zero support early on). This PR introduces new documentation and examples following our reference guide (https://www.comet.com/docs/opik/python-sdk-reference/integrations/adk/overview/). Please reach out if you need any changes.

### DIFF
--- a/docs/observability/opik.md
+++ b/docs/observability/opik.md
@@ -1,0 +1,105 @@
+# Agent Observability with Opik
+
+[Opik](https://www.comet.com/opik/) is Comet's observability, evaluation, and optimization platform for LLM and agent workloads. It captures the full execution context of your Google ADK applications—including agent runs, tool invocations, model calls, and custom business logic—so you can debug faster, monitor cost and latency, and ship production-ready agents with confidence.
+
+## Why Opik for ADK?
+
+Opik extends ADK's native OpenTelemetry support with:
+
+- **Rich span data** – Every agent, tool, and LLM call is logged with prompts, responses, metadata, and token usage for all model calls.
+- **Automatic cost tracking** – Built-in cost calculators let you surface spend per span, per agent, or per project without additional code.
+- **Multi-agent visualization** – The `track_adk_agent_recursive` helper renders hierarchical graphs of complex agent workflows directly in the UI.
+- **Thread-aware tracing** – ADK session IDs are automatically mapped to Opik threads so multi-turn conversations stay grouped together.
+- **Error analytics** – Exceptions raised inside agents, models, or tools are captured with stack traces and surfaced in the Opik dashboard.
+
+## Installation
+
+Install Opik alongside Google ADK:
+
+```bash
+pip install opik google-adk
+```
+
+## Configure Opik
+
+Initialize the Opik Python SDK using whichever method fits your deployment. The quickest way during development is:
+
+```bash
+opik configure
+```
+
+You can also call `opik.configure()` from code or set the `OPIK_API_KEY`, `OPIK_WORKSPACE`, and `OPIK_PROJECT_NAME` environment variables. See the [Python SDK configuration guide](https://www.comet.com/docs/opik/tracing/sdk_configuration/) for the full matrix covering cloud, enterprise, and self-hosted deployments.
+
+## Configure Google ADK
+
+Opik works with any ADK-supported model provider. For the example below we authenticate to Gemini through Google AI Studio:
+
+```bash
+export GOOGLE_GENAI_USE_VERTEXAI=FALSE
+export GOOGLE_API_KEY="<your-gemini-api-key>"
+```
+
+If you deploy on Vertex AI instead, set `GOOGLE_GENAI_USE_VERTEXAI=TRUE` and configure `GOOGLE_CLOUD_PROJECT` plus `GOOGLE_CLOUD_LOCATION` (or use Application Default Credentials).
+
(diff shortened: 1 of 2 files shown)


docs/observability/opik.md
```diff
@@ -0,0 +1,105 @@
...
+export GOOGLE_API_KEY="<your-gemini-api-key>"
+```
+
+If you deploy on Vertex AI instead, set `GOOGLE_GENAI_USE_VERTEXAI=TRUE` and configure `GOOGLE_CLOUD_PROJECT` plus `GOOGLE_CLOUD_LOCATION` (or use Application Default Credentials).
+
+## Basic agent setup with OpikTracer
+
+The `OpikTracer` automatically captures agent execution, tool calls, and model interactions. Attach it to your ADK agents as shown below:
+
+```python
```
koverholt, line 50: I get the following error when I try to run this code sample (and after installing the deps mentioned earlier):

```
{"error": "Fail to load 'opik-test' module. cannot import name 'SequenceNotStr' from 'openai._types' (/Users/koverholt/miniforge3/lib/python3.11/site-packages/openai/_types.py)"}
```

I got a similar error when running the sample code from https://www.comet.com/opik/koverholt/get-started?integration=google-adk.

docs/observability/opik.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Troubleshooting
+
+- Always await all events when using `Runner.run_async`. Exiting early prevents the tracer from closing spans.
+- Call `tracer.flush()` before your script exits to guarantee that buffered spans are delivered.
+- Keep `opik` and `google-adk` up to date to benefit from the latest integration improvements.
+
+## Additional resources
+
+- [Opik ADK integration reference](https://www.comet.com/docs/opik/python-sdk-reference/integrations/adk/overview/)
```
koverholt: This link gives a 404 error.

vincentkoc: @koverholt i have fixed this, thanks!

docs/observability/opik.md
vincentkoc: @koverholt appreciate the review. Could you try a new venv/env and install the packages please, you likley have outdated package somewhere. Team and I have tested this on a clean env and were not able to recreate, if not can you share package versions for us to re-create the particular issue(s).

koverholt: Thanks, I updated to the latest version of Opik, and that error went away. But I ran into some other issues with the agent's design, and I will make a separate comment to update the code sample here.

docs/observability/opik.md
```diff
@@ -0,0 +1,116 @@
...
+    app_name="weather_app",
+    before_agent_callback=tracer.before_agent_callback,
+    after_agent_callback=tracer.after_agent_callback,
+    before_model_callback=tracer.before_model_callback,
+    after_model_callback=tracer.after_model_callback,
+)
+
+response = runner.run(user_request="What's the weather in New York?")
+tracer.flush()  # ensure spans are delivered before the script exits
+```
```
koverholt, line 74: 1. The `*_callback` kwargs should go into the `LlmAgent` instance rather than the `Runner` class, which is causing an error.

2. The `runner.run` syntax here is not working for me since `user_request` is not a valid kwarg. Instead, one would need to add a `new_message` kwarg with a content object, plus a session and user ID, which would require setting up a session service, and other config that would make this code sample much longer.

3. ADK is complaining about this agent because there is no `root_agent` defined.

**Proposed solution to address all of these**:

I suggest updating this code example to the following, then instruct users to run `adk web` or `adk run` (which I tested and it's working as expected) to interact with the agent and see the traces in Opik. That way those interfaces will handle setting up the session for the user and the code sample is simpler and focuses on the basic tracing implementation w/ callbacks.

---

The `OpikTracer` automatically captures agent execution, tool calls, and model interactions. Attach it to your ADK agents as shown below:

```python
from google.adk.agents import Agent
from opik.integrations.adk import OpikTracer

tracer = OpikTracer(project_name="adk-weather-demo")

root_agent = Agent(
    name="weather_agent",
    instruction="You provide friendly weather updates.",
    model="gemini-2.0-flash",
    before_agent_callback=tracer.before_agent_callback,
    after_agent_callback=tracer.after_agent_callback,
    before_model_callback=tracer.before_model_callback,
    after_model_callback=tracer.after_model_callback,
)

tracer.flush()  # Ensure spans are delivered before the script exits
```

Then run the agent with ADK using `adk web` or `adk run` and send a sample message such as:

> What's the weather in New York?

Now you can open the Opik dashboard and navigate to your project to inspect the trace. You will see the agent span, the downstream LLM call, tool invocations (if any), token usage, and cost estimates.

---

Thoughts on this version of the code sample?

koverholt (changes requested):
Thanks for the updates!

The code sample is not working as-is, so I left another comment along with a proposed version to get it working and make it simpler.

docs/observability/opik.md
vincentkoc: @koverholt super appreciate the response and a suggested example. Let me review and circle back asap 👍

koverholt:
Doing some cleanup in this repo and closing this PR due to staleness, but please feel free to reopen or submit a new PR with the requested changes when ready. Thanks again!

PR #743, opened 2025-10-08
Add AG-UI adk-middleware tutorial

maxkorp:
This adds a tutorial page for using the AG-UI adk-middleware to enable a python ADK agent to use AG-UI.

### DIFF
--- a/docs/tutorials/ag-ui.md
+++ b/docs/tutorials/ag-ui.md
@@ -1,51 +1,37 @@
-# Building in-app agentic chat experiences with ADK and the AG-UI protocol
-
----
+# Building chat experiences with AG-UI and CopilotKit
 
 ## Overview
 
-As an agent builder, you want your agent to be able to interact with users through a rich and responsive interface. Building that up from scratch will require a lot of effort, especially to support streaming events and client state. That's exactly what AG-UI was designed for, rich user experiences directly connected to an agent.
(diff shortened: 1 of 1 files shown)


docs/tutorials/ag-ui.md
zeroasterisk, line 1: I would prefer we start with the value props in the header, instead of the technology.

"As an agent builder, you want your agent to be able to interact with users in a rich interface. To do so, from scratch will require a lot of effort, especially to support streaming events and client state. 

That's exactly what AG UI was designed for, rich user experiences directly connected to an agent.

zeroasterisk, line 1: We should be more clear about AG UI and CopilotKit.

Introduce CopilotKit as an optional starter kit for developing a rich UI. List others which support AG UI. Then continue with this guide.

maxkorp:
@zeroasterisk and @koverholt I think I got everything addressed, let me know what you think!

maxkorp:
@zeroasterisk @koverholt anything else here? Fixed the cla issue (wanted an individual instead of a company one).

koverholt:
Thanks for updating your CLA, this all LGTM.

In the future when we rework tutorials, we might refactor this page and move it to a UI tutorials section or list of integrations. For now, this is a great addition to our existing tutorials to showcase the UI integrations out there in the ADK ecosystem.

PR #739, opened 2025-10-06
Phase 1: ADK Onboarding journey revision

joefernandez:
- Quickstart: split by language and reduce time
- Build your agent: new section for agent building
- Home page: Simplify onboarding choices
- Surface "About ADK" in navigation

PR #771, opened 2025-10-10
fix: Fix relative links on tutorials page

koverholt:
Uses relative links to the MD files on the updated tutorial page, which mkdocs will resolve to the correct URL path.

Followup PR to #739.

### DIFF
--- a/docs/tutorials/index.md
+++ b/docs/tutorials/index.md
@@ -19,16 +19,16 @@ applications with ADK. Explore our collection below and happy building:
 
     Create a workflow that uses multiple tools.
 
-    [:octicons-arrow-right-24: Build a multi-tool agent](/adk-docs/get-started/index.md)
+    [:octicons-arrow-right-24: Build a multi-tool agent](../get-started/quickstart.md)
 
 -   :material-console-line: **Agent team**
 
     ---
 
     Build an multi-agent workflow including agent delegation,
-    session management, and safety callbacks. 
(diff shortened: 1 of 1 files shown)


docs/tutorials/index.md
```diff
@@ -19,24 +19,24 @@ applications with ADK. Explore our collection below and happy building:
 
     Create a workflow that uses multiple tools.
 
-    [:octicons-arrow-right-24: Build a multi-tool agent](/adk-docs/get-started/index.md)
+    [:octicons-arrow-right-24: Build a multi-tool agent](../get-started/quickstart.md)
```
joefernandez, line 22:
```suggestion
    [:octicons-arrow-right-24: Build a multi-tool agent](/adk-docs/get-started/quickstart/)
```

joefernandez: keep the absolute links so they don't break when we move this page

koverholt: The mkdocs convention is to use a relative link to the MD file, which it then resolves in the built version: https://www.mkdocs.org/user-guide/writing-your-docs/#linking-to-pages. Otherwise if you use absolute links like `/adk-docs/get-started/quickstart/`, then mkdocs is unaware if it's resolving to the right place and is unable to verify the link during the build process.

joefernandez: I'm disagreeing and committing. :-) approved. Let's talk separately.

PR #774, opened 2025-10-13
feat: Add edit button to propose changes to existing docs pages

koverholt:
This PR allows contributors to quickly get to the appopriate edit page in the docs repository to propose changes to existing docs pages:

- Adds the theme feature to edit existing docs (`content.action.edit`)
- Adds the appropriate `edit_uri` prefix for the ADK docs repo
- Reworks the ADK library links in the top nav bar (due to required changes for edit functionality)

Example of newly added edit button:

Screenshot 2025-10-13 at 1 40 07 PM

Which then leads to the edit page for that .md file and displays a notification for external contributors similar to the following:

> You’re making changes in a project you don’t have write access to. Submitting a change will write it to a new branch in your fork koverholt/adk-docs, so you can send a pull request.

### DIFF
--- a/docs/deploy/agent-engine.md
+++ b/docs/deploy/agent-engine.md
@@ -58,9 +58,6 @@ You need the following resources configured to use this deployment path:
 -   **Make tool**: Build automation tool. This tool is part of most
     Unix-based systems, for installation details, see the 
     [Make tool](https://www.gnu.org/software/make/) documentation.
--   **Terraform**: Infrastructure and services deployment on Google Cloud.
-    For installation details, see 
-    [Install Terraform](https://developer.hashicorp.com/terraform/install).
 
 ### Prepare your ADK project {#prepare-ad}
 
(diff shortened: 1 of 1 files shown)


mkdocs.yml
```diff
@@ -8,8 +8,9 @@ exclude_docs:
   api-reference/java/legal/*
 
 # Repository
-repo_name: adk-python
-repo_url: https://github.com/google/adk-python
+repo_name: adk-docs
+repo_url: https://github.com/google/adk-docs
+edit_uri: edit/main/docs/
```
joefernandez, line 13: If you don't have edit access to main, I think this redirects you, but I don't even want folks who have edit access to main to make edits there. I think we should change this to a designated "edit" branch:

edit_uri: edit/edit-docs/docs/

test implementation: go/adk-docs-stage

koverholt: This `edit_uri` option is actually used for the source branch. Without this option defined, it [defaults](https://www.mkdocs.org/user-guide/configuration/#edit_uri) to `edit/master/docs` which leads to 404s, so we just need to align it with `main` for our repo.

For the destination branch, the user needs to specify a branch (or branch on their fork) within the GitHub edit UI, as shown in [Step 8 here](https://docs.github.com/en/repositories/working-with-files/managing-files/editing-files#editing-files-in-your-repository). So unfortunately we cannot pre-specify a destination branch to use with this option, rather that gets set in GitHub's edit UI after clicking the edit button.

joefernandez: OK, thanks for the explanation.

PR #784, opened 2025-10-17
docs: Add Resume feature docs

joefernandez:


docs/runtime/resume.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Add resumable configuration
+
+Enable the resume function for an agent workflow by applying a Resumabiltiy
+configuration to the App object of your ADK workflow, as shown in the following
+code example:
+
+```python
+app = App(
+    name='human_tool_confirmation',
```
DeanChensj: nit. The app name can be updated to something not that specific

docs/runtime/resume.md
```diff
@@ -0,0 +1,241 @@
...
+
+!!! warning "Caution: Long Running Functions, Confirmations, Authentication"
+    For agents that use
+    [Long Running Functions](/adk-docs/tools/function-tools/#long-run-tool),
+    [Confirmations](/adk-docs/tools/confirmation/), or
+    [Authentication](/adk-docs/tools/authentication/)
+    requiring user input, adding a resumable confirmation changes how these features
+    operate. For more information, see the documentation for those features.
+
+!!! info "Note: Custom Agents"
```
DeanChensj, line 41: Maybe this can be moved to later section?

docs/runtime/resume.md
```diff
@@ -0,0 +1,241 @@
...
+adk api_server my_resumable_agent/
+
+# resume the agent:
+curl -X POST http://localhost:8000/run_sse \
+ -H "Content-Type: application/json" \
+ -d '{
+   "app_name": "my_resumable_agent",
+   "user_id": "u_123",
+   "session_id": "s_abc",
+   "invocation_id": "invocation-123",
```
DeanChensj, line 69: let's call out what invocation or invocation_id is, and where to find them before using it here?

docs/runtime/resume.md
```diff
@@ -0,0 +1,241 @@
...
+ }'
+```
+
+You can also restart a workflow using the Runner object Run Async method, as
```
DeanChensj: s/restart/resume

docs/runtime/resume.md
```diff
@@ -0,0 +1,241 @@
...
+shown below:
+
+```python
+runner.run_async(user_id='u_123', session_id='s_abc', 
+    invocation_id='invocation-123' ,new_message=)
```
DeanChensj: Remove new_message here?  Since we already cover it in below comments.

docs/runtime/resume.md
```diff
@@ -0,0 +1,241 @@
...
+
+# when new_message is set to a function response,
```
DeanChensj: Capitalize first letter?

docs/runtime/resume.md
```diff
@@ -0,0 +1,241 @@
...
+    command line (CLI) tool is not currently supported.
+
+## How it works
+
+The resume functionality works by logging completed Agent workflow tasks,
+including incremental steps using
+[Events](/adk-docs/events/) and
+[Event Actions](/adk-docs/events/#detecting-actions-and-side-effects).
+tracking completion of agent tasks within a resumable workflow. If a workflow is
+interrupted and then later restarted, the resume functionality reinstates the
```
DeanChensj: Should we just say "ADK reinstates the interrupted workflow" rather than "the resume functionality"? 

Same for other paragraphs.

docs/runtime/resume.md
```diff
@@ -0,0 +1,241 @@
...
+-   **Sequential Agent**: Reads the current_sub_agent from its saved state
+    to find the next sub-agent to run in the sequence.
+-   **Loop Agent**: Uses the current_sub_agent and times_looped values to
+    continue the loop from the last completed iteration and sub-agent.
+-   **Parallel Agent**: Determines which sub-agents have already completed
+    and only runs those that have not finished.
+
+Event logging includes results from Tools which successfully returned a result.
+So if an agent successfully executed Function Tools A and B, and then failed
+during execution of tool C, the resume feature reinstates the results from the
```
DeanChensj: ditto, should we just use ADK as the subject?

docs/runtime/resume.md
```diff
@@ -0,0 +1,241 @@
...
+    When resuming a workflow with Tools, the Resume feature ensures
+    that the Tools in an agent are run ***at least once***, and may run more than
+    once when resuming a workflow. If your agent uses Tools where duplicate runs
+    would have a negative impact, such as purchases, you should modify the Tool to
+    check for and prevent duplicate runs.
+
+!!! note "Note: Workflow modification with Resume not supported"
+    Do not modify a stopped agent workflow before resuming the workflow. 
+    For example adding or removing agents from workflow that has stopped
+    and then resuming that workflow is not supported.
```
DeanChensj, line 124: I think we should change the wording, "not supported" sounds like it can be supported, but the thing is this is undefined behavior

DeanChensj (approved):

docs/runtime/resume.md
joefernandez: done.

docs/runtime/resume.md
joefernandez: I'm including it here, because once you add the resumable config, that is going to impact your Custom Agents if you have them, so we need to surface that consequence early.

docs/runtime/resume.md
joefernandez: updated.

docs/runtime/resume.md
joefernandez: Yes. I did that by introducing "Invocation ID" in the intro paragraph and the link to the Event topic is for.

docs/runtime/resume.md
joefernandez: removed it.

docs/runtime/resume.md
joefernandez: done.

docs/runtime/resume.md
joefernandez: updated.

docs/runtime/resume.md
joefernandez: updated

docs/runtime/resume.md
joefernandez: not supported means not supported. :-) It's not a promise to support it in the future, and calling "undefined behavior" is too vague; someone might interpret that as "it might work! try it!"

PR #791, opened 2025-10-20
feat: Add sections and groupings to sidebar navigation menu

koverholt:
This PR adds navigation sections to group related pages into more easily readable sections:

- Adds `navigation.sections` to config
- Adds sections for `Building Agents`, `Running Agents`, `Components`, and `Reference`
- Adds CSS for improved styling and readability of nav sections
- **No changes** to page URLs or page locations in folders

Current sidebar nav:

Screenshot 2025-10-20 at 6 54 29 PM

New sidebar nav:

Screenshot 2025-10-20 at 6 54 35 PM

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -48,7 +48,7 @@ from simple tasks to complex workflows.
 <p style="text-align:center;">
   <a href="/adk-docs/get-started/python/" class="md-button" style="margin:3px">Start with Python</a>
   <a href="/adk-docs/get-started/java/" class="md-button" style="margin:3px">Start with Java</a>
-  <a href="/adk-docs/get-started/about/" class="md-button" style="margin:3px">Technical overview</a>
+  <a href="/adk-docs/get-started/about/" class="md-button" style="margin:3px">Technical Overview</a>
 </p>
 
 ---


mkdocs.yml
```diff
@@ -125,123 +126,129 @@ plugins:
...
-  - Custom Tools:
-    - tools-custom/index.md
-    - Function tools:
-      - Overview: tools/function-tools.md
-      - Tool performance: tools/performance.md
-      - Action confirmations: tools/confirmation.md
-    - MCP tools: tools/mcp-tools.md
-    - OpenAPI tools: tools-custom/openapi-tools.md
-    - Authentication: tools/authentication.md
+  - Building Agents:
```
joefernandez: Build Agents

generally we want to avoid gerunds (*-ing) in titles.

mkdocs.yml
```diff
@@ -125,123 +126,129 @@ plugins:
...
+    - Home:
```
joefernandez: I think this would be better *before* the "Building Agents" heading, actually, since it really just a jumping off point.

If that's not technically possible, then this placement is OK.

mkdocs.yml
```diff
@@ -125,123 +126,129 @@ plugins:
```
joefernandez, line 127: is their any way we can get rid of the "Agent Development Kit" heading above the "Building Agents" heading? Seems redundant given the top nav bar.

mkdocs.yml
```diff
@@ -125,123 +126,129 @@ plugins:
...
+      - tools-custom/index.md
+      - Function tools:
+        - Overview: tools/function-tools.md
+        - Tool performance: tools/performance.md
+        - Action confirmations: tools/confirmation.md
+      - MCP tools: tools/mcp-tools.md
+      - OpenAPI tools: tools/third-party/openapi-tools.md
+      - Authentication: tools/authentication.md
+    - Technical overview: get-started/about.md
   - Running Agents:
```
joefernandez: "Run Agents"

mkdocs.yml
```diff
@@ -125,123 +126,129 @@ plugins:
...
-  - Community Resources: community.md
-  - Contributing Guide: contributing-guide.md
-  - API Reference:
-    - api-reference/index.md
-    - Python ADK: api-reference/python/index.html
-    - Java ADK: api-reference/java/index.html
-    - CLI Reference: api-reference/cli/index.html
-    - Agent Config reference: api-reference/agentconfig/index.html
-    - REST API: api-reference/rest/index.md
+    - Agent Runtime:
```
joefernandez, line 186: No action for now, but we actually need a "Run Agents" topic that covers the various ways to execute an Agent workflow (CLI interface, Web UI, API server.

Then we can move this topic (or some of it) to the Components section

mkdocs.yml
```diff
@@ -125,123 +126,129 @@ plugins:
...
+      - Agent Runtime: runtime/index.md
+      - Runtime Config: runtime/runconfig.md
+      - Resume Agents: runtime/resume.md
+    - Deploy:
```
joefernandez: Deployment

(make the language parallel to other heading titles; nouns instead of verbs)

mkdocs.yml
```diff
@@ -125,123 +126,129 @@ plugins:
...
+      - Cloud Run: deploy/cloud-run.md
+      - GKE: deploy/gke.md
+    - Observability:
+      - Logging: observability/logging.md
+      - Cloud Trace: observability/cloud-trace.md
+      - AgentOps: observability/agentops.md
+      - Arize AX: observability/arize-ax.md
+      - Phoenix: observability/phoenix.md
+      - W&B Weave: observability/weave.md
+    - Evaluate:
```
joefernandez: Evaluation

(make the language parallel to other heading titles)

mkdocs.yml
```diff
@@ -125,123 +126,129 @@ plugins:
...
+    - Custom Tools:
+      - tools-custom/index.md
+      - Function tools:
+        - Overview: tools/function-tools.md
+        - Tool performance: tools/performance.md
+        - Action confirmations: tools/confirmation.md
+      - MCP tools: tools/mcp-tools.md
+      - OpenAPI tools: tools/third-party/openapi-tools.md
+      - Authentication: tools/authentication.md
+    - Technical overview: get-started/about.md
```
joefernandez: The "Technical overview" fits better at the top of the "Components" section, now.

mkdocs.yml
```diff
@@ -125,123 +126,129 @@ plugins:
...
+      - AgentOps: observability/agentops.md
+      - Arize AX: observability/arize-ax.md
+      - Phoenix: observability/phoenix.md
+      - W&B Weave: observability/weave.md
+    - Evaluate:
+      - evaluate/index.md
+      - Criteria: evaluate/criteria.md
+    - Safety and Security:
+      - safety/index.md
+  - Components:
```
joefernandez, line 208: based on the relative importance of the components, I think the order should be:

- Technical overview
- Context
- Sessions & Memory
- Events
- Artifacts
- Callbacks
- Plugins
- MCP
- Agent2Agent (A2A)
- Streaming (Live Bidi)
- Grounding

mkdocs.yml
```diff
@@ -125,123 +126,129 @@ plugins:
...
+      - Java: get-started/java.md
+    - Build your agent:
+      - tutorials/index.md
+      - Multi-tool agent: get-started/quickstart.md
+      - Agent team: tutorials/agent-team.md
+      - Streaming agent:
+        - get-started/streaming/index.md
+        - Python: get-started/streaming/quickstart-streaming.md
+        - Java: get-started/streaming/quickstart-streaming-java.md
+      - Testing: get-started/testing.md
```
joefernandez: "Testing" (rename: "Test" should go under "Running Agents", after the "Agent Runtime" topic

The "Testing" topic is a mess; it doesn't know what it wants to be, and it probably should be several pages, not one. Task: b/457218750

joefernandez (approved):
See comments. Approved but let's talk about when you merge this.

mkdocs.yml
koverholt: Fixed in 1671490.

mkdocs.yml
koverholt: Fixed in 1671490.

mkdocs.yml
koverholt: Fixed in 3e3d985.

mkdocs.yml
koverholt: Fixed in 3e3d985.

mkdocs.yml
koverholt: +1, agree very much! 😁

mkdocs.yml
koverholt: Fixed in 448ca69.

mkdocs.yml
koverholt: In the spirit of [this other comment](https://github.com/google/adk-docs/pull/791/#discussion_r2485037486) and since the `Testing` page is almost entirely about ADK's API server, I renamed this page to `Use the API Server` and moved it under `Run Agents > Runtime` in b50c2e2. (Also added a redirect from the old Testing page URL).

mkdocs.yml
koverholt: Mostly agree with the order here, except I reversed the order of Events, Artifacts, and Callbacks to Callbacks, Artifacts, and Events. And I suspect we will tweak the order more in the future based on traffic / usage patterns. 😁 

Fixed in b44dccb.

mkdocs.yml
koverholt: Agreed! I moved this up, which makes it appear as a styled top-level section. Fixed in 6a118cc.

mkdocs.yml
koverholt: Agreed. I redefined an override for `nav.html` in 6a118cc to remove the redundant site title.

koverholt:
Thanks for the thorough review and comments. All comments have been addressed. New nav bar:

Screenshot 2025-11-03 at 11 51 49 AM

PR #822, opened 2025-10-27
docs: Remove incorrect state wrapper in session post payload

kevinkassimo:
Will close https://github.com/google/adk-python/issues/3223 with this merged.

Docs incorrectly wrap state in an additional `state` key. This results in `"state": {"state": {...}}` in the actual session state. This is because in the original source code, FastAPI would place every remaining payload under the state param
https://github.com/google/adk-python/blob/36ca4f1204fff4302f143dddccbf6259d3ae7872/src/google/adk/cli/adk_web_server.py#L761-L776

Update docs to reflect this.

**Testing:** Verified no state key is needed by creating a minimal agent and run session creation:

```bash
# Setup a simple my_sample_agent

$ adk api_server 2>&1 > /dev/null &

$ curl -X POST http://localhost:8000/apps/my_sample_agent/users/u_123/sessions/s_abc   -H "Content-Type: application/json"   -d '{"visit_count": 5}'
{"id":"s_abc","appName":"my_sample_agent","userId":"u_123","state":{"visit_count":5},"events":[],"lastUpdateTime":1761590753.623595}
```


koverholt (approved):
Thanks for the PR and for your contribution. ❤️ 

That's a great catch, thanks for linking to the issue and including info on the verification test. I also tested and can confirm the revised requests are working as expected without wrapping the vars with another `state` key. I also checked and it looks like you covered all of the occurrences of where the changes need to happen.

LGTM, and cc @joefernandez for additional / second review.

PR #825, opened 2025-10-27
Fix the code sample of using BigQuery with the application default credentails

slilichenko:
Current code incorrectly gets the auth token at the time the toolset is created. Once the token expires, the authentication will stop working


examples/python/snippets/tools/built-in-tools/bigquery.py
```diff
@@ -34,13 +34,10 @@
 # Define a tool configuration to block any write operations
 tool_config = BigQueryToolConfig(write_mode=WriteMode.BLOCKED)
 
-# Define a credentials config - in this example we are using application default
-# credentials
+# We don't configure the credentials in this example. The agent will use
+# so called Application Default Credentials:
```
koverholt: Should we reword this comment to mention that the ADC should be managed outside of the scope of the agent script / code to avoid coupling the auth token to this script's lifecycle?

slilichenko: We can try to re-word it a bit and mention that "by default, ADC would be used by the BigQuery APIs if the explicit credentials are not provided". Or can you suggest a better wording?

koverholt: Yes, that's a better explanation. How about something like:

```
# Uses externally-managed Application Default Credentials (ADC) by default.
# This decouples authentication from the agent / tool lifecycle.
# https://cloud.google.com/docs/authentication/provide-credentials-adc
```

slilichenko: SG. Made the updates.

PR #830, opened 2025-10-28
fix: Update Java ADK docs to reflect updated agent dir

Poggecci:
Resolving issues raised in https://github.com/google/adk-java/issues/456#issuecomment-3424597355


docs/get-started/java.md
```diff
@@ -252,7 +252,7 @@ Run your agent with the ADK web interface using the following Maven command:
 # Remember to load keys and settings: source .env OR env.bat
 mvn compile exec:java \
     -Dexec.mainClass="com.google.adk.web.AdkWebServer" \
-    -Dexec.args="--adk.agents.source-dir=target --server.port=8000"
+    -Dexec.args="--adk.agents.source-dir=. --server.port=8000"
```
koverholt: I wonder why `--adk.agents.source-dir=target` is working in my environment (where the compiled classes end up in `target/classes/com/example/agent`) but was not working for you and others in https://github.com/google/adk-java/issues/456? Do your compiled classes end up in a different location? Asking since we want to make sure we're matching common dev environments as accurately as possible.

docs/get-started/streaming/quickstart-streaming-java.md
```diff
@@ -122,7 +122,7 @@ Run the following command from the terminal to launch the Dev UI.
 ```console title="terminal"
 mvn exec:java \
     -Dexec.mainClass="com.google.adk.web.AdkWebServer" \
-    -Dexec.args="--adk.agents.source-dir=src/main/java" \
+    -Dexec.args="--adk.agents.source-dir=. --server.port=8080"
```
koverholt: We should be consistent across both pages and either use port 8000 (to match the Python getting started guide) or port 8080 (to match the default when using adk web with Java). I'd lean towards using 8080 in both since it's the default.

docs/get-started/streaming/quickstart-streaming-java.md
```diff
@@ -135,8 +135,7 @@ the dropdown. Select "science-app".
 !!!note "Troubleshooting"
 
     If you do not see "science-app" in the dropdown menu, make sure you
-    are running the `mvn` command at the location where your Java source code
-    is located (usually `src/main/java`).
+    are running the `mvn` command from the root of your maven project.
```
koverholt, line 138: [No change needed]

Thanks for this more general clarification!

docs/get-started/java.md
joefernandez: DO NOT MERGE. The existing command is tested and verified working. An update like this requires that we fully retest quickstart with the update before updating. DO NOT MERGE.

docs/get-started/streaming/quickstart-streaming-java.md
joefernandez: This quickstart change should be retested before merging.

Poggecci: I'll revert the port specification to just focus on getting the critical change in this PR through.

docs/get-started/java.md
Poggecci: Ah, I had update this for consistency with the other change, but I am ok with keeping this one as is since it is WAI. Reverting this change.

Poggecci:
@joefernandez  ptal for re-review. I just updated the single line definitively raising issues for users as originally reported.

joefernandez (approved):

PR #840, opened 2025-10-31
docs: Add Context compaction docs

joefernandez:
task: b/455598275


docs/context/compaction.md
```diff
@@ -0,0 +1,63 @@
+# Compress workflow context for performance
+
+As an ADK agent runs a workflow it builds up information, including user
```
hangfei: i don't know if it's specific to "workflow". see if there is a more generic word.

docs/context/compaction.md
```diff
@@ -0,0 +1,63 @@
...
+instructions, collected data, tool responses, and generated content. As the size
+of this in-process data grows, agent processing times typically also increase.
```
hangfei: in-process might be confusing as it confuses with computer process

docs/context/compaction.md
```diff
@@ -0,0 +1,63 @@
...
+More and more data is sent to the generative AI model used by the agent,
+increasing processing time and slowing down responses. The ADK Context
+Compaction feature is designed to reduce the size of this in-process data, or
+context, by summarizing older parts of the agent workflow event history. 
+
+The Context Compaction feature uses a *sliding window* approach for workflow
```
hangfei: sliding window -> windowing

docs/context/compaction.md
```diff
@@ -0,0 +1,63 @@
...
+event data. When configured, it summarizes data from older events once it
+reaches a threshold of a specific number of workflow events.
+
+## Configure context compaction
+
+Add context compaction to your agent workflow by adding an Event Compaction
+Configuration setting to the App object of your workflow. As part of the
+configuration, you must specify an AI model to perform the event summarization,
```
hangfei: model is optional. if not set, it will use the root-agents model.

docs/context/compaction.md
```diff
@@ -0,0 +1,63 @@
+# Compress workflow context for performance
```
hangfei: was wondering if you plan to cover this: https://github.com/google/adk-python/blob/main/src/google/adk/apps/base_events_summarizer.py

hangfei: we offer a default impl of it: https://github.com/google/adk-python/blob/main/src/google/adk/apps/llm_event_summarizer.py

hangfei: users can customize their summary prompt through: https://github.com/google/adk-python/blob/main/src/google/adk/apps/llm_event_summarizer.py#L60

docs/context/compaction.md
```diff
@@ -0,0 +1,63 @@
...
+        compactor=my_compactor,
+        compaction_interval=5,  # Trigger compaction every 5 new invocations.
+        overlap_size=2          # Include last 2 invocations from the previous window.
+    ),
+)
+```
+
+Once configured, the ADK `Runner` automatically handles the compaction
+process in the background after each complete agent invocation event.
+
```
hangfei, line 42: do you think if we should talk about it will be compacted into EventAction.compaction: https://github.com/google/adk-python/blob/731bb9078d01359ae770719a8f5c003680ed9f3e/src/google/adk/events/event_actions.py#L97.

Then in later model call, the event action compaction(summary) will be used to cover those ranges that it summarized over.

docs/context/compaction.md
```diff
@@ -0,0 +1,63 @@
...
+As an ADK agent runs a workflow it builds up information, including user
+instructions, collected data, tool responses, and generated content. As the size
+of this in-process data grows, agent processing times typically also increase.
+More and more data is sent to the generative AI model used by the agent,
+increasing processing time and slowing down responses. The ADK Context
+Compaction feature is designed to reduce the size of this in-process data, or
+context, by summarizing older parts of the agent workflow event history. 
+
+The Context Compaction feature uses a *sliding window* approach for workflow
+event data. When configured, it summarizes data from older events once it
```
hangfei: we should mention it works off events of ADK session.

docs/context/compaction.md
```diff
@@ -0,0 +1,63 @@
...
+```python
+from google.adk.apps.sliding_window_compactor import SlidingWindowCompactor
+from google.adk.apps.app import EventsCompactionConfig
+from google.adk.models import Gemini
+
+# Define AI model to create summaries.
+summarization_llm = Gemini(model="gemini-2.5-flash")
+my_compactor = SlidingWindowCompactor(llm=summarization_llm)
+
+app = App(
```
yeesian, line 28: it's missing an import statement for `App`

docs/context/compaction.md
```diff
@@ -0,0 +1,63 @@
...
+    name='test',
+    root_agent=Mock(spec=BaseAgent),
```
yeesian: it's missing an import statement for `Mock` and `BaseAgent`

koverholt (approved):
Proposed changes LGTM, other reviewers have covered my issues in their comments.

One thing I want to check for consistency is that we're okay with using the terms "compaction", "compression", and "summarization" interchangeably to describe this feature throughout this page. The API mostly uses the term "compaction", and the underlying mechanism is "summarization". The term "compression" is not used outside of the (different) functionality for audio streaming.

docs/context/compaction.md
joefernandez: removed

docs/context/compaction.md
joefernandez: reworded.

docs/context/compaction.md
joefernandez: "windowing" is a less clear description than "sliding window". Stet.

docs/context/compaction.md
joefernandez: added mention of Session scope.

docs/context/compaction.md
joefernandez: got it. thanks for that detail.

docs/context/compaction.md
joefernandez: added

docs/context/compaction.md
joefernandez: Removed these and simplified.

docs/context/compaction.md
joefernandez: it's not necessary, since the system takes care of this, and you aren't exposing an additional control surface to the developer with EventAction.compaction .

docs/context/compaction.md
joefernandez: That's pretty deep in the technical weeds, but I'll add this detail to the configuration settings section.

joefernandez:
> Proposed changes LGTM, other reviewers have covered my issues in their comments.
> 
> One thing I want to check for consistency is that we're okay with using the terms "compaction", "compression", and "summarization" interchangeably to describe this feature throughout this page. The API mostly uses the term "compaction", and the underlying mechanism is "summarization". The term "compression" is not used outside of the (different) functionality for audio streaming.

We need to use different terms to describe the functionality depending on context, summarization is more applicable when we are talking about what is actually going on to "compress" the context. Compression is generally a more widely used term, so using it at the conceptual level is better for explanation of the feature. Talking about "compaction" is the least desireable, because it's not a commonly used term in software, but we named API functions with that term, so we have to use it when explaining the programming functions.

PR #845, opened 2025-10-31
fix(docs): Update bigquery.py creds

zeroasterisk:
I required this to make it work
application_default_credentials, _ = google.auth.default()


koverholt:
We actually just changed this in https://github.com/google/adk-docs/pull/825 to do the inverse since this pattern of using `google.auth.default()` within the tool couples the auth lifecycle to the agent's tool lifecycle and leads to unexpected token expirations. Using the currently documented approach, did you obtain ADC before running the code sample?

Pinging @slilichenko for guidance / clarification on this PR.

slilichenko:
The code sample you provided has a subtle bug. It works for a short period of time, and then the auth token will expire and the calls will start failing. Not providing the credentials at all will cause the BigQuery API to fall back on getting the credentials using ADC every time.

examples/python/snippets/tools/built-in-tools/bigquery.py
```diff
@@ -37,7 +37,10 @@
 # Uses externally-managed Application Default Credentials (ADC) by default.
 # This decouples authentication from the agent / tool lifecycle.
 # https://cloud.google.com/docs/authentication/provide-credentials-adc
```
koverholt:
```suggestion
# Use Application Default Credentials (ADC) for BigQuery authentication
# https://cloud.google.com/docs/authentication/provide-credentials-adc
```

koverholt (approved):
Revisiting this PR based on another incoming report of this error in #1150, which I was able to reproduce. Will merge this to fix. Thanks for the PR!

Note that the same pattern shown here with `google.auth.default()` and `BigQueryCredentialsConfig()` is also used in https://google.github.io/adk-docs/observability/bigquery-agent-analytics/.

PR #850, opened 2025-11-03
docs: add language support tags to pages

joefernandez:
- replace img.shield tags
- add language support tags where missing

### DIFF
--- a/docs/runtime/api-server.md
+++ b/docs/runtime/api-server.md
@@ -1,5 +1,9 @@
 # Use the API Server
 
+<div class="language-support-tag">
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v0.1.0</span><span class="lst-java">Java v0.1.0</span>
+</div>
+
 Before you deploy your agent, you should test it to ensure that it is working as
 intended. The easiest way to test your agent in your development environment is
 to use the ADK API server.


docs/agents/llm-agents.md
```diff
@@ -399,7 +405,10 @@ Control whether the agent receives the prior conversation history.
 
 ### Code Execution
 
-![python_only](https://img.shields.io/badge/Currently_supported_in-Python-blue){ title="This feature is currently available for Python. Java support is planned/ coming soon."}
+<div class="language-support-tag" title="hover text.">
+   <span class="supported">Supported in ADK</span>
+   <span class="python">Python v0.1.0</span>
```
koverholt: I think the line break here is causing this badge to render with a space in between the boxes.

Screenshot 2025-11-03 at 10 36 09 AM

docs/stylesheets/language-tags.css
```diff
@@ -0,0 +1,60 @@
...
+ color: black;
+ padding: 4px 6px;
+ display: inline-block;
+}
+.go {
+ background-color: #00ADD8;
+ color: black;
+ padding: 4px 6px;
+ display: inline-block;
+}
```
koverholt: For some reason this CSS selector is conflicting with and changing the bg color for code that is syntax highlighted as `console`. For example on this page: http://google.github.io/adk-docs/tools/gemini-api/computer-use/

Screenshot 2025-11-03 at 10 42 52 AM

docs/runtime/runconfig.md
```diff
@@ -313,7 +317,9 @@ This comprehensive example configures an agent with:
 
 ### Enabling Experimental CFC Support
```
koverholt, line 318: There is a note (admonition) higher on this page that states:

> The support_cfc feature is experimental and its API or behavior might change in future releases.

Do we want to replace that note with these new badges?

docs/agents/models.md
```diff
@@ -499,7 +506,9 @@ http://localhost:11434/api/chat \
 
 ### Self-Hosted Endpoint (e.g., vLLM)
 
-![python_only](https://img.shields.io/badge/Supported_in-Python-blue)
+<div class="language-support-tag" title="hover text.">
```
koverholt: Some of the language support labels have a title that shows on hover:

> title="This feature is an experimental preview release."
> title="Vertex AI Agent Engine currently supports only Python."

Recommend removing the titles throughout for simplicity / consistency.

docs/agents/custom-agents.md
```diff
@@ -4,6 +4,10 @@
 
 # Custom agents
 
+<div class="language-support-tag">
+  <span class="supported">Supported in ADK</span><span class="python">Python v0.1.0</span><span class="java">Java v0.2.0</span>
```
koverholt: In general, these version callouts are difficult to parse throughout the docs. But without proper docs versions, I can see why we need these to convey which release this became available in.

To simplify a bit, what if we only show versions for ADK Python > 1.0.0 to reduce the amount of older versions that are shown. So we change `Python v0.1.0`, `Python v0.2.0`, and `Python v0.5.0` to just `Python`?

docs/a2a/index.md
```diff
@@ -1,5 +1,9 @@
 # ADK with Agent2Agent (A2A) Protocol
 
+<div class="language-support-tag">
+  <span class="supported">Supported in ADK</span><span class="python">Python</span><span class="preview">Experimental</span>
```
koverholt: Is there a definition of experimental that we can link to throughout?

docs/stylesheets/language-tags.css
```diff
@@ -0,0 +1,60 @@
...
+.typescript {
+ background-color: #235A97;
+ padding: 4px 6px;
+ display: inline-block;
+}
+.preview {
```
koverholt: We might want to change this class name to `experimental` so we don't conflict with other definitions of preview that might clash with preview features in Cloud. For example,

> The Agent Engine Code Execution feature is a Preview release. For more information, see the [launch stage descriptions](https://cloud.google.com/products#product-launch-stages).

docs/a2a/index.md
joefernandez: Unfortunately not. Something we should add, though (with some negotiation).

docs/agents/llm-agents.md
joefernandez: fixed.

docs/agents/models.md
joefernandez: whoops! removed this

docs/runtime/runconfig.md
joefernandez: I'm updating that notice to be more consistent with other such notices, but leaving it. I want folks to have more than one opportunity to get that message, since it's the exception to most of the rest of this topic.

the pattern of use for the the tags should be that you see the tag in the primary place where it's documented.

docs/stylesheets/language-tags.css
joefernandez: good catch. thanks.

I renamed all the css classes to "lst-*" to avoid the class name conflicts like this.

docs/stylesheets/language-tags.css
joefernandez: Stet. Per prior discussions with PM, we want to eventually align with Cloud's "Preview" definition and not have our own "Experimental" release definition. Eventually "Experimental" goes away, and there is only "Preview."

docs/agents/custom-agents.md
joefernandez: I appreciate the feedback on the version notation, but this is the simplest, useful thing I think we can do (for now). The earliest release version helps developers see when feature was introduced (minimum installed version they need) and gives them a general idea of the feature's maturity.

Docs versions aren't going to solve this problem unless we do something like version the docs based on code release. Unfortunately, that's probably infeasible, since Python and Java are on separate release schedules and we probably don't want to maintain completely separate doc sets per supported language.

PR #863, opened 2025-11-04
Add documentation for User Simulation

KPJoshi:
This commit adds documentation for the User Simulation feature in ADK, which is already available but currently undocumented.

Tested locally using `mkdocs serve`.


docs/evaluate/user-sim.md
```diff
@@ -0,0 +1,153 @@
...
+## User simulator configuration
+
+You can override the default user simulator configuration to change the model,
+internal model behavior, and the maximum number of user-agent interactions.
+The below `EvalConfig` shows the default user simulator configuration:
+
+```json
+{
+  "criteria": {
+    # same as before
```
koverholt, line 132: Just wanted to make one quick callout that JSON doesn't support comments, and so this line renders the JSON as invalid - suggest removing this line. Otherwise thanks for the PR on this awesome new feature!

KPJoshi: I did this for compactness but I can restore it to be same as before. Alternatively I can reduce line count a bit by inlining some of the braces. WDYT?
Personally I'm fine with it as it is. The focus is the user sim config and copying the criteria is easy.

koverholt: Got it. Since it's less than 10 additional lines to include, I would suggest including the full code. Otherwise users might get JSON errors if they try to copy/paste that `EvalConfig` as-is. But I don't consider it a blocking issue.

PR #885, opened 2025-11-06
[ADK Go] Merging all Go docs from the private repo

ivanmkc:
This contains all ADK Go docs that were worked on in ivanmkc/adk-docs-private

### DIFF
--- a/docs/agents/llm-agents.md
+++ b/docs/agents/llm-agents.md
@@ -66,7 +66,7 @@ First, you need to establish what the agent *is* and what it's *for*.
             .build();
     ```
 
-=== "Golang"
+=== "Go"
 
     ```go
     // Example: Defining the basic identity
@@ -143,7 +143,7 @@ tells the agent:
             .build();
     ```
 
-=== "Golang"
+=== "Go"
 
     ```go
     // Example: Adding instructions
@@ -222,7 +222,7 @@ on the conversation and its instructions.
             .build();
     ```
 
-=== "Golang"
+=== "Go"
 
     ```go
     --8<-- "examples/go/snippets/agents/llm-agents/snippets/main.go:tool_example"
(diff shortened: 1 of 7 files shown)


mkdocs.yml
```diff
@@ -128,135 +128,123 @@ plugins:
...
-    - API Reference:
-      - api-reference/index.md
-      - Python ADK: api-reference/python/index.html
-      - Java ADK: api-reference/java/index.html
-      - CLI Reference: api-reference/cli/index.html
-      - Agent Config reference: api-reference/agentconfig/index.html
-      - REST API: api-reference/rest/index.md
-    - Community Resources: community.md
-    - Contributing Guide: contributing-guide.md
+  - Home: index.md
```
joefernandez: BLOCKING: nav has been reverted to old style. We CANNOT PUBLISH THIS: MUST FIX

joefernandez: Fix:
1. revert to published Nav
2. re-add (line ~136):
         - Go: get-started/go.md 
3. add (line ~258):
      - Go ADK: api-reference/java/index.html

docs/get-started/go.md
```diff
@@ -0,0 +1,176 @@
...
+        type nul > my_agent\agent.go
+        type nul > my_agent\env.bat
+        ```
+
+    === "MacOS / Linux"
+
+        ```bash
+        mkdir -p my_agent/ && \
+            touch my_agent/agent.go \
+            touch my_agent/.env
```
koverholt: This gets interpreted as:

```
touch my_agent/agent.go touch my_agent/.env
```

and creates a file called `touch`.

You can either add a `&&` before the `\` on the second line, or combine this into a single `touch` command to create both files.

Note that this issue also shows up in the Java page.

mkdocs.yml
koverholt: This is fixed in ce66983 and neighboring commits. Thanks!

PR #926, opened 2025-11-08
docs: MOVING files to /tools-custom/

joefernandez:
- move files, update links and navigation
- add redirects


joefernandez (commented):
Ready for review

PR #938, opened 2025-11-11
Change session update method to PATCH

scastiel:
Updated the method for session management from POST to PATCH and modified the request body structure accordingly.

Assuming that [the code is the source of truth](https://github.com/google/adk-python/blob/2b0f9532554d92b808e7f80988d62f582eb1ae31/src/google/adk/cli/adk_web_server.py#L814) here, the right method seems to be PATCH and not POST.

Fixes google/adk-docs#940

### DIFF
--- a/docs/runtime/api-server.md
+++ b/docs/runtime/api-server.md
@@ -321,26 +321,28 @@ curl -X GET http://localhost:8000/list-apps
 
 Sessions store the state and event history for a specific user's interaction with an agent.
 
-#### Create or Update a Session
+#### Update a Session
 
-Creates a new session or updates an existing one. If a session with the given IDs already exists, its state will be overwritten with the new state provided.
+Updates an existing session.
 
-*   **Method:** `POST`
+*   **Method:** `PATCH`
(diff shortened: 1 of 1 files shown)


koverholt:
Thanks for the PR, linked issue, and for your contribution!

One thing is that we will want to document how to create a session, which I assume is done with a `POST` per the original docs since `PATCH` is only used for updating, as you mentioned in #940.

scastiel:
> Thanks for the PR, linked issue, and for your contribution!
> 
> One thing is that we will want to document how to create a session, which I assume is done with a `POST` per the original docs since `PATCH` is only used for updating, as you mentioned in #940.

The session creation is already documented. Creating one with this POST endpoint (to create a session with ID) is actually deprecated in the code, which is why I removed documentation about it. I guess it can be kept, maybe mentioning it as deprecated?

koverholt:
Got it, I missed "Create a new session" at the top of the page since I was looking for it in the nearby sub-sections. Your changes make sense then, thanks for clarifying!

PR #948, opened 2025-11-13
Fix the broken Go build

rakyll:


ivanmkc (approved):
LGTM

I tested the revised code and it works well. Thanks for submitting!

PR #956, opened 2025-11-13
Replace the outdated `SlidingWindowCompactor` with the correct `LlmEventSummarizer`

ymodak:
I have rewritten the documentation to replace the outdated `SlidingWindowCompactor` with the correct `LlmEventSummarizer`, and I've updated the code example and explanatory text to match.


ymodak:
Fixes https://github.com/google/adk-docs/issues/965

PR #971, opened 2025-11-14
docs: Add Visual Builder guide

joefernandez:


docs/visual-builder/index.md
```diff
@@ -0,0 +1,117 @@
...
+    *   [**Sequential Agent:**](/adk-docs/agents/workflow-agents/sequential-agents/)
+        A workflow agent that executes a series of sub-agents in a sequence.
+    *   [**Loop Agent:**](/adk-docs/agents/workflow-agents/loop-agents/)
+        A workflow agent that repeatedly executes a sub-agent until a certain condition is met.
+    *   [**Parallel Agent:**](/adk-docs/agents/workflow-agents/parallel-agents/)
+        A workflow agent that executes multiple sub-agents concurrently.
+*   **Tools**
+    *   [**Prebuilt tools:**](/adk-docs/tools/built-in-tools/)
+        A limited set of ADK-provided tools can be added to agents.
+    *   [**Custom tools:**](/adk-docs/tools/built-in-tools/)
```
koverholt: This should link to `/adk-docs/tools-custom/` instead.

docs/visual-builder/index.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Get started
+
+The Visual Builder interface is part of the ADK Web tool user interface.
+Make sure you have ADK library
+[installed](/adk-docs/get-started/installation/#python)
+and then run the ADK Web user interface.
+
+```console
+adk web --port 8000
```
koverholt, line 26: Port 8000 is the default - should we remove the flag here?

docs/visual-builder/index.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+!!! note "Editing generated agents"
+
+    You can edit the generated files in your development environment. However,
+    some changes may not be compatible with Visual Builder.
+
+## Next steps
+
+Using the Visual Builder development Assistant, try building an new agent using
```
koverholt: "an new" -> "a new"

Jacksunwei (approved):

docs/visual-builder/index.md
joefernandez: I put the --port flag on there because that's usually the thing that causes a failure to load (port already in use). So including the port setting it makes easy to adjust (and you know what port it's grabbing).

docs/visual-builder/index.md
```diff
@@ -0,0 +1,117 @@
+# Visual Builder for agents
+
```
joefernandez, line 2: from Yifan Wang:

1) Can we emphasize: Be sure to hit Save the first time you are in the builder after making changes (or no changes), if you cancel directly the first time inside the builder for the app, then you won't be able to edit it in the builder later.

2) Edit is only enabled for apps that is built with builder. Agents with python code are not considered built by builder.

3) For function tool, you do need to provide the full python function name (meaning that you need to write the function tool in other IDE first and have the path ready)

4) Can we transform some of the screencast from this issue b/437396822 to gifs and attach them?

docs/visual-builder/index.md
joefernandez: Thanks! fixed

PR #974, opened 2025-11-15
Update Dockerfile with entrypoint which actually works

andruhon:
The example command didn't actually work in Cloud Run.

The changed command has `compile` step and points to `target` rather than to sources.


andruhon:
Signed CLA

joefernandez:
thanks for the update!

PR #977, opened 2025-11-17
Fix typo in roll_die function name to roll_dice

dastrobu:


koverholt:
This does not appear to be a typo in the docs, since `roll_die()` is mentioned on this page (https://github.com/google/adk-python/tree/main/contributing/samples/a2a_basic) and the `roll_die()` function is defined in this code (https://github.com/google/adk-python/blob/main/contributing/samples/a2a_basic/agent.py). Please leave a comment here or open an issue if there is something we are missing here. Thanks!

PR #985, opened 2025-11-19
ci: Add CI workflow to check docs build in PRs

koverholt:
This PR adds a GitHub action / workflow to build the docs in a PR as a pass/fail check.

It runs with `mkdocs build --strict` to fail on any warnings, and will ensure that a given branch / PR will not fail the docs build if merged, which could lead to silent failures until manually caught and fixed.


koverholt:
Tested with an example breakage to `mkdocs.yml` in 3cc926b that linked to a non-existent `.md` file, and the test failed as expected.

joefernandez (approved):

joefernandez:
@koverholt - Let's monitor this closely for a few days to make sure this does not significantly impact forward movement of content updates.

koverholt:
@joefernandez, will do. Note that this will not block any current or future PR merges, only show up a failing test. We would need to add a ruleset to enforce this, which we don't need to do at this time. So you can consider this a "soft" CI test for PRs.

PR #1009, opened 2025-11-26
Fix invalid APP_NAME example in Cloud Run deployment

PrismaCloudAI:
app name don't support "-"


koverholt (approved):
Thanks for the PR and for your contribution! I reproduced this issue and verified that the fix addresses it. LGTM!

Since ADK uses `--app_name` as a Python module name internally, the configured app name should use valid Python identifiers, otherwise it will fail with:

> Invalid app name 'capital-agent-app': must be a valid identifier consisting of letters, digits, and underscores.

PR #1010, opened 2025-11-26
fix(docs): Add missing imports/constants to sequential example

YOON331:
The sequential agents example snippet did not include the necessary imports and constants, which could cause the code to break when copied and executed as-is.


joefernandez (commented):
Thanks for the contribution. We intentionally don't include the imports because it bloats the code within the page.

Keep as-is and close.

PR #1012, opened 2025-11-27
fix: update docs to correct instructions related to running plugins with adk web

### LINKED ISSUE #1011: Plugins Documentation incorrectly suggesting it is impossible to run plugins with adk-web
In the [Plugins Documentation](https://github.com/google/adk-docs/blob/main/docs/plugins/index.md), there's a warning section indicating that: 

>  Plugins are not supported by the [ADK web interface](https://google.github.io/adk-docs/evaluate/#1-adk-web-run-evaluations-via-the-web-ui). If your ADK workflow uses Plugins, you must run your workflow without the web interface.

In another section, within the same page, there's also another reference:

> Plugins are not supported by the [ADK web interface](https://google.github.io/adk-docs/evaluate/#1-adk-web-run-evaluations-via-the-web-ui). If your ADK workflow uses Plugins, you must run your workflow without the web interface.

which I don't think it's accurate. 
In fact, it is possible to bind plugins via the `--extra_plugins` argument when running `adk web`: 

```bash
  --extra_plugins TEXT            Optional. Comma-separated list of extra plugin classes or instances to enable (e.g., my.module.MyPluginClass or my.module.my_plugin_instance).
```

I find this a bit misleading, as it gives the impression that it's impossible to run plugins via `adk web`.

gmyrianthous:
In the current version of Plugins documentation, it is mentioned that plugins can not be integrated with agents when running them via `adk web`, which this seems not to be the case. 

This PR updates the documentation, and adds instructions on how users can enable plugins when running agents locally, via `adk web` command. 

  - Fixes #1011

### DIFF
--- a/docs/plugins/index.md
+++ b/docs/plugins/index.md
@@ -30,12 +30,6 @@ Some typical applications of Plugins are as follows:
     better modularity and flexibility than Callbacks. For more details, see
     [Callbacks and Plugins for Security Guardrails](/adk-docs/safety/#callbacks-and-plugins-for-security-guardrails).
 
-!!! warning "Caution"
-    Plugins are not supported by the
-    [ADK web interface](../evaluate/#1-adk-web-run-evaluations-via-the-web-ui).
-    If your ADK workflow uses Plugins, you must run your workflow without the
-    web interface.
-
 ## How do Plugins work?
 
 An ADK Plugin extends the `BasePlugin` class and contains one or more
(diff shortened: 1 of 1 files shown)


joefernandez:
superseded by: https://github.com/google/adk-docs/pull/1290

Note: `--extra_plugins` is not actually required, per the ADK engineering team. That option is mostly used for testing.

PR #1018, opened 2025-12-02
Add Atlassian to third-party tools

koverholt:
Adds Atlassian page to third-party tools, including card, nav, and instructions on usage.


docs/tools/third-party/atlassian.md
```diff
@@ -0,0 +1,101 @@
...
+                    timeout=30,
+                ),
+            )
+        ],
+    )
+    ```
+
+!!! note
+
+    When you run this agent for the first time, a browser window will open
```
joefernandez: window opens

(avoid future tense: use present tense)

koverholt: Thanks! Fixed in 129a5e0.

PR #1028, opened 2025-12-04
Update freeplay.md

jeremysilva1098:
Replace the loom link with an embedded youtube video


koverholt (approved):
Thanks for the update! ❤️ I switched it to use YouTube embed code so that it's more obvious that it's a clickable / playable video.

PR #1047, opened 2025-12-11
fix: deleting the streaming quickstarts

kazunori279:
Deleting the streaming quickstarts as:
- They are not linked from the quickstart page anymore
- They will be replaced by the new bidi-demo in the Bidi-streaming section


kazunori279:
> Thanks for the updates!
> 
> Check the docs CI build output / failure, you'll need to:
> 
> * Remove the lines for deleted pages from nav config in `mkdocs.yml`
> * Remove the card / link on https://google.github.io/adk-docs/tutorials/

ah I didn't realize the streaming quickstart is still referenced by those index. Closing the PR.

Issue #1075, opened 2025-12-16
docs: DatabaseSessionService example is misleading for SQLite


AlejandroBlanco2001:
**Describe the bug**

In the following [documentation](https://google.github.io/adk-docs/sessions/session/#sessionservice-implementations) for the `DatabaseSessionService`, a synchronous implementation is shown for the SQLite example. However, the internal implementation of the service (built on top of SQLAlchemy) does not allow using SQLite in a synchronous approach, which causes the code to crash. This can be misleading for users.

**To Reproduce**

1. Go to https://google.github.io/adk-docs/sessions/session/#sessionservice-implementations  
2. Navigate to the `DatabaseSessionService` section  
3. Copy and paste the example code into your IDE  
4. Run the code  

**Expected behavior**

The code should run successfully without major issues and create the SQLite database.

**Screenshots**

Image extracted from the following issue: https://github.com/google/adk-python/issues/3916  

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e84dc236-cb7a-4f0b-9d58-8d7d6c624d38" />

**Versions**

- OS: Arch Linux  
- ADK version: 1.21.0  
- Python version: CPython 3.11  

**Additional context**

I tested the same scenario as the issue creator on my device and encountered the same problem. After switching to the async driver, it worked without major issues.

PR #1077, opened 2025-12-16
Specify the async driver for SQLite examples for the `DatabaseSessionService`

AlejandroBlanco2001:
During the investigation of issue https://github.com/google/adk-python/issues/3916
, I noticed that DatabaseSessionService relies on a fully asynchronous implementation for all supported dialects. However, for dialects such as PostgreSQL, this works transparently without the need to explicitly specify an async driver in the connection URL, as described in the SQLAlchemy documentation.

In contrast, most ADK examples use SQLite. SQLite does not provide the same seamless behavior as PostgreSQL in this context, and the current examples rely on the synchronous variant. As a result, following the documentation as-is leads to runtime errors, unless the developer inspects the internal implementation or arrives at the solution through trial and error.

To address this, I am adding notes to the relevant examples to explicitly explain the requirement for an async SQLite driver, and updating the examples accordingly to prevent confusion.

Closes https://github.com/google/adk-docs/issues/1075


joefernandez:
Thanks for update and sharing your findings!

PR #1085, opened 2025-12-19
Reorganize Built-In tools, Cloud, and Gemini tools pages

joefernandez:
- split up built-in-tools page content into separate, per-tool pages
- deleted built-in tools page
- relocated and broke up google-cloud-tools.md file into per-tool pages
- created redirects for moved and deleted pages
- added new navigation entries
- fixed broken internal links to moved pages

PR #1092, opened 2025-12-19
Re-enable agents dir in llms.txt generation

koverholt:
Re-adding `agents/*.md` in llms.txt docs config that was blocked in #1085 and #1091 and was resulting in build failures such as https://github.com/google/adk-docs/actions/runs/20378285966/job/58566866306. The upstream policy that was blocking this should be resolved now.

After this is merged, all docs under the `docs/agents/` dir should be included in the autogenerated llms.txt and the corresponding `.md` files should get published successfully.


koverholt:
Confirmed that the docs build & publish action is working as expected now with the updated rule/policy allowlisting for this repo.

PR #1098, opened 2025-12-26
Describe how to configure FileArtifactStorage from CLI

Spikhalskiy:
This PR describes how to configure `FileArtifactStorage` usage from CLI.

Closes google/adk-python#4328


koverholt:
Thanks for the PR! This description actually gets generated from the docstrings in the Python library:

- https://github.com/google/adk-python/blob/main/src/google/adk/cli/cli_deploy.py
- https://github.com/google/adk-python/blob/main/src/google/adk/cli/cli_tools_click.py

We generate API/CLI docs from the Python library and then vendor them into this repo to be served with the rest of the ADK docs. So if we were to change the HTML files downstream in this docs repo, they would get overwritten in the next API/CLI docs update.

Could you open an issue/PR in the upstream `adk-python` repo to fix the docstrings there?

https://github.com/google/adk-python/pulls

PR #1099, opened 2025-12-29
Fix syntax for AgentCard initialization

tom-010:
The current version looked like a mix from JSON and Python and was not valid Python code.


koverholt (approved):
I tested and the changes are working as expected. LGTM. Thanks for the formatting fixes! ❤️

PR #1104, opened 2026-01-02
Fix state update

PangLuo:
This PR fixes a bug where session state updates were not persisting. The issue was that `get_session()` returns a copy of the session object, so modifications to `current_session.state["topic"]` were not reflected in the actual session. When `runner.run_async()` retrieves the session later, it would still see the old topic value.

The new code `current_session = session_service.sessions[APP_NAME][USER_ID][SESSION_ID]` `current_session.state["topic"] = user_input_topic` will really update the state.


koverholt (approved):
I tested this and with the changes it's updating the state correctly and working as expected. LGTM. Thanks for your contribution! 🎉

PR #1105, opened 2026-01-05
fix: infinite loop in LoopAgent exit tool example

sultanfariz:
**What**
Added skip_summarization=True to the exit_loop tool in the LoopAgent documentation example.

**Why**
Without skip_summarization=True, the exit_loop tool causes an infinite loop because the agent tries to summarize the tool call result before escalating. This creates a cycle where the loop never actually exits.
This fix ensures the loop terminates immediately when exit_loop is called, which is the expected behavior for breaking out of a LoopAgent iteration.

**Related Issue**
Fixes infinite loop issue [#2988](https://github.com/google/adk-python/issues/2988)


joefernandez:
Thanks for the fix!

PR #1134, opened 2026-01-12
chore: add a new gha to build and host an open PR

### LINKED ISSUE #614: Have Pull Requests actually build the site and temporarily host it somewhere live to allow reviewers to easily preview changes
**Is your feature request related to a problem? Please describe.**

It would be cool if docs changes from PRs in this repo would actually build the site and temporarily host it somewhere live to allow reviewers to easily preview changes.

This is fairly common in many "web" open source projects. There are a 3rd party services offering solutions for this. We would have to see if Google has something like this, but my guess is that in Firebase there is stuff for doing this?

Sita04:
**New PR preview workflow:**

* Adds a `.github/workflows/pr-preview.yaml` workflow that builds and deploys a MkDocs preview site for pull requests labeled with "build-preview", using the `rossjrw/pr-preview-action` to manage the preview lifecycle.
* Includes a cleanup job to automatically remove the preview when the PR is closed or the "build-preview" label is removed.

How to use this:
1. Create a PR
2. Add the label 'build-preview' to the PR to host the branch

Fixes #614

### DIFF
--- a/docs/observability/bigquery-agent-analytics.md
+++ b/docs/observability/bigquery-agent-analytics.md
@@ -70,6 +70,17 @@ from google.adk.agents import Agent
 from google.adk.models.google_llm import Gemini
 from google.adk.tools.bigquery import BigQueryToolset, BigQueryCredentialsConfig
 
+
+# --- OpenTelemetry Initialization (Optional) ---
+# Recommended for enabling distributed tracing (populates trace_id, span_id).
+# If not configured, the plugin uses internal UUIDs for span correlation.
+try:
+    from opentelemetry import trace
+    from opentelemetry.sdk.trace import TracerProvider
+    trace.set_tracer_provider(TracerProvider())
+except ImportError:
+    pass # OpenTelemetry is optional
+
 # --- Configuration ---
 PROJECT_ID = os.environ.get("GOOGLE_CLOUD_PROJECT", "your-gcp-project-id")
 DATASET_ID = os.environ.get("BIG_QUERY_DATASET_ID", "your-big-query-dataset-id")
@@ -138,6 +149,16 @@ FROM `your-gcp-project-id.your-big-query-dataset-id.agent_events_v2`
 ORDER BY timestamp DESC
 LIMIT 20;
 ```
+
+```
+
+## Tracing and Observability
+
+The plugin supports **OpenTelemetry** for distributed tracing.
+
+- **Automatic Span Management**: The plugin automatically generates spans for Agent execution, LLM calls, and Tool executions.
+- **OpenTelemetry Integration**: If an OpenTelemetry `TracerProvider` is configured (as shown in the example above), the plugin will use valid OTel spans, populating `trace_id`, `span_id`, and `parent_span_id` with standard OTel identifiers. This allows you to correlate agent logs with other services in your distributed system.
+- **Fallback Mechanism**: If OpenTelemetry is not installed or configured, the plugin automatically falls back to generating internal UUIDs for spans and uses the `invocation_id` as the trace ID. This ensures that the parent-child hierarchy (Agent -> Span -> Tool/LLM) is *always* preserved in the BigQuery logs, even without a full OTel setup.
 
 ## Configuration options
 


koverholt:
Thanks for this! I experimented with the same plugin before (and a few other options).

My one drawback was that it will publish preview docs to the same docs endpoint as the main docs, so how can we ensure that those preview docs don't get indexed and end up in search results when they are not yet merged / published? Perhaps by using robots.txt, or publishing to a different endpoint, or something else?

.github/workflows/pr-preview.yaml
```diff
@@ -0,0 +1,50 @@
...
+  contents: write
+  pull-requests: write
+concurrency: preview-${{ github.ref }}
+jobs:
+  preview:
+    runs-on: ubuntu-latest
+    if: | # builds the open PR whenever it sees the label 'build-preview'
+      github.event.action != 'closed' &&
+      github.event.action != 'unlabeled' &&
+      contains(github.event.pull_request.labels.*.name, 'build-preview')
```
koverholt, line 31: If we do enable / merge this functionality, I would suggest having build preview for all PRs rather than having to add a label.

joefernandez:
> My one drawback was that it will publish preview docs to the same docs endpoint as the main docs, so how can we ensure that those preview docs don't get indexed and end up in search results when they are not yet merged / published? Perhaps by using robots.txt, or publishing to a different endpoint, or something else?

We don't want the preview content to be shown with the approved and merged content. That will create user confusion and potentially disrupt the browsing experience for the currently published content (especially for PRs where we are reorganizing content and temporarily breaking the browsing experience). Can we push the preview versions to a separate Preview site?

Sita04:
@koverholt Excellent point! I've added the robots.txt to disallow indexing. Hope this is sufficient to prevent crawling.

@joefernandez All the PRs will be hosted under the pr-preview directory: https://google.github.io/adk-docs/pr-preview/pr-{number}/

"/pr-preview/pr-{number}/" will not be present for the actual hosted docs. But I do understand the concern of having it in the same domain. Let me check if we can change the domain.

koverholt:
> Can we push the preview versions to a separate Preview site?

The good news is that the `pr-preview-action` supports a different base URL for previews, but the bad news is there is not an easy way to do this with GitHub pages in terms of hosting more than one docs endpoint, like you can with Netlify or Vercel or other static site hosts. We would need to set up a separate static host for this using Firebase or Cloudflare Pages (like Jules does for PR previews) and manage it ourselves, which is sub-optimal.

@Sita04, are there other approaches or plugins that might make this easier, or is our limiting factor that GitHub Pages will only serve one docs endpoint in a given repo (and thus we would need a separate static hosting pipeline in any case)?

.github/workflows/pr-preview.yaml
Sita04: I thought about this as well. If we enable this by default, then do you think we might end up with tens/ hundreds of open hosted sites, potentially even for minor changes?

Couldn't weight in which direction to take, so thought if someone wanted to visualize, then it could be as simple as adding a label?

Sita04:
> are there other approaches or plugins that might make this easier, or is our limiting factor that GitHub Pages will only serve one docs endpoint in a given repo?

You are correct - only one endpoint per repo. And +100 to not creating overhead by adding Firebase/ Netlify. 

This leaves with one other workaround if we want domain separation - we can create an empty adk-docs-previews repo and use that domain to host all previews. 

edit: On a second thought, this doesn't entirely solve our problem though. We still have to add robots.txt to the other public repo and there's a potential to mess with the SEO as well.

.github/workflows/pr-preview.yaml
koverholt: I like that this action cleans up the PR after it's closed or merged, so in theory it would be one hosted preview site, with one URL path for each preview on the same base URL / site. And previews for all PRs would help authors review their rendered changes since folks don't always spin up mkdocs locally in dev and preview the rendered versions. So I am +1 to having it on for every PR by default, and ideally the plugin would handle the permutations and cleanup cleanly like this plugin does.

koverholt:
> This leaves with one other workaround if we want domain separation - we can create an empty adk-docs-previews repo and use that domain to host all previews.

This was also the same conclusion I came to when I was experimenting with this before. However the "preview" repo would need to be always synced with the adk-docs repo (including PRs), and the action would need to run from the "preview" repo so that it publishes into the gh-pages branch in that repo, correct? In the end, the coupling of these two repos would need to be more complex than I originally thought rather than just an empty repo / placeholder for the domain.

Sita04:
> However the "preview" repo would need to be always synced with the adk-docs repo (including PRs), and the action would need to run from the "preview" repo so that it publishes into the gh-pages branch in that repo, correct?

I was thinking more on the lines of: "The GHA runs in the `adk-docs` repo and only the final html pages will be git pushed to the other repo's gh-pages". So, essentially the docs-preview repo will be empty except for the rendered html pages in different branches. wdyt?

Since Google has Enterprise version, we might even be able to create a privately hosted Github pages for review.

koverholt:
> The GHA runs in the adk-docs repo and only the final html pages will be git pushed to the other repo's gh-pages.

Got it, I see that option in the action's settings now. This pipeline makes the most sense to me since in theory we would only need to change the `deploy-repository` and `pages-base-url` setting for this to work. I think this is worth setting up and seeing if it behaves as we expect. And we could still have robots.txt just to guard the preview-repo's site and directory from getting indexed at all.

.github/workflows/pr-preview.yaml
joefernandez: > If we do enable / merge this functionality, I would suggest having build preview for all PRs rather than having to add a label.

I'm with Sita on this: Proliferation of preview content is a concern, because there are A LOT of open PRs. Also, we only really need the preview when we are doing a review (or the author is self reviewing), so as-needed, based on a label, makes more sense.

koverholt: There is no real overhead to having a preview site for each build, since they are namespaced and all behind the root prefix in `robots.txt`. And it's a nice pattern to have the bot respond immediately to a PR with the preview link.

But I'm okay either way, and if we do go the label route, would be nice if a bot posted a message in a new PR about which label to add to enable previews! 😁

docs/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Disallow: /pr-preview/
```
koverholt: Add newline at end of file.

.github/workflows/pr-preview.yaml
```diff
@@ -0,0 +1,50 @@
...
+      (github.event.action == 'unlabeled' && github.event.label.name == 'build-preview')
+    steps:
+      - uses: actions/checkout@v4
+      - name: Cleanup Preview
+        uses: rossjrw/pr-preview-action@v1
+        with:
+          source-dir: ./site
+          umbrella-dir: pr-preview
+          action: remove
+          token: ${{ secrets.GITHUB_TOKEN }}
```
koverholt: Add newline at end of file.

docs/robots.txt
Sita04: Done.

.github/workflows/pr-preview.yaml
Sita04: Done.

.github/workflows/pr-preview.yaml
Sita04: Let's test it out with the secret "label" handshake for the time being and open it up to all PRs, if things work smoothly! 

Good idea on the automated bot response. Will follow up on that.

koverholt:
Also tested things in this PR and the build preview and cleanup are working as expected. Thanks!

PR #1141, opened 2026-01-14
docs: Drop GCS bucket from the instructions for deploying AgentEngine

yeesian:
It isn't required


joefernandez:
@yeesian - Have you tested this change and confirmed that deployment works without the GCP bucket? Agent Engine deployment is a hotly viewed set of instructions. We should not update this unless the change has been tested and confirmed.

yeesian:
> @yeesian - Have you tested this change and confirmed that deployment works without the GCP bucket? Agent Engine deployment is a hotly viewed set of instructions. We should not update this unless the change has been tested and confirmed.

Yes it is tested and confirmed to work without the GCP bucket. The GCP bucket argument for deploying to AgentEngine will be deprecated in future releases of Google ADK.

joefernandez (approved):

PR #1144, opened 2026-01-16
docs: (reorg) moving page for Vertex AI express mode

joefernandez:
express mode supports sessions and memory, AND it ALSO supports other features, so moving it to a more appropriate location


mkdocs.yml
```diff
@@ -309,7 +311,7 @@ nav:
         - Rewind sessions: sessions/rewind.md
       - State: sessions/state.md
       - Memory: sessions/memory.md
-      - Vertex AI Express Mode: sessions/express-mode.md
+
```
koverholt: Remove extra space in nav config.

docs/tools/google-cloud/express-mode.md
```diff
@@ -0,0 +1,138 @@
...
+    )
+
+    agent_engine = client.agent_engines.create(
+      config={
+        "display_name": "Demo Agent Engine",
+        "description": "Agent Engine for Session and Memory",
+      })
+    ```
+
+3. Replace `YOUR_AGENT_ENGINE_DISPLAY_NAME` and `YOUR_AGENT_ENGINE_DESCRIPTION` with your use case.
```
koverholt: Happened before this PR, but there are no PLACEHOLDER_VARS in the above code snippet to replace.

joefernandez: thank for catching this. removing this step, because it's not actually necessary given the code snippet

mkdocs.yml
joefernandez: done

PR #1164, opened 2026-01-23
docs: document built-in data-agent tools

Genesis929:


docs/tools/google-cloud/data-agent.md
```diff
@@ -0,0 +1,17 @@
+# Data Agent tool for ADK
```
shobsi: [nit] tools, same should be done for BigQuery tools

Genesis929: Updated for data agent tools, maybe bigquery tools name update should be in another PR?

docs/tools/google-cloud/data-agent.md
```diff
@@ -0,0 +1,29 @@
+# Data Agents tools for ADK
+
+<div class="language-support-tag">
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v1.1.0</span>
```
joefernandez: version is not correct. This class was released in 1.23.0 

https://github.com/google/adk-python/commit/82fa10b71e037b565cb407c82e9e908432dab0ff

Genesis929: Updated, thanks!

PR #1165, opened 2026-01-23
Add Daytona to third-party tools

mislavivanda:
This PR adds [Daytona](https://www.daytona.io/) to the third-party tools sections.

 **Changes**:
  - Added Daytona card to the third-party tools index pages and sidebar navigation
  - Added `daytona.md` documentation page with usage instructions, available tools, and relevant links

Daytona provides secure sandbox environments for ADK agents to execute code, run shell commands, and manage files in isolated environments.


koverholt:
Thanks for the PR and the great ADK integration/plugin! ❤️ 

As we discussed, we'll wait until https://github.com/daytonaio/daytona/pull/3469 is merged before we review and merge this PR.

docs/tools/index.md
```diff
@@ -303,6 +303,16 @@ Check out the following pre-built tools that you can use with ADK agents:
...
+  <a href="/adk-docs/tools/third-party/daytona/" class="tool-card">
+    <div class="tool-card-image-wrapper">
+      <img src="/adk-docs/assets/plugins-daytona.png" alt="Daytona">
+    </div>
+    <div class="tool-card-content">
+      <h3>Daytona</h3>
+      <p>Execute code, run shell commands, and manage files in secure sandboxes</p>
+    </div>
+  </a>
+
```
koverholt: Move this card up to be sorted alphabetically among the other cards.

docs/tools/third-party/index.md
```diff
@@ -149,4 +149,14 @@ Check out the following third-party tools that you can use with ADK agents:
...
+  <a href="/adk-docs/tools/third-party/daytona/" class="tool-card">
+    <div class="tool-card-image-wrapper">
+      <img src="/adk-docs/assets/plugins-daytona.png" alt="Daytona">
+    </div>
+    <div class="tool-card-content">
+      <h3>Daytona</h3>
+      <p>Execute code, run shell commands, and manage files in secure sandboxes</p>
+    </div>
+  </a>
+
```
koverholt: Move this card up to be sorted alphabetically among the other cards.

docs/tools/third-party/daytona.md
```diff
@@ -0,0 +1,68 @@
+# Daytona
+
+The [Daytona ADK plugin](https://pypi.org/project/daytona-adk/) connects your ADK
```
koverholt: Suggest linking to the GitHub repo instead: https://github.com/daytonaio/daytona-adk-plugin

docs/tools/third-party/daytona.md
```diff
@@ -0,0 +1,68 @@
...
+pip install daytona-adk
+```
+
+## Use with agent
+
+```python
+from daytona_adk import DaytonaPlugin
+from google.adk.agents import Agent
+
+plugin = DaytonaPlugin()
```
koverholt: In this sample code, suggest adding one of the methods of passing your API key to the plugin, perhaps using `plugin = DaytonaPlugin(api_key="your-daytona-api-key")` in this case.

No need to mention all possible methods, just makes it easier for users to copy/paste and get started with a self-contained code sample.

docs/tools/third-party/daytona.md
```diff
@@ -0,0 +1,68 @@
...
+
+root_agent = Agent(
+    model="gemini-2.5-pro",
+    name="sandbox_agent",
+    instruction="Help users execute code and commands in a secure sandbox",
+    tools=plugin.get_tools(),
+)
```
koverholt, line 46: I tested an agent running code in a sandbox and it works great!

docs/tools/third-party/index.md
mislavivanda: Fixed

docs/tools/index.md
mislavivanda: Fixed

docs/tools/third-party/daytona.md
mislavivanda: Fixed

docs/tools/third-party/daytona.md
mislavivanda: The API key can be passed explicitly via the `api_key` parameter; however, if it’s not provided, the Daytona SDK will automatically attempt to read it from the environment variable. As such, the current code snippet represents the minimal example required to use this plugin.

Given this, would you still recommend explicitly setting the `api_key` value, or should we keep the example as it is?

koverholt: Feel free to show the preferred method, but let's list all explicit steps for the selected method.

If we go with the env var approach: 1) add a doc step to define the env var with an API key, 2) add a comment to the line with `plugin = DaytonaPlugin()` that it will pick up the API key from this env var.

or,

Use the form `plugin = DaytonaPlugin(api_key="your-daytona-api-key")`, and add a comment that users can also specify this via an env var.

Otherwise in the current version of the page there is no explicit step for a user to configure the env var, so they would need to figure that our on their own.

mislavivanda: Agree. Applied your suggestion.

PR #1196, opened 2026-01-28
Revise TypeScript setup instructions and dependencies

LukeSchlangen:
Updated Node.js and npm version requirements, adjusted project setup instructions, and modified package.json structure.

Might depend on https://github.com/google/adk-js/pull/46

### DIFF
--- a/docs/get-started/typescript.md
+++ b/docs/get-started/typescript.md
@@ -72,7 +72,7 @@ const getCurrentTime = new FunctionTool({
 
 export const rootAgent = new LlmAgent({
   name: 'hello_time_agent',
-  model: 'gemini-3.0-flash',
+  model: 'gemini-2.5-flash',
   description: 'Tells the current time in a specified city.',
   instruction: `You are a helpful assistant that tells the current time in a city.
                 Use the 'getCurrentTime' tool for this purpose.`,


LukeSchlangen:
What if we remove the explicit zod dependency? That's what we had in the original get-started, so I think it should still work? (Need to test it.)

Then we can do a separate PR for the zod update when that's ready?

LukeSchlangen:
Tested! It works. I removed Zod 4. I think it's good to merge if that works for you!

docs/get-started/typescript.md
```diff
@@ -3,136 +3,103 @@
...
+  execute: ({city}) => {
+    return {status: 'success', report: `The current time in ${city} is 10:30 AM`};
+  },
+});
 
-```console
-npx tsc
+export const rootAgent = new LlmAgent({
+  name: 'hello_time_agent',
+  model: 'gemini-3.0-flash',
```
koverholt: `gemini-3.0-flash` is not a valid model string. Should be either `gemini-3-flash-preview` or `gemini-2.5-flash`. Gemini 3 doesn't currently have a free tier in Gemini API, so 2.5 Flash would be more accessible to users.

docs/get-started/typescript.md
```diff
@@ -3,136 +3,103 @@
...
-  },
   "dependencies": {
-    "@google/adk": "^0.2.0",
+    "@google/adk": "^0.2.0"
+  },
+  "devDependencies": {
     "@google/adk-devtools": "^0.2.0"
   }
 }
 ```
```
koverholt: I'm wondering if we need to show `package.json` on this page for verification. If someone copy/pastes it, it does currently include an old version number for ADK JS, since the latest version is at 0.2.5 now. Whereas if we remove this we don't have to keep bumping the packages to the latest versions with each release. Thoughts?

docs/get-started/typescript.md
koverholt: Should we change this to `npx adk run agent.ts`, or leave as is?

koverholt: Should we change this to `npx adk web`, or leave as is?

koverholt:
And once those comments are addressed, I will pass to @joefernandez for review since this is a heavily visited page / entrypoint.

docs/get-started/typescript.md
LukeSchlangen: Interesting. It ran sucessfully with `gemini-3.0-flash`, but good call on the pricing. I put it back.

docs/get-started/typescript.md
LukeSchlangen: Good call. I pulled it out.

docs/get-started/typescript.md
LukeSchlangen: I didn't know that was possible. That's very cool updated!

LukeSchlangen: I didn't know that was possible. That's very cool updated!

koverholt:
Thanks for the updates! There is an issue with the CLA check due to commits from a different GitHub account. Can you rebase to switch authorship of those commits, or otherwise fix the failing CLA check? Might have happened due to misconfigured git config user / email.

https://github.com/google/adk-docs/pull/1196/checks?check_run_id=61911522036

LukeSchlangen:
I wasn't smart enough to figure out how to do that, so I'll close this one and open a new PR with the correct email: https://github.com/google/adk-docs/pull/1201

PR #1210, opened 2026-01-30
docs: add infrastructure for integrations catalog

joefernandez:
task: b/483452815


scripts/integrations.py
```diff
@@ -0,0 +1,92 @@
+import os
```
koverholt: Need to add a license header here since that check is failing.

koverholt (approved):
Discussed a few topics, noting them here for posterity.

1. Which version of mkdocs-macros-plugin to use?
2. For the top-level page / folder (future, not yet reflected in this PR), should we use `/integrations` instead of `/catalog`?
3. Suggested putting the Python script at `scripts/integrations.py` in the repo root and using `module_name: scripts/integrations` in the config for better organization and future scripts.
4. Took me a second to understand why BQ Agent Analytics was missing yet other new tools were added until I saw the glob on `tools/google-cloud/*.md`. Got it now!
5. Should probably use a log handler for plugins in the Python script so that any missing descriptions or images or such will fail the docs build instead of failing silently.

PR LGTM and approving for when you're ready to merge!

scripts/integrations.py
```diff
@@ -0,0 +1,93 @@
...
```
koverholt: The os import here is unused, safe to remove.

scripts/integrations.py
```diff
@@ -0,0 +1,93 @@
...
+                        if line.startswith('# '):
+                            title = line[2:].strip()
+                            break
+                # Fallback to filename
+                if not title:
+                    title = file_path.stem.replace('-', ' ').title()
+
+                description = frontmatter.get('catalog_description',
+                    frontmatter.get('description', ''))
+                icon = frontmatter.get('catalog_icon',
```
koverholt: Note for a future PR, can add checks and error handling for missing icons and/or descriptions.

koverholt (approved):
Updates LGTM!

Only "blocking" issue in the PR is the license header (and some conflicts to deal with).

Left some other comments that can be handled here or in a future PR. Thanks!

scripts/integrations.py
joefernandez: done

scripts/integrations.py
joefernandez: removed.

scripts/integrations.py
joefernandez: sure. for now, we have some reasonable default behavior

PR #1212, opened 2026-01-31
docs: refactor loop agent example - clean imports and test code

adityamulik:
Refactored the code to use the imports in the example as it would become an easier copy paste for newcomers to try the multi-workflow agent without refactoring imports or making changes. Commented the rest of the code by adding comments for local testing for the maintainers of this repository.

Also, I added the logic of adding the state (initial_topic) so it becomes easier to copy and test the code right away. 

Please let me know if theres any feedback or if this change is not needed.


adityamulik:
Tested using ADK Web
image

koverholt (approved):
Thanks for the PR and for your contribution! I updated the Gemini model to 2.5 Flash along with some additional overdue cleanup on this script. Thanks for the fixes and making the example self-contained and easier for readers to run.

PR #1219, opened 2026-02-03
chore: use Netlify for PR Previews

Sita04:
This pull request removes the GitHub Actions workflow for PR previews and updates the project configuration to use Netlify for building and deploying documentation. It also removes restrictions on web crawlers for PR preview content.

**CI/CD and Deployment changes:**

* Deleted the `.github/workflows/pr-preview.yaml` file, removing the automated PR preview workflow based on GitHub Actions.
* Added a Netlify configuration in `netlify.toml` to build the documentation using MkDocs and Python 3.12, replacing the previous preview deployment workflow.

PR #1220, opened 2026-02-03
chore: test previews - do not merge

Sita04:


koverholt:
We moved from `pr-preview-action` to Netlify builds in #1219 and #1278. PR preview builds are working as expected now. Thanks!

PR #1250, opened 2026-02-05
Update CLI reference docs for ADK 1.24.0; add generation script

koverholt:
This PR:

- Updates CLI reference docs for ADK Python 1.24.0 release
- Adds script for generating CLI docs
- Simplifies CLI reference docs from 2 pages to 1
- Shows ADK version in intro text

### DIFF
--- a/docs/get-started/java.md
+++ b/docs/get-started/java.md
@@ -102,8 +102,8 @@ An ADK agent project requires this dependency in your
 <dependencies>
     <dependency>
         <groupId>com.google.adk</groupId>
-        <artifactId>adk-core</artifactId>
-        <version>0.3.0</version>
+        <artifactId>google-adk</artifactId>
+        <version>0.5.0</version>
     </dependency>
 </dependencies>
 ```


docs/api-reference/cli/genindex.html
```diff
@@ -4,11 +4,11 @@
   <head>
     <meta charset="utf-8" />
```
joefernandez, line 5: need to add in the Google Analytics tags, here:


  <script async src="https://www.googletagmanager.com/gtag/js?id=G-DKHZS27PHP"></script>
  <script>
    window.dataLayer = window.dataLayer || [];
    function gtag(){dataLayer.push(arguments);}
    gtag('js', new Date());
    gtag('config', 'G-DKHZS27PHP');
  </script>

docs/api-reference/cli/search.html
```diff
@@ -4,12 +4,12 @@
   <head>
     <meta charset="utf-8" />
```
joefernandez, line 5: add the GA tag:


  <script async src="https://www.googletagmanager.com/gtag/js?id=G-DKHZS27PHP"></script>
  <script>
    window.dataLayer = window.dataLayer || [];
    function gtag(){dataLayer.push(arguments);}
    gtag('js', new Date());
    gtag('config', 'G-DKHZS27PHP');
  </script>

tools/python-cli-docs/generate.sh
```diff
@@ -0,0 +1,96 @@
...
+   :prog: adk
+   :nested: full
+EOF
+
+# Build HTML
+echo "Building HTML..."
+make html
+
+popd > /dev/null || exit 1
+
```
joefernandez, line 107: add function for injecting the GA tags into the html files

tools/python-cli-docs/generate.sh
joefernandez, line 1: this probably should be in the <root>/tools/ directories, rather than /scripts/

the split should be:
/scripts/ - website runtime code (will break the site if not working)
/tools/ - code run separately the runtime of the docs site

koverholt: Good idea, moved in 3519193.

tools/python-cli-docs/generate.sh
koverholt: Done, added in 3519193 to all generated HTML using Sphinx templating.

docs/api-reference/cli/genindex.html
koverholt: Done, added in 3519193 to all generated HTML using Sphinx templating, and confirmed it is included in the generated HTML files in this PR.

docs/api-reference/cli/search.html
koverholt: Done, added in 3519193 to all generated HTML using Sphinx templating, and confirmed it is included in the generated HTML files in this PR.

PR #1269, opened 2026-02-06
After Agent Call back behavior: change replace to append

joefernandez:
- Context: https://github.com/google/adk-docs/pull/1255
- requestor: @sasha-gitg


koverholt (approved):
Verified the behavior in base_agent.py appends instead of replaces. LGTM.

- https://github.com/google/adk-python/blob/main/src/google/adk/agents/base_agent.py#L150-L163
- https://github.com/google/adk-python/blob/main/src/google/adk/agents/base_agent.py#L489-L548

PR #1287, opened 2026-02-11
Fix grammar in event loop documentation

SANOOPKV:
Correct grammatical structure in event loop explanation.


joefernandez:
The sentence needed a bit more work. Updated and merged. Thanks for the contribution!

PR #1290, opened 2026-02-11
docs: MCP pages, add tip pointer to Tools and integrations

joefernandez:
Point various pages at new Integrations catalog

Bonus fix:
Fixes: https://github.com/google/adk-docs/issues/1011
Plugin / Web compatibility change info: b/483772794


joefernandez:
> Should we add this tip to https://google.github.io/adk-docs/mcp/ as well?
Yes! Added.

PR #1313, opened 2026-02-13
docs: Add documentation for Skills support

joefernandez:


mkdocs.yml
```diff
@@ -304,6 +304,8 @@ nav:
       - OpenAPI tools: tools-custom/openapi-tools.md
       - Authentication: tools-custom/authentication.md
       - Tool limitations: tools/limitations.md
+    - Skills for Agents:
```
koverholt, line 308: Seems odd to put this in top-level nav, should we put this under "Build your Agent" instead? Or some other existing top level?

docs/skills/index.md
```diff
@@ -0,0 +1,154 @@
...
+!!! example "Experimental"
+    The Skills feature is experimental and has some
+    [known limitations](#known-limitations). We welcome your
+    [feedback](https://github.com/google/adk-python/issues/new?template=feature_request.md&labels=skills)!
+
+## Get started
+
+Use the `SkillToolset` class to include one or more Skills in your agent
+definition and then add to your agent's tools list. You can define a
+[Skill in code](#inline-skill),
```
koverholt: Mismatch between this anchor link (inline-skill) and the actual anchor definition (inline-skills).

docs/skills/index.md
```diff
@@ -0,0 +1,154 @@
...
+<div class="language-support-tag">
+    <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v1.25.0</span><span class="lst-preview">Experimental</span>
+</div>
+
+An agent ***Skill*** is a self-contained unit of functionality that an ADK agent
+can use to perform a specific task. An agent Skill encapsulates the necessary
+instructions, resources, and tools required for a task, based on the
+[Agent Skill specification](https://agentskills.io/specification).
+The structure of a Skill allows it to be loaded incrementally to minimize the
+impact on operating context window of the agent.
```
koverholt: "on the operating" ?

mkdocs.yml
joefernandez: Good question. Skills are functionally equivalent to a tool, so putting them in parallel with Tools and Integrations and Custom Tools is reasonable. Placing this under Build your Agent isn't inline with the long-term information architecture of that section.

Leaving this where it is, for now. We can always move it later.

docs/skills/index.md
joefernandez: good catch! thanks.

docs/skills/index.md
joefernandez: fixed.

PR #1320, opened 2026-02-17
docs: Add "ADK" to site title for findability

joefernandez:


mkdocs.yml
```diff
@@ -1,8 +1,8 @@
 # Project information
-site_name: Agent Development Kit
+site_name: Agent Development Kit (ADK)
 site_url: https://google.github.io/adk-docs
 site_description: >-
-  Build powerful multi-agent systems with Agent Development Kit
+  Build powerful multi-agent systems with Google Agent Development Kit (ADK)
```
koverholt: Suggest removing the "Google" here since it's not part of the official name. We used to have this directive (use "ADK" not "Google ADK") in the internal style guide, but it seems to have been left off in recent versions.

```suggestion
  Build powerful multi-agent systems with Agent Development Kit (ADK)
```

joefernandez: done

PR #1378, opened 2026-02-27
Update CLI reference docs for ADK 1.26.0

koverholt:


docs/api-reference/cli/genindex.html
```diff
@@ -4,11 +4,11 @@
   <head>
```
joefernandez, line 4: Please add the Google Analytics tags to the HTML files.

docs/api-reference/cli/index.html
```diff
@@ -5,11 +5,11 @@
     <meta charset="utf-8" />
```
joefernandez, line 5: Please add the Google Analytics tags to the HTML files.

docs/api-reference/cli/genindex.html
koverholt: GA tags are present based on the [template](https://github.com/google/adk-docs/blob/main/tools/python-cli-docs/generate.sh#L75-L82), they are just not showing up in the diff view in GitHub since nothing changes w/ them.

docs/api-reference/cli/index.html
koverholt: Thanks! See previous comment.

PR #1384, opened 2026-03-02
Add example code for OpenAPI tools

Astroa7m:
Added an example code for OpenAPI tools.


docs/tools-custom/openapi-tools.md
```diff
@@ -76,9 +76,236 @@ Follow these steps to integrate an OpenAPI spec into your agent:
 ## Example
 
 This example demonstrates generating tools from a simple Pet Store OpenAPI spec (using `httpbin.org` for mock responses) and interacting with them via an agent.
+```
```
joefernandez, line 79: Thanks for contributing! However, the following syntax includes the code in the page, so directly adding it to the page is unnecessary:

```python title="openapi_example.py"
    --8<-- "examples/python/snippets/tools/openapi_tool.py"

PR #1433, opened 2026-03-13
fix: Add Technical Overview page link to GetStarted page.

msampathkumar:
Technical overview page is part of get started page (https://github.com/google/adk-docs/tree/main/docs/get-started) and provides insights on ADK's working concepts. Somehow this page is not included in the docs "Get Started" page.

fix: In this PR, I propose add missing URL link to Technical Concepts.

This is valuable for users(senior leads/engineers) who would like know concepts before deep diving into code.


joefernandez (approved):
Thanks for the update.

PR #1455, opened 2026-03-17
add site owner verification file (b/493636107)

haomengchao:
This change will allow Vertex AI Searchto index the docs link. We need this file to perform Google Search site owner verification.


docs/googledee034e68b1cf515.html
joefernandez, line 1: This file is in the wrong location if you want it to be accessible from the ADK docs website:
Expected location: https://google.github.io/adk-docs/googledee034e68b1cf515.html

You need to move this file to the website content directory
https://github.com/google/adk-docs/tree/main/docs

PR #1467, opened 2026-03-18
Added missing java snippets to function-tools.md

happyhuman:
Also replaced dummy stock data with the real yahoo finance API.


examples/java/snippets/src/main/java/tools/StockPriceAgent.java
```diff
@@ -11,24 +11,17 @@
...
-  // NOTE: This is a MOCK implementation. In a real Java application,
-  // you would use a financial data API or library.
-  private static final Map<String, Double> mockStockPrices = new HashMap<>();
-
-  static {
-    mockStockPrices.put("GOOG", 1.0);
-    mockStockPrices.put("AAPL", 1.0);
-    mockStockPrices.put("MSFT", 1.0);
-  }
+  // No longer using mock stock data - we fetch it live!
```
joefernandez, line 24: This is cool, but I'm not sure it's a good idea to make this sample dependent on 3P API. I'll leave up to you, though.

PR #1468, opened 2026-03-18
Added the missing java samples to compaction.md

happyhuman:


docs/context/compaction.md
```diff
@@ -25,19 +25,37 @@ Configuration setting to the App object of your workflow. As part of the
...
+        ),
+    )
+    ```
+
+=== "Java"
+
+    ```java
+    import com.google.adk.apps.App;
+    import com.google.adk.summarizer.EventsCompactionConfig;
+
```
bnepe-lang, line 49: Should we also add a Java supported tag? 

`<span class="lst-supported">Supported in ADK</span><span class="lst-java">Java v0.X.0</span>`

happyhuman: Good catch. Added it.

PR #1478, opened 2026-03-20
Add documentation for ADK optimization features

KPJoshi:
This PR adds documentation for ADK's new agent optimization feature. The documentation walks through an example that uses the new `adk optimize` command, then introduces the inbuilt optimization-related classes, and finally discusses extensibility.


mkdocs.yml
```diff
@@ -335,6 +335,8 @@ nav:
       - evaluate/index.md
       - Criteria: evaluate/criteria.md
       - User Simulation: evaluate/user-sim.md
+    - Optimization:
```
joefernandez: Because this feature is dependent on evaluation, let's list it under Evaluation, but let's leave the page in the optimization/ directory, because I can imagine a feature that optimizes without first doing evals.

joefernandez: updated.

joefernandez (approved):

PR #1485, opened 2026-03-23
Integrations: Add Resilience tag and clean up catalog tags

koverholt:
## Summary

- Add new `resilience` tag to Temporal, Restate, and Reflect and Retry integrations
- Remove `evaluation` tag for consistency since only one integration was using that tag


docs/integrations/reflect-and-retry.md
```diff
@@ -2,7 +2,7 @@
 catalog_title: Reflect and Retry Plugin
 catalog_description: Automatically retry tool calls that fail
 catalog_icon: /adk-docs/integrations/assets/adk.png
-catalog_tags: ["google"]
+catalog_tags: ["google", "durability"]
```
drewhoskins-temporal: Not sure this word quite fits, and it's also non-standard.  Would propose one of two possible fixes here.  I'm just getting up to speed on reflect and retry, so take this with a grain of salt:

* Make a generic "reliability" tag encompassing reflect and retry, restate, and temporal.
* Make a "durable execution" tag (an industry-standard term--Google "durable execution agents" and see langchain, temporal, pydantic AI, restate, dbos, inngest, microsoft, and several more)

koverholt: Agreed on all points, thanks for confirming! Let's go with the standard term (pending review from our technical writing team). Updated in 749f1db.

I also removed the tag from the reflect and retry plugin since it is not an implementation of durable execution, and fixed an issue in our parser when using more than one word in tags (4a27234).

docs/integrations/restate.md
```diff
@@ -2,6 +2,7 @@
 catalog_title: Restate
 catalog_description: Resilient agent execution and orchestration with durable sessions and human approvals
 catalog_icon: /adk-docs/integrations/assets/restate.svg
+catalog_tags: ["durable execution"]
```
joefernandez: This seems way too specific. What if we call this "Resilience" as a broader category of capabilities, and then it applies to the reflect and retry tool as well.

docs/integrations/temporal.md
```diff
@@ -2,6 +2,7 @@
 catalog_title: Temporal
 catalog_description: Resilient, scalable agents, long-running agents and tools, human approvals, and safe versioning
 catalog_icon: /adk-docs/integrations/assets/temporal.png
+catalog_tags: ["durable execution"]
```
joefernandez: resilience

docs/integrations/reflect-and-retry.md
joefernandez: add : "resilience"

koverholt:
> This seems way too specific. What if we call this "Resilience" as a broader category of capabilities, and then it applies to the reflect and retry tool as well.

Makes sense, thanks! I can see how "Durable Execution" may be too specific for our catalog size and the integrations we have today.

@drewhoskins-temporal @marcusmotill - We're going to go with "Resilience" as the category name. It covers durable execution platforms and also includes the Reflect and Retry plugin and a broader set of future integrations in this space. In the future as our integration catalog continues growing and if this category gets too large, we can revisit using "Durable Execution".

Updated in 27fa6c6.

docs/integrations/restate.md
koverholt: Renamed and fixed in 27fa6c6.

docs/integrations/temporal.md
koverholt: Renamed and fixed in 27fa6c6.

docs/integrations/reflect-and-retry.md
koverholt: Renamed and fixed in 27fa6c6.

PR #1496, opened 2026-03-25
Update mcp tools java

glaforge:


glaforge:
I've also changed the Gemini model versions to use 2.5 instead of the deprecated 2.0 models (including in the Python snippets)

PR #1494, opened 2026-03-25
docs: add complete REST API reference for ADK web server

akshay-kumar-bm:
## Summary

Closes google/adk-python#2511

The existing REST API reference page (docs/api-reference/rest/index.md) only documented two endpoints (/run and /run_sse). This PR expands it to cover all 40 endpoints exposed by the ADK web server.

### What's added

- **System** — /health, /version, /list-apps
- **Agent Execution** — POST /run, POST /run_sse, WebSocket /run_live
- **Sessions** — full CRUD + state patch (6 endpoints)
- **Artifacts** — upload, versioning, metadata (8 endpoints)
- **Memory** — PATCH /memory
- **Events** — event graph visualisation
- **Eval** — eval sets, eval cases, run, results, metrics (9 endpoints)
- **Debug** — trace endpoints (dev only)
- **Builder** — agent editor endpoints (web UI only)
- **Data models** — Content, Event, Session, ArtifactVersion with field-level descriptions and JSON examples

All endpoint descriptions were derived directly from the source in src/google/adk/cli/adk_web_server.py and src/google/adk/cli/fast_api.py. The deprecated POST /sessions/{session_id} endpoint is included and marked as deprecated.

## Test plan

- [x] Verify rendered docs locally with mkdocs serve
- [x] Spot-check a running adk web server at /docs to confirm endpoint signatures match
- [x] Confirm all internal doc links resolve correctly

---
cc @polong-lin @boyangsvl

PR #1500, opened 2026-03-27
Autogenerate Python REST API reference docs from FastAPI OpenAPI spec

koverholt:
Followup to #1494.

Fixes https://github.com/google/adk-python/issues/2511.

This PR replaces hand-written Python REST API docs with Swagger UI generated from the `adk-python` FastAPI app's OpenAPI spec:

- Remove existing hand-written Python REST API doc markdown file
- Add Python REST API docs generation script (similar approach as Python CLI doc generation script)
- Generation script also contains adk-docs Google Analytics snippet/ID
- Adds generated HTML and JSON files for ADK Python 1.28.0
- Update reference in `mkdocs.yml` to point to new generated HTML file

This auto-generates coverage of all 43 endpoints and schemas directly from the Pydantic models and route definitions in adk-python, so the docs stay accurate without manual maintenance. Just run the generation script at release time, same as we do for CLI docs.

Current Python REST API docs page: https://google.github.io/adk-docs/api-reference/rest/

New Python REST API docs page (preview): https://deploy-preview-1500--adk-docs-preview.netlify.app/api-reference/rest/

### DIFF
--- a/docs/api-reference/rest/index.html
+++ b/docs/api-reference/rest/index.html
@@ -17,14 +17,13 @@
 <body>
   <div id="swagger-ui"></div>
   <script src="https://cdn.jsdelivr.net/npm/swagger-ui-dist@5/swagger-ui-bundle.js"></script>
-  <script src="https://cdn.jsdelivr.net/npm/swagger-ui-dist@5/swagger-ui-standalone-preset.js"></script>
   <script>
     SwaggerUIBundle({
       url: "./openapi.json",
       dom_id: "#swagger-ui",
       deepLinking: true,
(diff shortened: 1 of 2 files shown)


docs/api-reference/rest/index.html
```diff
@@ -0,0 +1,31 @@
+<!DOCTYPE html>
```
joefernandez, line 1: it's a little be annoying that the swagger logo dominates the header for this, but I'm willing to let that pass.

koverholt: Noted! I switched to a simpler `BaseLayout` without the top nav bar, we don't need the nav bar or search box since it's only rendering one API spec. Fixed in f2d08de.

PR #1504, opened 2026-03-27
Update CLI reference docs for ADK 1.28.0

koverholt:
Generated using `bash tools/python-cli-docs/generate.sh 1.28.0` in this repo.


koverholt:
The link checker CI job is currently failing (here and in other PRs) due to a transient 429 error, so hopefully that will resolve on its own.

PR #1535, opened 2026-03-31
Update site_url and add CNAME for adk.dev custom domain

koverholt:
- Update `site_url` in mkdocs.yml from `google.github.io/adk-docs` to `https://adk.dev`
- Add `docs/CNAME` file with `adk.dev` to preserve the custom domain across docs deploys (per [mkdocs docs](https://www.mkdocs.org/user-guide/deploying-your-docs/#custom-domains))


koverholt:
Blocked waiting for DNS config to become active and GitHub pages config to be updated.

PR #1520, opened 2026-03-31
fix(docs): correct A2UI pip package name to a2ui-agent-sdk

zeroasterisk:
The A2UI Python SDK is published on PyPI as [`a2ui-agent-sdk`](https://pypi.org/project/a2ui-agent-sdk/), not `a2ui`.

This PR fixes two places in `docs/integrations/a2ui.md`:
1. The install command: `pip install a2ui` → `pip install a2ui-agent-sdk`
2. The PyPI link in the Resources section

cc @nanyu

PR #1541, opened 2026-04-01
fix(docs): correct A2UI pip package name to a2ui-agent-sdk

nan-yu:
The A2UI Python SDK is published on PyPI as `a2ui-agent-sdk`, not `a2ui`. Fixes the install command and the PyPI link in the resources section.


nan-yu:
This is a copy of https://github.com/google/adk-docs/pull/1520 which fixes the CLA check error.

PR #1542, opened 2026-04-01
docs: move Community nav second row

joefernandez:
- move community docs to /community/ directory


koverholt (approved):
LGTM, thanks for making these pages more visible!

Issue #1546, opened 2026-04-02
Add documentation for better Guidance for ExecuteBashTool and BashToolPolicy


shahsohil123:
**Is your feature request related to a problem? Please describe.**

While working with:

from google.adk.tools.bash_tool import ExecuteBashTool, BashToolPolicy

I found that the current documentation lacks practical guidance and real-world examples for safely and effectively using the Bash tool within agents.
A clear and concise description of what the problem is.

There is currently no official documentation for ExecuteBashTool and BashToolPolicy. As a result, it is unclear how these components are intended to be used, how policies should be configured, and what the expected behavior is for command execution and output handling. This creates a barrier to adoption and makes it difficult to safely use the Bash tool in agent workflows.

**Describe the solution you'd like**

Add official documentation for ExecuteBashTool and BashToolPolicy, including:

Basic usage and initialization examples
Explanation of how BashToolPolicy works (e.g., command matching and restrictions)
End-to-end example of using the tool within an agent
Description of command execution behavior and output structure
Best practices for safe and secure usage

**Describe alternatives you've considered**

Reading the source code to infer intended behavior
Experimenting with trial-and-error implementations
Creating custom wrappers to control execution

However, these approaches are inefficient and do not provide clarity on best practices or intended usage.
**Additional context**

This issue was encountered while building an agent that relies on shell commands (e.g., Git operations). Proper documentation would make it significantly easier for developers to adopt these tools and use them safely in real-world scenarios.

koverholt:
Thanks for filing this. The tool is currently in an experimental state that makes it premature to add to ADK docs. The tool exists as of ADK Python 1.26.0, but is marked experimental, not publicly exported, and requires user confirmation for every command. We'll revisit docs once the API is stabilized a bit more.

shahsohil123:
@koverholt - I see, thanks for the detailed clarification and I completely understand the concern around stability and not wanting to prematurely formalize the Tool.

That said, since this functionality is already going to be included in a released version (even if experimental and not publicly exported), there’s a real discoverability gap today. Advanced users will inevitably encounter or look for this capability, and in the absence of any mention, it can lead to confusion or duplicated effort trying to reverse engineer intended usage. 

Would it beneficial to mark these Tools as "Experimental" in an effort to ensure early adopters can understand its usage?

koverholt:
[...]

PR #1547, opened 2026-04-02
docs: add Bash Tool integration documentation #1546

shahsohil123:
* docs: add Bash Tool integration documentation

- Add documentation for the ADK Bash Tool, including `ExecuteBashTool` and `BashToolPolicy`.
- Provide installation instructions and Python code examples for secure command execution.
- Add the Bash Tool integration icon asset.

image


koverholt:
Thanks for the contribution, but this PR can't be merged as-is. The documented API doesn't match the actual implementation: the proposed documentation references nonexistent or hallucinated info such as package name, import paths, and constructor parameters that don't exist. While `bash_tool` exists in ADK, it is currently marked as experimental under a skills feature flag and is not publicly exported from tools. Given the experimental status, we'll wait until the API is stabilized and publicly exported before adding integration docs. Keeping #1546 open to track.

PR #1577, opened 2026-04-08
Fix Go multi-tool tutorial snippet to include two tools

rishipradeep-think41:
## Summary
- update the Go snippet for the multi-tool tutorial to define two tools: `get_weather` and `get_current_time`
- replace the previous single-tool `GoogleSearch` setup so the Go tab matches the tutorial’s multi-tool intent
- align the agent name and behavior with the tutorial flow (`weather_time_agent`, weather/time prompts)

## Why
The Go example under the “Build a multi-tool agent” tutorial currently demonstrates only one tool, which is inconsistent with the tutorial goal and with the Python/TypeScript/Java tabs.

Fixes #1561


koverholt (approved):
Thanks for fixing this example and the PR, and for aligning the Go code snippet with the Python/TS/Java examples per #1561. The intent and the tool logic all LGTM.

Since this was written before the ADK Go v2.0.0 release, I merged main in and resolved the conflict that involved v2 import paths and updating to 2.0.0 API. Builds cleanly against v2.0.0 and passes the Go snippets check.

Unrelated to this PR/issue, but I also bumped the model string to `gemini-flash-latest` for Python, TS, Java code snippets for this example.

Merging. Thanks for the contribution!

PR #1627, opened 2026-04-12
fix: avoid assigning result of delete_session which returns None

izualzhy:
The delete_session method does not return the session service instance, but None instead. Assigning its result back to temp_service leads to unexpected behavior where temp_service becomes None.

This change removes the incorrect assignment and calls delete_session without reassigning, ensuring temp_service remains a valid service instance.

This aligns with the intended usage of delete_session as a side-effect-only method.


koverholt (approved):
Thanks for the PR and the fix! LGTM, good catch.

PR #1688, opened 2026-04-27
chore: update ADK Java doc to version 1.2.0

tilgalas:


koverholt (approved):
Thanks for the updates to the ADK Java API docs. Everything LGTM. Merging.

I will make a separate PR to bump Java version strings on all docs pages.

PR #1697, opened 2026-04-28
Update `A Simple Math Toolset` example code

AlejandroBlanco2001:
The `A Simple Math Toolset` example contains a documentation error that still uses params a names of older versions, this PR is updating the docs to use the new params and ways to do it.

- Removing the name parameter of the `FunctionTool` because this param don't exist no more.
- Using the call to the parent class with the `tool_name_prefix` to achieve the same effect of modifying the call names of the tools.

# Working example

image


koverholt (approved):
Thank you for the PR and debugging/updating this code sample to work with the latest API. ❤️ 

I made some additional update to the code sample to remove the trailing underscore from the prefix and replace `Runner` with `InMemoryRunner`/`run_debug` for a self-contained runnable code sample. LGTM, merging!

PR #1700, opened 2026-04-28
docs(memory): refresh against current adk-python APIs

GWeale:
The memory page has drifted from the implementation:

- The `MultiMemoryAgent` example raises `ValueError` on construction
  (assigns to a Pydantic model with `extra="forbid"`).
- The auto-save callback reaches into the private `_invocation_context`
  attribute when public `Context.add_session_to_memory()` exists.
- The "MemoryService Role" section is missing `add_events_to_memory`
  and `add_memory`.
- The comparison table is missing `VertexAiRagMemoryService` (a third
  bundled implementation).
- "Searching Memory Within a Tool" has Go/TS/Java but no Python tab.
- "How Memory Works in Practice" describes the wrong return type
  (`MemoryResult`; actual is `MemoryEntry`) and shows positional args
  for `search_memory` (it's keyword-only).

This PR refreshes the page against adk-python 1.31.0. The replacement
two-memory example is tool-based (no Agent subclassing) and verified
to construct against current adk-python.

### DIFF
--- a/docs/sessions/memory.md
+++ b/docs/sessions/memory.md
@@ -13,24 +13,30 @@ Think of it this way:
 
 ## The `MemoryService` Role
 
-The `BaseMemoryService` (or `Service` in Go) defines the interface for managing this searchable, long-term knowledge store. Its primary responsibilities are:
+The `BaseMemoryService` (or `Service` in Go) defines the interface for managing this searchable, long-term knowledge store. It supports four operations:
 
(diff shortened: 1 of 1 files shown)


koverholt (approved):
Thank you for the updates! ❤️ 

I updated some product names (Vertex AI -> Agent Platform) in the prose for consistency, and I will open a new PR to make RAG Engine rename updates across the docs, including this page.

LGTM, merging.

PR #1686, opened 2026-04-27
Add documentation pages for metrics and traces

mhenc:
This change adds the documentation for both metrics and traces to provide information regarding OpenTelemetry concepts and configuration details for export telemetry to any destination.

PR #1705, opened 2026-04-29
Add information about 2 additional metrics

mhenc:
Updated `docs/observability/metrics.md` to document the new native emission of Google GenAI metrics by ADK and the behavior when external instrumentation is present.


koverholt (commented):
Thanks for the updates. I now see that #1686 is in a version of ADK Python that is not yet released, which was not clear in the original PR. We will need to un-publish that page and hold off on publishing it until a version of ADK is released with metrics support.

koverholt:
Update: I opened #1713 to readd the metrics page when it's available in a release. Should the content in this PR be merged along with #1713, or is it dependent on other upstream changes?

From what I can tell, these two metrics are emitted by `opentelemetry-instrumentation-google-genai`, not by ADK itself.

mhenc:
Sorry for not letting you know that metrics are not yet released.

No, this change still waits for the PR to be merged. I will let you know when it's ok to review this one.

joefernandez:
Thanks for the update. Looks like these changes landed through a separate PR. Closing

Issue #1710, opened 2026-04-30
Improve BigQuery Agent Analytics plugin doc structure and scannability


caohy1988:
## Problem

The BigQuery Agent Analytics plugin doc is technically rich, but it is difficult to scan and use because setup guidance, deployment instructions, configuration reference, schema details, event payload examples, GCS offloading, query recipes, security, and operational notes are all in one long flow.

After PR #1689, the code-doc parity gaps are mostly closed. The remaining issue is information architecture and readability:

- Readers cannot quickly answer "what events are captured?" without reading several subsections and JSON examples.
- Readers cannot quickly compare configuration options because they are currently a long bullet list.
- The table of contents mixes quickstart, Agent Runtime deployment, reference material, analysis examples, security, and multiprocessing guidance at the same level.
- GCS offloading examples are nested under "Event types and payloads", even though offloading is a storage/configuration behavior rather than an event type.
- The first local-use example is long and includes plugin setup, BigQuery tool setup, Vertex/Gemini environment setup, OpenTelemetry setup, and GCS offloading all at once. This makes the minimal path harder to find.

## Suggested Restructure

Proposed top-level flow:

1. Overview
2. Quickstart: log events to BigQuery
3. Prerequisites and IAM
4. Configuration
5. What the plugin captures
6. BigQuery schema and generated views
7. Storage behavior: truncation, multimodal content, and GCS offloading
8. Query recipes
9. Deploy to Agent Runtime
10. Security and redaction
11. Operations: tracing, flushing, troubleshooting, multiprocessing
12. SDK / dashboards / additional resources

The goal is to make the page answer three common reader questions quickly:

- "How do I turn it on?"
- "What data will it write?"
- "How do I query or configure it?"

### 1. Add a "Captured events" table near the top

Add a concise table before the JSON payload examples. This would make the event surface discoverable without forcing readers through every example.

Suggested columns:

| Event type | Captured when | Key payload fields | View |
| --- | --- | --- | --- |
| `USER_MESSAGE_RECEIVED` | A user message enters the invocation | text summary / content parts | `v_user_message_received` |
| `INVOCATION_STARTING` | An invocation starts | common columns only | `v_invocation_starting` |
| `INVOCATION_COMPLETED` | An invocation completes | common columns only | `v_invocation_completed` |
| `AGENT_STARTING` | Agent execution starts | instruction summary | `v_agent_starting` |
| `AGENT_COMPLETED` | Agent execution completes | latency | `v_agent_completed` |
[...]

PR #1711, opened 2026-04-30
Restructure BigQuery Agent Analytics doc for scannability

caohy1988:
## Summary

Full restructure of the BigQuery Agent Analytics plugin doc, addressing #1710. Two phases in one PR:

### Phase 1 — Additive tables (no section moves)
- **Captured-events summary table**: 19-row table near the top listing all event types, when they fire, key fields, and corresponding view
- **Configuration reference tables**: Converted bullet list to two 4-column tables (constructor params + BigQueryLoggerConfig)
- **Collapsible DDL**: Production DDL wrapped in `??? example` admonition
- **Conversational Analytics**: Demoted from standalone `##` to `###` under query section

### Phase 2 — Structural reorder
- **Shortened quickstart**: Minimal 3-import example; full example with GCS/OTel/BigQuery tools moved to collapsible block
- **Agent Runtime moved later**: From position 4 (interrupting core narrative) to position 9 (after query recipes)
- **GCS offloading extracted**: From nested `####` under "Event types" to its own `## Storage behavior: GCS offloading`
- **Query recipes regrouped by task**: Debug a run, Monitor cost/performance, Inspect tools/interactions, Analyze multimodal, AI root cause, Conversational Analytics
- **Operations consolidated**: Tracing, Public methods, and Multiprocessing grouped under one `## Operations` section
- **New anchors**: `{#prerequisites}`, `{#configuration-options}`

### New section order
1. Overview + Use cases + Captured events
2. Quickstart (shortened)
3. Prerequisites + IAM
4. Configuration (tables)
5. Schema + Views
6. Event types + payloads
7. Storage behavior: GCS offloading
8. Query recipes (task-grouped)
9. Deploy to Agent Runtime
10. Security + redaction
11. Operations (tracing, methods, multiprocessing)
12. SDK / dashboards
13. Feedback + resources

### Anchors
All 5 existing anchors preserved: `#deploy-agent-runtime`, `#event-types`, `#hitl-events`, `#security-credentials`, `#built-in-redaction`. Added: `#prerequisites`, `#configuration-options`.

## Test plan
- [ ] Quickstart renders with minimal example + collapsible full example
- [ ] Captured-events table renders all 19 rows
- [ ] Config tables render correctly (constructor + BigQueryLoggerConfig)
- [ ] DDL is collapsible
- [ ] GCS offloading section renders correctly as standalone `##`
- [ ] Query recipes show task-based `###` grouping with `####` subsections
- [ ] Agent Runtime section appears after query recipes
- [ ] Operations section groups tracing + public methods + multiprocessing
- [ ] All 7 anchors resolve correctly
- [ ] No broken internal cross-references

Ref: #1710

### DIFF
--- a/docs/integrations/bigquery-agent-analytics.md
+++ b/docs/integrations/bigquery-agent-analytics.md
@@ -76,6 +76,7 @@ from google.adk.models.google_llm import Gemini
 from google.adk.plugins.bigquery_agent_analytics_plugin import BigQueryAgentAnalyticsPlugin
 
 os.environ['GOOGLE_CLOUD_PROJECT'] = 'your-gcp-project-id'
+os.environ['GOOGLE_CLOUD_LOCATION'] = 'us-central1'
 os.environ['GOOGLE_GENAI_USE_VERTEXAI'] = 'True'
 
 plugin = BigQueryAgentAnalyticsPlugin(
@@ -431,7 +432,7 @@ The following table lists all 16 auto-created views and their event-specific col
 | **`v_hitl_input_request`** | `tool_name` (STRING), `tool_args` (JSON) |
 | **`v_a2a_interaction`** | `response_content` (JSON), `a2a_task_id` (STRING), `a2a_context_id` (STRING), `a2a_request` (JSON), `a2a_response` (JSON) |
 
-### Event types and payloads {#event-types}
+## Event types and payloads {#event-types}
 
 The `content` column now contains a **JSON** object specific to the `event_type`.
 The `content_parts` column provides a structured view of the content, especially useful for images or offloaded data.
@@ -441,7 +442,7 @@ The `content_parts` column provides a structured view of the content, especially
     - Variable content fields are truncated to `max_content_length` (configured in `BigQueryLoggerConfig`, default 500KB).
     - If `gcs_bucket_name` is configured, large content is offloaded to GCS instead of being truncated, and a reference is stored in `content_parts.object_ref`.
 
-#### LLM interactions (plugin lifecycle)
+### LLM interactions (plugin lifecycle)
 
 These events track the raw requests sent to and responses received from the LLM.
 
@@ -522,7 +523,7 @@ Logged when an LLM call fails with an exception. The error message is captured a
 }
 ```
 
-#### Tool usage (plugin lifecycle)
+### Tool usage (plugin lifecycle)
 
 These events track the execution of tools by the agent. Each tool event includes a `tool_origin` field that classifies the tool's provenance:
 
@@ -595,7 +596,7 @@ Logged when a tool execution fails with an exception. Captures the tool name, ar
 }
 ```
 
-#### State Management
+### State Management
 
 These events track changes to the agent's state, typically triggered by tools.
 
@@ -621,7 +622,7 @@ Tracks changes to the agent's internal state (e.g., custom application state upd
 }
 ```
 
-#### Agent lifecycle & Generic Events
+### Agent lifecycle & Generic Events
 
 <table>
   <thead>
@@ -655,7 +656,7 @@ Tracks changes to the agent's internal state (e.g., custom application state upd
   </tbody>
 </table>
 
-#### Human-in-the-Loop (HITL) Events {#hitl-events}
+### Human-in-the-Loop (HITL) Events {#hitl-events}
 
 The plugin automatically detects calls to ADK's synthetic HITL tools and emits dedicated event types for them. These events are logged **in addition to** the normal `TOOL_STARTING` / `TOOL_COMPLETED` events.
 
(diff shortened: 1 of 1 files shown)


caohy1988:
Found one small issue to tighten before merge:

- The new captured-events table says it "lists every event the plugin logs," but the `HITL_*_COMPLETED` row groups three concrete event types (`HITL_CREDENTIAL_REQUEST_COMPLETED`, `HITL_CONFIRMATION_REQUEST_COMPLETED`, `HITL_INPUT_REQUEST_COMPLETED`). Either expand that wildcard row into the three actual event rows, or change the intro wording to "event types and event families". The PR body/test-plan also says "17 event types," which has the same ambiguity.

Everything else in Phase 1 looks aligned: no section moves, the DDL admonition is low-risk, Conversational Analytics is now under query examples, and the known anchors still resolve (`#deploy-agent-runtime`, `#event-types`, `#hitl-events`, `#security-credentials`, `#built-in-redaction`). `git diff --check` is clean.

caohy1988:
Phase 2 is directionally good; the new order is much easier to follow. I found two things I would fix before merge:

1. **Quickstart may not be runnable with Vertex enabled.** The shortened quickstart sets `GOOGLE_GENAI_USE_VERTEXAI=True` but no longer sets `GOOGLE_CLOUD_LOCATION`. The old full example did set it before Gemini construction, and the Gemini setup docs show it as part of Vertex configuration. Please add something like:

   ```python
   os.environ["GOOGLE_CLOUD_LOCATION"] = "us-central1"
   ```

   or avoid forcing Vertex mode in the minimal quickstart and point readers to the auth/model setup path. Since this is the first runnable snippet, it should not depend on an unstated env var.

2. **`Event types and payloads` is still nested under `Schema and production setup`.** The PR body says the new flow is `Schema + Views` followed by `Event types + payloads`, but the heading is still `### Event types and payloads {#event-types}` after `## Schema and production setup`, so the generated TOC will present event payloads as part of schema. I’d promote it to:

   ```md
   ## Event types and payloads {#event-types}
   ```

   That preserves the existing anchor while matching the intended reader flow. The same applies conceptually to the section-order summary in the PR body.

Optional, but useful: add a tiny install/auth snippet before the quickstart code (`google-adk[bigquery]` requirements + `gcloud auth application-default login`) or explicitly say the commands are in Prerequisites. The quickstart is much shorter now, but it currently starts at code rather than setup.

I tried `uv run mkdocs build --strict`; the local env does not have `mkdocs` available, so I could not verify rendered output here. `git diff --check origin/main...HEAD` is clean.

caohy1988:
Thanks, both review items are addressed:

- Quickstart now sets `GOOGLE_CLOUD_LOCATION` before constructing `Gemini`, so the Vertex path is explicit.
- `Event types and payloads` is now a top-level `##` while preserving `{#event-types}`, and its child headings were promoted consistently.

I did one more source pass over anchors and internal links. The key explicit anchors are still present (`#prerequisites`, `#configuration-options`, `#event-types`, `#hitl-events`, `#deploy-agent-runtime`, `#security-credentials`, `#built-in-redaction`), and `git diff --check origin/main...HEAD` is clean.

No further blockers from me. I still could not run a full MkDocs render locally because this environment does not have `mkdocs` installed in the available `uv` context, so CI/render preview should be the final check for table/admonition layout.

haiyuan-eng-google:
Here is my feedback ::

in  “Captured events summary” table 
- There is a view column but we have not yet introduced the view creation configuration yet.  You can add a line in the description that  says "you can create per event views using option <> " Also add something like this in the  view column description “BigQuery View (optionally created using config <>)”

In BigQueryLoggerConfig options: https://deploy-preview-1711--adk-docs-preview.netlify.app/integrations/bigquery-agent-analytics/#bigqueryloggerconfig-options
- all of the configs are optional - correct?  We have optional keyword in only some like "gcs_bucket_name".  You can just state at the top that thiese are optional configs  and remove the optional keyword from couple of places. 

In Conversational analytics you can also point them to the sample agent.  https://console.cloud.google.com/bigquery/agents_hub;agentsHubTab=Agents;agentsPath=%2Fbq%2Fagents%2Fagent_df1f4aa5-1eda-402b-bf29-bc9c5b376092
they can create a similar agent for their table 

for https://screenshot.googleplex.com/5BoGL2GVRBiTw7m - create a section headers "Additional ways for consuming log data"  with two sub bulltets.  (1) Use Agent Analytic SDK for Agent Eval, Golden trajectory matching, and Trace visualization, and (2) Build a dashboard  using a sample notebook. Here you can also point out they can build the dashboard from the Notebook using Data app. Point to Data app docs. 

 In feedback section:  we make [bqaa-feedback@google.com](mailto:bqaa-feedback@google.com) as an hyperlink?

caohy1988:
Re-reviewed the latest commit `d6eeac0a` against the May 5 feedback.

The requested updates are covered:

- Captured-events table now explains that the **View** column refers to optional per-event BigQuery views created when `create_views` is enabled.
- `BigQueryLoggerConfig` now states that the options have sensible defaults. Keeping `Optional[...]` in the type column is reasonable because it describes whether `None` is an accepted value, not whether the config itself is required.
- Conversational Analytics now links to BigQuery Agents Hub and tells readers to connect an agent to their own `agent_events` table.
- The SDK/dashboard area is now reframed as “Additional ways to consume logged data,” with SDK use cases and dashboard/Data app guidance.
- Feedback email is now a `mailto:` link.

Validation:

```bash
.venv/bin/mkdocs build --strict
```

passes locally, and `git diff --check origin/main...HEAD` is clean. No further blockers from me.

PR #1727, opened 2026-05-05
docs(callbacks): clarify Python callback parameter names

### LINKED ISSUE #1205: Document that callback parameter names must match exactly (e.g., callback_context, not ctx)
## Summary

The callback documentation shows examples with specific parameter names like `callback_context` and `llm_request`, but doesn't explicitly state that these exact names are **required**. Users may assume they can use any parameter name (e.g., `ctx` instead of `callback_context`), leading to confusing runtime errors.

## Problem

When defining a callback with a different parameter name:

```python
async def my_callback(ctx: CallbackContext) -> Optional[types.Content]:
    # ...
```

The callback fails at runtime with:

```
TypeError: my_callback() got an unexpected keyword argument 'callback_context'
```

This happens because ADK invokes callbacks using keyword arguments:

```python
callback(callback_context=callback_context, llm_request=llm_request)
```

## Expected Behavior

The documentation should explicitly state that callback parameter names must match the expected names exactly, such as:

- `callback_context` (not `ctx`, `context`, etc.)
- `llm_request` / `llm_response`
- `tool`, `args`, `tool_context`, `tool_response`

## Suggested Fix

Add a note to the [Callbacks documentation](https://google.github.io/adk-docs/callbacks/) or [Types of Callbacks](https://google.github.io/adk-docs/callbacks/types-of-callbacks/) page, such as:

> **Important:** Callback parameter names must match exactly as shown in the examples. ADK invokes callbacks using keyword arguments, so using different parameter names (e.g., `ctx` instead of `callback_context`) will result in a `TypeError`.

## Affected Pages

- https://google.github.io/adk-docs/callbacks/
- https://google.github.io/adk-docs/callbacks/types-of-callbacks/

EfeDurmaz16:
## Summary
- Adds a note that Python callback parameters must use the documented names because ADK passes them by keyword
- Shows a short correct/incorrect `before_agent_callback` example to prevent `TypeError` failures

Fixes #1205.

## Verification
- git diff --check

### DIFF
--- a/docs/callbacks/types-of-callbacks.md
+++ b/docs/callbacks/types-of-callbacks.md
@@ -13,11 +13,12 @@ These callbacks are available on *any* agent that inherits from `BaseAgent` (inc
 !!! Note
     The specific method names or return types may vary slightly by SDK language (e.g., return `None` in Python, return `Optional.empty()` or `Maybe.empty()` in Java). Refer to the language-specific API documentation for details.
 
-!!! Important "Use the documented callback parameter names"
+??? warning "Python: Use the documented callback parameter names"
+
     In Python, callback function parameter names must match the documented
     names exactly because ADK passes callback arguments by keyword. For example,
     use `callback_context` for agent and model callbacks, and `tool_context` for
-    tool callbacks. Renaming these parameters to aliases such as `ctx` can cause
+    tool callbacks. Renaming these parameters to aliases such as `ctx` will cause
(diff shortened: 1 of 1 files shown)


koverholt (approved):
Thanks for the PR and contribution for an existing issue. I made some changes to the admonition formatting and type, and added a table for quick reference. LGTM, merging.

PR #1731, opened 2026-05-05
docs: clarify function tool defaults

### LINKED ISSUE #1429: Inconsistent advice in documentation regarding the use of default values in function tools
**Describe the bug**

On the parent page for [custom tools](https://github.com/google/adk-docs/blob/87019788e036fdf88bd2f682418c24a46b50c2cc/docs/tools-custom/index.md?plain=1#L360), the guidance strongly advises against the use of default values:
> - **Do not set default values** for parameters. E.g., `def my_func(param1: str = "default")`. Default values are not reliably supported or used by the underlying models during function call generation. All necessary information should be derived by the LLM from the context or explicitly requested if missing.

Whilst the sub page for function tools encourages the use of optional values in a dedicated section [optional parameters](https://google.github.io/adk-docs/tools-custom/function-tools/#optional-parameters) section
> ##### Optional Parameters
> A parameter is considered **optional** if you provide a **default value**. This is the standard Python way to define optional arguments. You can also mark a parameter as optional using `typing.Optional[SomeType]` or the `| None` syntax (Python 3.10+).

**Expected behaviour**

Clear and consistent guidance of when function tool argument defaults should and should not be specified.

EfeDurmaz16:
## Summary
- clarify that defaults should not hide required user or business inputs from tool schemas
- preserve guidance that defaults are valid for genuinely optional tuning parameters
- align the parent custom tools guidance with the function tools optional parameter section

Fixes #1429

## Verification
- git diff --check
- python3 -m mkdocs build --strict

### DIFF
--- a/docs/tools-custom/function-tools.md
+++ b/docs/tools-custom/function-tools.md
@@ -94,6 +94,8 @@ A well-defined function signature is crucial for the LLM to use your tool correc
 === "Python"
     A parameter is considered **optional** if you provide a **default value**. This is the standard Python way to define optional arguments. You can also mark a parameter as optional using `typing.Optional[SomeType]` or the `| None` syntax (Python 3.10+).
 
+    Use defaults only for values that are truly optional. Do not add defaults for information the model should derive from the user request or ask the user to provide.
+
     ???+ "Example: Optional Parameters"
         ```python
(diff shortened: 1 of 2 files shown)


koverholt (approved):
Verified these updates against adk-python source. LGTM, thanks!

PR #1763, opened 2026-05-14
Fix command flag for GCP metrics export

kumar-dhanagopal:


koverholt (approved):
Thanks for catching this! I fixed a couple of more occurrences of the same issue on two related observability pages. LGTM, merging.

PR #1764, opened 2026-05-15
Fix RunConfig docs: correct import path and add missing fields

Raman369AI:
## Summary

The `RunConfig` documentation page was out of sync with the actual implementation in [adk-python](https://github.com/google/adk-python/blob/main/src/google/adk/agents/run_config.py). This PR brings it back in line.

### Import path (critical)

All Python examples used:

```python
from google.genai.adk import RunConfig, StreamingMode  # does not exist
```

The actual import is:

```python
from google.adk.agents.run_config import RunConfig, StreamingMode
```

Fixed across all four occurrences.

### Missing fields

Added 11 previously undocumented fields to the Python class definition, the parameter table, and dedicated subsections:

- `avatar_config`
- `input_audio_transcription`
- `realtime_input_config`
- `enable_affective_dialog`
- `proactivity`
- `session_resumption`
- `context_window_compression`
- `save_live_blob`
- `tool_thread_pool_config` (with code example + GIL note)
- `custom_metadata`
- `get_session_config` (with code example)

### Deprecations

- Marked `save_input_blobs_as_artifacts` as deprecated and pointed readers to `SaveFilesAsArtifactsPlugin`. Removed it from the comprehensive Python example.

### Corrections

- `StreamingMode.BIDI`: the docs claimed it enables bidirectional streaming. The source comment is explicit that BIDI is not used by `run_async()` and that real bidirectional streaming goes through `runner.run_live()`. Updated the description accordingly.
- `output_audio_transcription`: default in the table corrected from `None` to `AudioTranscriptionConfig()` (the field uses `default_factory=types.AudioTranscriptionConfig`).

## Test plan

- [x] `mkdocs serve` builds with no errors or warnings
- [x] `/runtime/runconfig/` page renders and includes the new sections
- [x] No remaining references to `google.genai.adk` on the page

### DIFF
--- a/docs/runtime/runconfig.md
+++ b/docs/runtime/runconfig.md
@@ -1,459 +1,12 @@
 # Runtime Configuration
 
 <div class="language-support-tag">
-    <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v0.1.0</span><span class="lst-typescript">Typescript v0.2.0</span><span class="lst-go">Go v0.1.0</span><span class="lst-java">Java v0.1.0</span>
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v0.1.0</span><span class="lst-typescript">TypeScript v0.2.0</span><span class="lst-go">Go v0.1.0</span><span class="lst-java">Java v0.1.0</span>
 </div>
 
-`RunConfig` defines runtime behavior and options for agents in ADK. It controls
-speech and streaming settings, function calling, artifact saving, and limits on
-LLM calls.
-
-When constructing an agent run, you can pass a `RunConfig` to customize how the
-agent interacts with models, handles audio, and streams responses. By default,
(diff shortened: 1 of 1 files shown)


koverholt (approved):
Thanks for the PR and updates. This PR was a good reminder that the RunConfig page was due for a rewrite/refactor, since in its current state it looks more like a duplicate of the API reference page which is light on concepts/use cases and is high maintenance in terms of updating individual params and manual doc strings.

Instead, I pushed changes to rework the RunConfig docs page from a per-field API reference duplication into a use-case-oriented guide organized around five common tasks: session management, streaming, audio/speech, live agents, and runtime limits. And links to the API reference docs are provided so we don't have to maintain all of that here across all language SDKs.

Also fixed several code samples from the pre-existing page including incorrect Python imports, TypeScript type errors, deprecated Java builder methods.

Raman369AI:
Thank you @koverholt

PR #1776, opened 2026-05-19
Update environment-toolset.md

zyantw:
My contribution for https://github.com/google/adk-docs/pull/1745#event-25685414488


zyantw:
trying to fix the CLA with another pull, i already double check all my configurations

PR #1779, opened 2026-05-19
Updated the python code sample for routing.

happyhuman:
stage: https://deploy-preview-1779--adk-docs-preview.netlify.app/graphs/routes/#route-branches-and-conditional-execution


docs/graphs/routes.md
```diff
@@ -120,29 +120,83 @@ edges=[
...
 
 The subsequent rows of the ***edges*** arrays after the START keyword define
-additional execution logic for nodes. For branching paths, you define a node,
-usually a ***FunctionNode***, that outputs a ***route*** with one or more route
-values. In the edges graph, you define the execution logic with route values and
-target nodes, as shown in the following code example:
+additional execution logic for nodes. For branching paths, which is how you create a conditional node, you define a node,
+usually a ***FunctionNode***, that outputs an Event with a specific  ***route*** value. In the edges graph, you then define the conditional execution logic by mapping these route values to target nodes, as shown in the following code example of a weekend planner agent:
 
 ```python
```
joefernandez, line 129: This is a nice code example. However, inserting it inline here breaks the teaching pattern of this page. The idea is that we present the features and show the reader the smallest, relevant snippet to demonstrate that code. Putting the full example inline makes it harder to quickly grasp the relevant code pattern. Here's my request:

Can you revert to the old, simple example, put this more built-out example in the examples code directory, and then link to it from the end of this section? That way readers can get the quick idea of the code, AND also click through to get a fully-realized implementation if they want it.

joefernandez: thanks for update. This seems like a reasonable compromise

joefernandez (approved):
Thanks for updates!

PR #1786, opened 2026-05-20
Add Python API reference docs and generation script for ADK Python 2.0.0

koverholt:
Adds generation tooling and built API reference docs for ADK Python 2.0.0.

**API reference docs built output (`docs/api-reference/python/`):**

- 88 documented modules (up from ~55 in the ADK 1.x.x Python build)
- New: integrations, environment, labs, optimization, skills, workflow, and 18 new tool modules

**Tooling (`tools/python-api-docs/`):**

The `adk-python` repo does not currently include Sphinx config, so this tooling generates API docs independently (similar pattern as the CLI, REST, and Kotlin doc scripts).

- `generate.sh` - Clones adk-python at a version tag, installs it, and builds Sphinx HTML
- `discover_modules.py` - Discovers public modules from the installed package (generates google-adk.rst at build time)
- `source/conf.py` - Sphinx config (Furo theme, autodoc-pydantic, Google Analytics)
- `source/index.rst` - Sphinx toctree root

Usage: `bash tools/python-api-docs/generate.sh 2.0.0`

Staging preview: https://deploy-preview-1786--adk-docs-preview.netlify.app/api-reference/python/


joefernandez (approved):
Nice work. thanks!

PR #1793, opened 2026-05-22
Update e2a integration page with remote MCP server and 7 tools

jiashuoz:
## Summary

The e2a integration page has drifted from what the e2a MCP server actually ships. Two concrete gaps:

1. **Hosted MCP endpoint isn't documented.** e2a runs a hosted Streamable HTTP MCP at `https://mcp.e2a.dev/mcp` that accepts an API key in the `Authorization` header. The current page only shows the stdio path (`npx -y @e2a/mcp-server`), which doesn't work for ADK agents deployed to Cloud Run since Cloud Run [doesn't host stdio MCP servers](https://docs.cloud.google.com/run/docs/host-mcp-servers). This PR adds a **"Hosted MCP Server"** tab to both the Python and TypeScript code examples using `StreamableHTTPConnectionParams`.

2. **7 tools are missing from the table.** The e2a MCP server now registers 18 tools (verified against [`mcp/tests/http.test.ts`'s `lists every registered tool after initialize` test](https://github.com/Mnexa-AI/e2a/blob/main/mcp/tests/http.test.ts#L193-L217)). The current page lists 11. This PR adds:

   | Section | Added |
   |---|---|
   | Identity | `update_agent`, `delete_agent` |
   | Messages | `get_attachment_data` |
   | Domains *(new section)* | `list_domains`, `register_domain`, `verify_domain`, `delete_domain` |

## Minor description fixes

- `whoami`: notes `E2A_AGENT_EMAIL` is only required when the account has more than one agent (the hosted endpoint auto-resolves the sole agent at session init).
- `delete_agent` / `delete_domain`: flag the `confirm: true` required argument.
- `get_attachment_data`: corrected from "attachment id" to "0-based attachment index" (the actual schema field is `attachment_index: z.number()`).
- `E2A_BASE_URL`: clarified that the variable has no effect when connecting to the public hosted endpoint.

## What's not changing

- Frontmatter: `catalog_title`, `catalog_description`, `catalog_icon`, `catalog_tags: ["mcp"]` — all preserved verbatim.
- Use-cases bullets and the cloud-mode webhook warning — preserved.
- Existing stdio code blocks — preserved as the first tab under each language.

## Verification

- Tool list count (18) matches the MCP server's actual `registerTool` calls across [`mcp/src/tools/agents.ts`](https://github.com/Mnexa-AI/e2a/blob/main/mcp/src/tools/agents.ts), [`messages.ts`](https://github.com/Mnexa-AI/e2a/blob/main/mcp/src/tools/messages.ts), [`domains.ts`](https://github.com/Mnexa-AI/e2a/blob/main/mcp/src/tools/domains.ts), and [`hitl.ts`](https://github.com/Mnexa-AI/e2a/blob/main/mcp/src/tools/hitl.ts).
- TypeScript hosted code block uses `transportOptions.requestInit.headers` (matches the `@google/adk` `StreamableHTTPConnectionParams` type definition; the top-level `header` field is deprecated).
- Python hosted code block verified against `google.adk.tools.mcp_tool.mcp_session_manager.StreamableHTTPConnectionParams`.
- Hosted endpoint reachable: `curl https://mcp.e2a.dev/healthz` → 200, `/.well-known/oauth-protected-resource` returns valid metadata.

### DIFF
--- a/docs/integrations/e2a.md
+++ b/docs/integrations/e2a.md
@@ -53,7 +53,7 @@ your agent. Outbound mail can be held for human approval before it ships.
 
         ```python
         from google.adk.agents import Agent
-        from google.adk.tools.mcp_tool.mcp_toolset import McpToolset
+        from google.adk.tools.mcp_tool import McpToolset
         from google.adk.tools.mcp_tool.mcp_session_manager import StdioConnectionParams
         from mcp import StdioServerParameters
 
@@ -94,7 +94,7 @@ your agent. Outbound mail can be held for human approval before it ships.
         import os
 
         from google.adk.agents import Agent
-        from google.adk.tools.mcp_tool.mcp_toolset import McpToolset
+        from google.adk.tools.mcp_tool import McpToolset
         from google.adk.tools.mcp_tool.mcp_session_manager import (
             StreamableHTTPConnectionParams,
         )
--- a/docs/integrations/e2a.md
+++ b/docs/integrations/e2a.md
@@ -53,7 +53,7 @@ your agent. Outbound mail can be held for human approval before it ships.
 
         ```python
         from google.adk.agents import Agent
-        from google.adk.tools.mcp_tool import McpToolset
+        from google.adk.tools.mcp_tool.mcp_toolset import McpToolset
         from google.adk.tools.mcp_tool.mcp_session_manager import StdioConnectionParams
         from mcp import StdioServerParameters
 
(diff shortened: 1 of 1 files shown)


koverholt (approved):
Thanks for the updates to the integration page. I made some edits to fix the import paths to match repo conventions, renamed tabs from "Hosted" to "Remote", replaced placeholder patterns to be consistent with other integration pages, and trimmed verbose prerequisites/config/admonition text. Content LGTM, merging.

PR #94, opened 2025-04-16
Update tutorial.md

AshleyHenry15:
Fixes formatting issue for ul list

Closes #92

PR #1815, opened 2026-06-03
docs(integrations/bqaa): document 2.0.0 reliability fixes

caohy1988:
Reflects the three reliability/observability fixes that landed in [`google/adk-python` commit `a5fa3da0`](https://github.com/google/adk-python/commit/a5fa3da0) ("feat: BigQuery Agent Analytics reliability fixes"). Filed as a draft so the doc can land alongside the 2.0.0 release notes — please retarget to the release branch / convert to a regular PR when ready.

## Why this update is needed

Commit `a5fa3da0` bundles three changes that the current page at <https://adk.dev/integrations/bigquery-agent-analytics/> contradicts or omits:

| Code change | What the page currently says | What the page now needs to say |
|---|---|---|
| Plugin no longer calls `tracer.start_span(...)`; it tracks `span_id` / `trace_id` on its own contextvar stack | "The plugin **automatically generates spans** … will use **valid OTel spans**, populating trace_id/span_id with standard OTel identifiers" | Plugin generates `span_id` internally; **does not export OTel spans**; `trace_id` is **inherited from the ambient OTel span** (Agent Engine root, Runner span, …) so BigQuery rows still join to Cloud Trace |
| `BigQueryAgentAnalyticsPlugin.get_drop_stats()` added; per-reason drop tracking | not documented at all | new "Dropped-event observability" subsection with the `drop_reason` enum + polling example |
| Cross-region Storage Write API routing fix | the `location` parameter exists but cross-region routing wasn't reliable | mention in the 2.0.0 release notes block at the top of the page |

## Diff structure

| Section touched | Change |
|---|---|
| Intro version blurb | New "Version 2.0.0 bundles three reliability and observability fixes" subsection listing all three by name |
| Quickstart full-example code comment about the OTel TracerProvider | Rewritten so the user understands a `TracerProvider` is **optional** and that the plugin doesn't emit spans through it |
| `## Operations` → `### Tracing and observability` | Bullets rewritten end-to-end to match the new (no-export) behavior. New "Migrating from 1.x" admonition for anyone who was relying on the plugin to feed their own OTel exporter |
| `### Public methods` | Added `plugin.get_drop_stats()` to the list |
| `### Dropped-event observability` (new) | Drop-reason table + read-the-counts example + polling/export loop |

No structural changes to the schema, event-types, query-recipes, Agent Engine deploy walkthrough, GCS-offloading, or security sections.

```
docs/integrations/bigquery-agent-analytics.md | 143 ++++++++++++++++++++++----
1 file changed, 122 insertions(+), 21 deletions(-)
```

## Verification

- **End-to-end validation against real Agent Engine + real BigQuery** for the "no duplicate spans" claim is captured on the customer issue at <https://github.com/haiyuan-eng-google/BigQuery-Agent-Analytics-SDK/issues/94#issuecomment-4597369748> — Cloud Trace shows exactly the framework's 7 spans (no plugin-instrumented ones), and BigQuery rows still share the same `trace_id` as Cloud Trace.
- `markdown.markdown(...)` parses the updated page cleanly (no syntax regressions).

## Why draft

- Want to give the doc team a chance to verify the version number (`2.0.0`) and the wording around "Migrating from 1.x" before merging.
- Open to splitting into three smaller PRs if you'd rather land each fix's docs independently — the file is structured so each set of edits is contiguous and easy to extract.

## Refs

- `google/adk-python` commit `a5fa3da0`
- haiyuan-eng-google/BigQuery-Agent-Analytics-SDK#94 (customer bug the Cloud Trace fix closes)


caohy1988:
Addressed all three findings in [4cd6176](https://github.com/google/adk-docs/pull/1815/commits/4cd6176):

**1. Important — schema reference now matches the 2.0.0 tracing section.**
- `trace_id` column row: dropped "OpenTelemetry Trace ID" wording. New description: "32-character hex Trace ID. Inherited from the ambient OpenTelemetry span when one is active (e.g. Agent Engine's invocation span or the ADK Runner span) so BigQuery rows join cleanly to your existing Cloud Trace traces; otherwise generated by the plugin per invocation." Example value updated to a real 32-hex value from the Agent Engine validation on #94 (`a2c7f13d3a3f0bbb8793692f76a6012a`).
- `span_id` column row: rewrote to call out the plugin-generated 16-hex shape and that `tracer.start_span` is not called against the configured provider. Example value updated to a real 16-hex value (`3916f5762bcd4d42`); the prior `69867a836cd94798be2759d8e0d70215` was 32 chars, contradicting the "16-char hex" label.
- `parent_span_id` column row: same fix on length + example (`4c4a42bfdeb84934`).
- Manual DDL `OPTIONS(description=...)` strings for all three columns rewritten so anyone copy-pasting the production DDL sees the correct semantics.

**2. Minor — executable `TracerProvider()` removed from the quickstart full example.** Replaced with a non-code block of inline `#` comments that explain: BQAA generates `span_id` internally; `trace_id` inherits from whichever ambient span is open (Agent Engine wires this automatically, local framework-instrumented runners do the same, otherwise the plugin falls back per invocation); and explicitly state that **setting a bare `TracerProvider` with no ambient span will NOT populate `trace_id` with a "real" OTel id — only an active span does.** Cross-link to the Tracing section for details. This removes the inference path the reviewer was worried about.

**3. Minor — `queue_max_size` named directly in the drop-reason table.** Reworded the `queue_full` row from "Increase `batch_size` / queue capacity or scale the consumer" to: "Increase `queue_max_size` on `BigQueryLoggerConfig`, raise `batch_size` to drain in larger chunks, or scale the consumer side (more concurrent invocations finishing faster)." Verified `queue_max_size` is the actual `BigQueryLoggerConfig` option (already documented at L332 / L378 of the same page).

Round-2 diff: +21 / −18 over the prior PR head. Markdown parses cleanly. Branch is still one commit on top of `main`.

caohy1988:
Round 3 update: removed all `2.0.0` version anchors from the doc per maintainer request — wording is now version-agnostic so the page can land independently of any specific ADK release tag.

Specifically:
- "Version 2.0.0 bundles three reliability and observability fixes:" → "The plugin includes three reliability and observability fixes:"
- Tracing & observability section: dropped "Starting in **2.0.0** the plugin generates…" → just "The plugin generates…"
- Admonition retitled: "Migrating from 1.x" → "If you relied on the plugin to feed your OTel exporter". Body dropped "in 2.0.0" qualifier.
- `plugin.get_drop_stats()` line: dropped the `*(2.0.0+)*` tag.
- Dropped-event observability section intro: dropped "*Added in ADK Python 2.0.0.*" lead and "Before 2.0.0…" / "Starting in 2.0.0…" history clause; replaced with present-tense description of the API.

Round-3 diff vs prior PR head: +17 / -18. Markdown still parses. Net change vs main is smaller now (~104 added / ~30 deleted).

PR #1822, opened 2026-06-09
docs(plugins): correct before_run_callback flow control behavior

yana-gi:
## Summary

The **Runner start callbacks** section of the Plugins guide describes `before_run_callback` as returning a `types.Content` to **replace** the user's original message. That is actually the behavior of `on_user_message_callback` — the text (including the *Purpose* line) appears to have been duplicated from the *User Message callbacks* section.

In the implementation, returning a `Content` from `before_run_callback` **halts the run** (early exit): the `Runner` yields it as an early-exit event and skips all subsequent processing instead of invoking the agent.

## Evidence

`Runner._exec_with_plugin` in [`src/google/adk/runners.py`](https://github.com/google/adk-python/blob/main/src/google/adk/runners.py):

```python
early_exit_result = await plugin_manager.run_before_run_callback(...)
if isinstance(early_exit_result, types.Content):
    early_exit_event = Event(..., content=early_exit_result)
    ...
    yield early_exit_event      # early exit, agent never runs
else:
    # normal execution
```

This also matches the `BasePlugin.before_run_callback` docstring:
> Returning a value to halt execution of the runner and ends the runner with that event.

The early-exit behavior was established in adk-python `bf72426a` ("fix: runner was expecting Event object instead of Content object when using early exit feature", 2025-07-29), which predates this guide (added 2025-08-01). So the doc has been inaccurate since it was first written.

## Changes

Fix the *When It Runs*, *Purpose*, and *Flow Control* bullets for `before_run_callback`:

- **When It Runs** — clarify it runs *after* `on_user_message_callback` (the order shown in the callback diagram), not "before any other processing".
- **Purpose** — global setup/initialization before the invocation (was duplicated from `on_user_message_callback`).
- **Flow Control** — returning `Content` **halts execution** / early-exits, rather than replacing the user message.

The `on_user_message_callback` section is unchanged — its description is correct.


koverholt (approved):
Thanks for the PR and the accuracy fix! Verified against the adk-python SDK. LGTM, merging.

PR #1672, opened 2026-04-22
docs(integrations): add Cisco AI Defense integration page

shiva-guntoju-09:
## Summary

Add an integration page for the [Cisco AI Defense plugin for Google ADK](https://github.com/cisco-ai-defense/ai-defense-google-adk), which provides runtime LLM and MCP tool inspection with monitor and enforce modes.

**Files added:**

- `docs/integrations/cisco-ai-defense.md` — integration documentation page
- `docs/integrations/assets/cisco-ai-defense.png` — logo asset (400×400 square PNG)

**Page sections:**

- Use cases, prerequisites, installation
- Quick start (one-liner `defend()` API)
- Plugin approach (`CiscoAIDefensePlugin`) and callback approach (`make_aidefense_callbacks`)
- Modes table (monitor / enforce / off)
- Callbacks covered table
- Violation callback
- Inspector-based variant (`AgentsecPlugin`) with retry and fail-open
- Resource links (GitHub, PyPI, Cisco AI Defense)

**Package:** Published on [PyPI](https://pypi.org/project/cisco-aidefense-google-adk/) as `cisco-aidefense-google-adk` v1.0.0.

Resolves #1665

## Test plan

- [x] Ran `mkdocs serve` locally — page builds without errors
- [x] Verified the Cisco AI Defense card appears in the integration catalog
- [x] Verified the full integration page renders correctly with all code examples
- [x] Frontmatter follows the standard template (`catalog_title`, `catalog_description`, `catalog_icon`, `catalog_tags`)

PR #1850, opened 2026-06-18
docs(integrations): add Agent Threat Rules (ATR) guardrail plugin

eeee2345:
This adds an integrations catalog page for `adk-atr-guardrail`, a standalone, in-process security guardrail plugin for ADK backed by the open-source Agent Threat Rules (ATR) ruleset.

This is the follow-up requested in google/adk-python#6130: that PR added the guardrail as an ADK sample, and the maintainer recommended shipping it instead as a standalone package listed in the integrations catalog. The package is now published, so the page documents installing and using it.

What it provides
- A `BasePlugin` that enforces ATR detection at three fail-closed Runner callbacks: before_run (halts the run on a malicious user message), before_model (skips the model call when the assembled prompt still carries a threat), and before_tool (returns an error instead of executing a tool whose arguments match a rule).
- Detection is deterministic, in-process via the pyatr engine -- no model call, no network, no API key.

Details
- Page: docs/integrations/atr-guardrail.md (+ icon at docs/integrations/assets/atr-guardrail.png), tagged `resilience`, mirroring the structure of the existing cisco-ai-defense page.
- Package on PyPI: https://pypi.org/project/adk-atr-guardrail/ (pip install adk-atr-guardrail)
- Package source: https://github.com/eeee2345/adk-atr-guardrail
- The code example is complete and runnable; the rule count is read at runtime rather than hard-coded.
- Verified locally with mkdocs build (page renders and the catalog card appears under Resilience).

Disclosure: I maintain the open-source ATR ruleset referenced here.


eeee2345:
Context for reviewers: this page was created at the request of @rohityan, who closed the original adk-python sample PR (google/adk-python#6130) with the guidance to "release the feature as a standalone package that we will then share through" the integrations catalog. The package is now published on PyPI (`pip install adk-atr-guardrail`), so this page documents installing and using it.

It follows the same structure as the existing Cisco AI Defense integration page (#1672) — a guardrail plugin tagged `resilience` — and was verified locally with `mkdocs build`. Happy to adjust anything to fit catalog conventions.

koverholt (approved):
Thanks for the PR and for contributing this integration! I made some small edits to switch to the App plugin pattern (avoids the deprecated plugins= warning), bumped the prereq to ADK >= 2.0.0, added model="gemini-flash-latest", and dropped the resilience tag. LGTM, merging.

PR #1851, opened 2026-06-21
Update link for Dynamic workflows section

Astroa7m:


joefernandez (approved):
Thanks for the update!

PR #1868, opened 2026-06-25
Runtime docs cleanup for Go

ToniCorinne:
- Differentiating between CLI flag options for Python vs. Go
- Clarifies adk web usage with ADK Go


docs/runtime/api-server.md
```diff
@@ -126,6 +160,27 @@ The output should appear similar to:
 
 Your server is now running locally. Ensure you use the correct **_port number_** in all the subsequent commands.
 
+!!! note "Go: default port and path prefix"
```
koverholt: Suggest moving this note so that it's only visible under the `Go` tab above, otherwise it adds a lot of visual noise to the page when it's only really pertinent to Go users in this section.

docs/runtime/web-interface/index.md
```diff
@@ -87,29 +115,87 @@ Use the following command to start the ADK web interface:
...
 browser to use the web interface:
 
-```shell
-+-----------------------------------------------------------------------------+
-| ADK Web Server started                                                      |
-|                                                                             |
-| For local testing, access at http://localhost:8000.                         |
-+-----------------------------------------------------------------------------+
-```
+=== "Python / TypeScript / Java"
```
koverholt: If we decide to use different consolidated tab labels like `Python / TypeScript / Java`, then it breaks the tab sync when selecting a language on the page since they are named differently.

Suggest either using separate language tabs so that they remain in sync when selected (at the expense of duplicating some content), or if the content duplication is too high, we could use embedded code snippets instead.

docs/runtime/web-interface/index.md
```diff
@@ -87,29 +115,87 @@ Use the following command to start the ADK web interface:
...
+    2025/01/01 00:00:00 Web servers starts on http://localhost:8080
+    2025/01/01 00:00:00        webui:  you can access API using http://localhost:8080/ui/
+    2025/01/01 00:00:00        api:  you can access API using http://localhost:8080/api
+    ```
 
 ## Common options
 
-Here are some commonly used options for the `adk web` command. Run `adk web
---help` to see all available options.
+=== "Python / TypeScript"
```
koverholt: See previous comment on separate language tabs.

docs/runtime/command-line.md
```diff
@@ -40,17 +64,40 @@ Use the following command to run your agent in the ADK command line interface:
...
-responses directly in your terminal:
+responses directly in your terminal.
 
-```shell
-Running agent my_agent, type exit to exit.
-[user]: What's the weather in New York?
-[my_agent]: The weather in New York is sunny with a temperature of 25°C.
-[user]: exit
-```
+=== "Python / TypeScript / Java"
```
koverholt: See previous comment on separate language tabs.

koverholt (approved):
Thanks for the additions. I verified that all of the Go code and references are correct.

I left a couple of comments about one Go note and a few combined tab labels. Beyond those we can perform some additional cleanup to make the reference to multiple language SDKs cleaner, but that can be done in a future PR and is not a blocker for this one.

docs/runtime/api-server.md
ToniCorinne: done

docs/runtime/web-interface/index.md
ToniCorinne: Sounds good - I wasn't sure the best approach on that due to the content duplication, but I went ahead and did that for these.

docs/runtime/command-line.md
ToniCorinne: fixed

docs/runtime/web-interface/index.md
ToniCorinne: done

PR #1876, opened 2026-06-25
Updating Java documentation for BigQueryAgentAnalyticsPlugin

vm-r:
Fixing  java version, adding java example for formatter, small format…

### DIFF
--- a/docs/integrations/bigquery-agent-analytics.md
+++ b/docs/integrations/bigquery-agent-analytics.md
@@ -8,7 +8,7 @@ catalog_tags: ["observability", "google"]
 # BigQuery Agent Analytics plugin for ADK
 
 <div class="language-support-tag">
-  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v1.21.0</span><span class="lst-java">Java v1.4.0</span>
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v1.21.0</span><span class="lst-java">Java v1.5.0</span>
 </div>
 
 !!! important "Version Requirement"
@@ -1906,6 +1906,7 @@ Provide a custom `content_formatter` function in `BigQueryLoggerConfig` to strip
 or mask sensitive fields before they are written:
 
 
+=== "Python"
 
     ```python
     import json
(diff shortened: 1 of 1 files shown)


haiyuan-eng-google (approved):

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -8,7 +8,7 @@ catalog_tags: ["observability", "google"]
 # BigQuery Agent Analytics plugin for ADK
 
 <div class="language-support-tag">
-  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v1.21.0</span><span class="lst-java">Java v1.4.0</span>
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v1.21.0</span><span class="lst-java">Java v1.5.0</span>
```
koverholt, line 11: The docs version tag convention is when this functionality was "introduced in". In `adk-java` the plugin shipped in 1.0.0 with some updates in subsequent versions. If this page is meant to document the new event types added in 1.5.0, the bump makes sense; otherwise it diverges from the "introduced in" convention. Can you confirm the intent?

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -1942,26 +1943,48 @@ or mask sensitive fields before they are written:
...
-
-    BiFunction<Object, String, Object> redactCredentials = (content, eventType) -> {
-      String text = content.toString();
-      for (String key : SENSITIVE_KEYS) {
-        // Redact values in JSON-like strings: "client_secret": "GOCSPX-xxx"
-        text = text.replaceAll(
-            "(?i)(\"" + key + "\"\\s*:\\s*)\"[^\"]*\"",
-            "$1\"[REDACTED]\""
-        );
+    ...
```
koverholt: The `...` hides imports the snippet actually needs. Suggest listing them explicitly so the example compiles when copied: `com.google.adk.models.LlmRequest`, `com.google.adk.models.LlmResponse`, `com.google.genai.types.Content`, `com.google.genai.types.Part`, `java.util.List`, `java.util.ArrayList`.

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -1942,26 +1943,48 @@ or mask sensitive fields before they are written:
...
+        }
+        return res;
+      } else if (content instanceof Content content2) {
+        return maskContent(content2);
+      } else if (content instanceof String string && string.contains("secret")) {
+        return string.replace("secret", "****");
       }
-      return text;
+      return content;
     };
```
koverholt: Tool and HITL events deliver content as `ImmutableMap<String,Object>`, not as `Content/String`. This formatter returns Maps unchanged, so secrets in tool args/results won't be masked. Suggest adding a `content instanceof Map` branch that masks map values.

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -1942,26 +1943,48 @@ or mask sensitive fields before they are written:
...
+        }
+        return req.toBuilder().contents(maskedContents).build();
+      } else if (content instanceof LlmResponse res) {
+        if (res.content().isPresent()) {
+          return res.toBuilder().content(maskContent(res.content().get())).build();
+        }
+        return res;
+      } else if (content instanceof Content content2) {
+        return maskContent(content2);
+      } else if (content instanceof String string && string.contains("secret")) {
```
koverholt: No content is ever passed to the formatter as a bare `String`, so this branch never executes. Recommend removing it.

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -1942,26 +1943,48 @@ or mask sensitive fields before they are written:
...
+      return content;
     };
 
     BigQueryLoggerConfig config = BigQueryLoggerConfig.builder()
-        .contentFormatter(redactCredentials)
-        // ... other options
-        .build();
+    .contentFormatter(maskObject)
+    // ... other options
+    .build();
```
koverholt: `.contentFormatter(...)` and `.build()` are dedented to the same level as `BigQueryLoggerConfig`. Indent under the builder to match the style used elsewhere on this page.

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -1942,26 +1943,48 @@ or mask sensitive fields before they are written:
...
+
+    private static Content maskContent(Content originalContent) {
```
koverholt: The `private static Content maskContent(...)` method appears after the `config` builder statements inside the same block, which isn't valid at that position. Suggest defining the helper before its use, or add a comment noting it's a separate helper method.

koverholt (changes requested):
Thanks for improving these examples! Requesting changes on two functional points, details are in inline comments.

docs/integrations/bigquery-agent-analytics.md
vm-r: The intent here is to notify user can start using plugin  starting 1.5.0. I would say previous versions contained plugin with partial functianality.

koverholt: Makes sense, thanks!

docs/integrations/bigquery-agent-analytics.md
vm-r: added almost full example

docs/integrations/bigquery-agent-analytics.md
vm-r: Added this branch.

koverholt (approved):
Thanks for working through the issues. Updates LGTM, merging.

PR #1877, opened 2026-06-26
feat(integrations): add Mimir Memory — persistent cross-session memory for ADK

tcconnally:
Adds integration documentation for [**Mimir Memory**](https://github.com/Perseus-Computing-LLC/adk-mimir-memory) — a persistent, local, encrypted cross-session memory backend for ADK agents.

### What this integration provides

- **`MimirMemoryService`**: drop-in replacement for `InMemoryMemoryService` with persistent SQLite storage
- **AES-256-GCM encryption**: memory data encrypted at rest
- **Hybrid search**: FTS5 keyword + dense vector retrieval
- **Zero cloud dependencies**: single Rust binary, fully local
- **Perseus live context**: optional workspace-aware agent sample

### Backend comparison

| Backend | Encryption | Hybrid Search | Local |
|---|---|---|---|
| InMemoryMemoryService | ❌ | ❌ | ✅ |
| VertexAiMemoryBankService | ❌ | Gemini-driven | ❌ |
| VertexAiRagMemoryService | ❌ | GCP vector | ❌ |
| **MimirMemoryService** | ✅ AES-256 | ✅ BM25+FTS5+Dense | ✅ |

### Installation

```bash
pip install adk-mimir-memory
```

### References

- Package: https://github.com/Perseus-Computing-LLC/adk-mimir-memory
- Backing service (Mimir): https://github.com/Perseus-Computing-LLC/mimir
- Original ADK PR (closed — migrated to standalone): https://github.com/google/adk-python/pull/6169

PR #1880, opened 2026-06-26
docs(mimir): sharpen the Mimir Memory positioning sentence

tcconnally:
A small follow-up to #1877 (Mimir Memory integration). Adds one positioning sentence to the intro that leads with what's distinctive about Mimir among ADK memory backends — the **combination** of fully on-device hybrid search (embeddings included, so no external API keys or cloud calls) **and** AES-256-GCM encryption at rest — rather than listing those features in isolation. This is the angle that matters for private / air-gapped / compliance-sensitive deployments.

No structural, table, or code changes — intro paragraph only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


koverholt:
Thanks for the update. You'll need to remove Claude Code as a commit author since that is causing the CLA check to fail. You can do that by rebasing and force pushing to this PR/branch, or by opening a new PR.

tcconnally:
Done — removed the Claude Code co-author trailer and force-pushed. The branch is now a single-file change (`docs/integrations/mimir.md`) authored solely under my CLA-signed identity, so the CLA check should re-evaluate cleanly. Thanks for the pointer!

PR #1875, opened 2026-06-25
Document ID token authentication for service accounts (#1350 - 8)

zyantw:
Documents ID token support for service-account authentication, needed when an agent calls IAM-restricted services such as private Cloud Run or Cloud Functions. Adds a "Use ID token" section under "Configure tools with authentication" in the Tool Authentication guide, with an end-to-end OpenAPIToolset example, the ServiceAccount config parameters, and troubleshooting tips.

Rendered preview: https://deploy-preview-1875--adk-docs-preview.netlify.app/tools-custom/authentication/#use-id-token

Original Issue: https://github.com/google/adk-docs/issues/1350 (see item 8)

Original adk-bot PR: https://github.com/google/adk-docs/pull/1365. This PR supersedes #1365 (adk-bot's auto-generated draft, which used a helper signature that doesn't exist); it uses the actual `ServiceAccount` API instead, so #1365 can be closed.

PR #1881, opened 2026-06-26
Removed brackets in the example of phoenix

zyantw:
Preview: https://deploy-preview-1881--adk-docs-preview.netlify.app/integrations/phoenix/

Removing the brackets in this example that was triggering a mistake in the link check of the branch for my pull request: https://github.com/google/adk-docs/pull/1875


koverholt:
The example link that this PR is attempting to fix is already handled/ignored by the .lycheeignore rule here:
https://github.com/google/adk-docs/blob/main/.lycheeignore#L13-L14.

This link validator CI check was actually failing because `phoenix.arize.com` went offline, I updated the two places that the failing link was used in https://github.com/google/adk-docs/pull/1877/changes/027855e07d8fc3d5a5a6db805392610458b2e2a1 as part of #1877. So if you sync your existing PRs with main now using the "Update branch" button, the CI check will pass.

PR #1882, opened 2026-06-26
LangChain Tavily search tool new file path

jcwriter74:
Instead of adding the whole piece of code in the text, a file path was relocated: examples/python/snippets/tools/third-party/langchain_tavily_search.py (probably a 404 will appear, that file path doesn't appear on the main branch)

### DIFF
--- a/docs/tools/third-party/index.md
+++ b/docs/tools/third-party/index.md
@@ -72,6 +72,62 @@ Here's the full code combining the steps above to create and run an agent using 
 --8<-- "examples/python/snippets/tools/third-party/langchain_tavily_search.py"
 ```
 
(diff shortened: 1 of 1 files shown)


koverholt:
I don't see `docs/tools/third-party/index.md` or  `adk-docs/examples/python/snippets/tools/third-party/langchain_tavily_search.py` in adk-docs main, so I'm not sure if this is coming from an older checkout? These pages have been superseded by https://adk.dev/integrations/.

Also, we have removed all LangChain mentions/integrations from the docs as of #837 and search & scraping-related integrations from the docs as of #1084. So I recommend closing this PR, and closing any issues that mention LangChain or Tavily (or other search related tools) as "wontfix".

PR #1891, opened 2026-06-27
Add Perseus Context integration page

tcconnally:
## What

Adds an integration page for **[`adk-perseus-context`](https://pypi.org/project/adk-perseus-context/)** — a small ADK extension that injects a *deterministically compiled* context into an agent's system instruction, powered by [Perseus](https://github.com/Perseus-Computing-LLC/perseus), an open-source (MIT) context compiler.

- `docs/integrations/perseus.md` (catalog frontmatter: `catalog_title: Perseus Context`, `catalog_tags: ["code", "mcp"]`)
- `docs/integrations/assets/perseus.svg`

The integrations catalog auto-renders from the page frontmatter, so no nav/index changes are needed.

## How it works

Perseus resolves directives (`@file`, `@search`, `@memory`, `@include`) into one byte-stable context string at inference time — no retrieval index, no embeddings, no extra LLM round-trip. The package exposes two extension points, both verified against `google-adk` 2.3.0:

- `PerseusContextPlugin` — a `BasePlugin` whose `before_model_callback` appends the compiled context via `LlmRequest.append_instructions`, applied Runner-wide.
- `perseus_before_model_callback` — the per-agent `before_model_callback` variant.

It also documents consuming the Perseus MCP server through ADK's `McpToolset` (hence the `mcp` tag).

## Meets the review criteria

- **Completeness/testability**: `pip install adk-perseus-context` (on PyPI, `0.1.0`); the page's examples are runnable; the package has a CI-tested suite.
- **Developer value**: deterministic, index-free context with full coverage at a fixed size — complements (does not duplicate) the existing [Mimir Memory](https://adk.dev/integrations/mimir/) page, which covers the *memory* lane.
- **Publishability**: MIT-licensed, public repo and package.

Distinct from Mimir: Mimir = persistent memory backend; Perseus = deterministic context compiler. The page cross-links the two as a compose pair.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1892, opened 2026-06-28
docs(mimir): matched icon + add "mcp" catalog tag

tcconnally:
Two small updates to the existing Mimir Memory integration page:

1. **New icon** — replaces the placeholder with a flat, professionally designed mark (fading memory layers, violet `#8B5CF6` on `#161420`). It's one of a matched pair with the Perseus icon (proposed in #1891), so the two companion integrations read as a coherent set on the catalog grid. Renders crisply at tile size (no gradients).

2. **Adds `"mcp"` to `catalog_tags`** (`["data"]` → `["data", "mcp"]`). Mimir is an MCP (stdio) server — the `adk-mimir-memory` service speaks MCP JSON-RPC to the `mimir` binary — so it belongs under the catalog's MCP filter, where it also surfaces alongside its companion Perseus.

No prose or code changes; the page already documents the integration. Single-author commits under the signed CLA.

_Note: a reciprocal catalog link from this page to the Perseus integration page will follow once #1891 (which adds `docs/integrations/perseus.md`) is merged — kept out of this PR to avoid a dead internal link in the build._

### DIFF
--- a/docs/integrations/mimir.md
+++ b/docs/integrations/mimir.md
@@ -2,7 +2,7 @@
 catalog_title: Mimir Memory
 catalog_description: Add persistent, local, encrypted cross-session memory to ADK agents
 catalog_icon: /integrations/assets/mimir.svg
-catalog_tags: ["data", "mcp"]
+catalog_tags: ["data"]
 ---
 
 # Mimir Memory integration for ADK
--- a/docs/integrations/mimir.md
+++ b/docs/integrations/mimir.md
@@ -2,7 +2,7 @@
 catalog_title: Mimir Memory
 catalog_description: Add persistent, local, encrypted cross-session memory to ADK agents
 catalog_icon: /integrations/assets/mimir.svg
-catalog_tags: ["data"]
+catalog_tags: ["data", "mcp"]
 ---
 
 # Mimir Memory integration for ADK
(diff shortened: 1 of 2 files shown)


docs/integrations/mimir.md
```diff
@@ -2,7 +2,7 @@
 catalog_title: Mimir Memory
 catalog_description: Add persistent, local, encrypted cross-session memory to ADK agents
 catalog_icon: /integrations/assets/mimir.svg
-catalog_tags: ["data"]
+catalog_tags: ["data", "mcp"]
```
koverholt: This integration does not have any user-facing MCP functionality like other MCP integrations that use the `McpToolset` in ADK to connect to local or remote MCP servers. Therefore the `mcp` catalog tag here is not justified and should not be added.

koverholt (changes requested):

tcconnally:
Good point — thanks for the catalog guidance. The integration implements `BaseMemoryService` rather than wiring up an `McpToolset`, so the MCP transport is internal to `adk-mimir-memory` (it speaks JSON-RPC to the `mimir` binary) and never surfaces to the ADK developer. That makes the `mcp` filter misleading here, so I've reverted the tag back to `["data"]`.

The remaining change is just the matched flat icon. Re-requesting review — thanks!

koverholt (approved):
Logo update LGTM. Thanks! Merging.

PR #1897, opened 2026-06-29
 #1350 - 10 SessionNotFoundError exception

zyantw:
- Rendered view: https://deploy-preview-1897--adk-docs-preview.netlify.app/sessions/session/#troubleshoot-session-errors
- Agent pull request: https://github.com/google/adk-docs/pull/1370
- Original Issue: https://github.com/google/adk-docs/issues/1350


docs/sessions/session/index.md
```diff
@@ -399,3 +399,21 @@ during a conversation turn:
 
 This cycle highlights how the `SessionService` ensures conversational continuity
 by managing the history and state associated with each `Session` object.
+
+## Troubleshoot session errors
+
+`SessionNotFoundError`: During execution, ADK can raise specific exceptions to help you identify configuration or state issues.
```
zyantw: This is not an agent suggestion, just something I noticed, should I add `AsyncDriverError` and `SessionLockError` to this list?

docs/sessions/session/index.md
```diff
@@ -399,3 +399,21 @@ during a conversation turn:
...
+Raised when a runner attempts to access or execute a session that does not exist in the active session store. For backward compatibility,
+this exception inherits from `ValueError`.
+
+* **Common Causes:** Passing an invalid, expired, or missing session_id. Attempting to run a session before you create it.
+  
+* **How to resolve:** Verify the session_id is correctly passed from your client-side storage. Use `list_sessions()` to audit existing active IDs.
+
+```python
+
+  class SessionNotFoundError(ValueError):
```
zyantw: lines 24 and 25: https://github.com/google/adk-python/blob/d2ebacb9d7e0f8597bdcbde845e6f5c536c352b7/src/google/adk/errors/session_not_found_error.py#L5

docs/sessions/session/index.md
koverholt: The `AsyncDriverError` and `SessionLockError` error classes don't actually exist anywhere in ADK. They look like they were derived from the "Async Driver Requirement" and "Concurrency and locking" sections on this page rather than actual exception classes.

There are other real relevant classes (`AlreadyExistsError` and `StateSchemaError`), but let's not document those here, we can keep this PR scoped to `SessionNotFoundError` only as mentioned by the agent issue. We can expand the list in a follow-up PR/issue if we decide to.

docs/sessions/session/index.md
```diff
@@ -399,3 +399,21 @@ during a conversation turn:
...
+  def __init__(self, message="Session not found."):
+    super().__init__(message)
+
+```
```
koverholt: This snippet reproduces ADK's internal class definition, which users should not need to define themselves and will be confusing. A troubleshooting entry should tell the user how to resolve/prevent the error, not show the exception's internals.

Let's remove the class-definition block and instead make the "How to resolve" prose concrete and actionable:

- Ensure the session exists before running it, by creating it first with `create_session(...)`.
- Or construct the `Runner` with `auto_create_session=True`, which is the flag that prevents this error from being raised (it auto-creates the session if it's missing).

No code sample is needed here, clear prose covering those two resolution paths is enough.

koverholt: FYI only, since we're planning to drop this class-definition snippet entirely, the following don't need fixing, I'm just noting them for future edits/PRs/checks.

1. Invalid indentation: `def __init__` is at the same indent level as `class SessionNotFoundError`, so it isn't actually inside the class and won't parse as valid Python.
2. Stray formatting: there's a leading blank line and an odd base indent on the whole block.
3. Message mismatch: the default `"Session not found."` here isn't what users actually see at runtime, the raised message is `Session not found: {session_id}`

docs/sessions/session/index.md
```diff
@@ -399,3 +399,21 @@ during a conversation turn:
...
 This cycle highlights how the `SessionService` ensures conversational continuity
 by managing the history and state associated with each `Session` object.
+
+## Troubleshoot session errors
+
+`SessionNotFoundError`: During execution, ADK can raise specific exceptions to help you identify configuration or state issues.
+Raised when a runner attempts to access or execute a session that does not exist in the active session store. For backward compatibility,
+this exception inherits from `ValueError`.
+
+* **Common Causes:** Passing an invalid, expired, or missing session_id. Attempting to run a session before you create it.
```
koverholt: `session_id` here isn't wrapped in backticks, so it renders as plain prose instead of inline code. Please format it as  `session_id`  for consistency with the other code identifiers on the page. Applies to both the "Common Causes" and "How to resolve" bullets.

docs/sessions/session/index.md
```diff
@@ -399,3 +399,21 @@ during a conversation turn:
...
+  
+* **How to resolve:** Verify the session_id is correctly passed from your client-side storage. Use `list_sessions()` to audit existing active IDs.
```
koverholt: There are blank lines with trailing spaces between these bullet items, which forces a "loose" list and renders with extra vertical spacing. Please remove the blank lines (and trailing whitespace) between the bullets so it renders as a proper list.

docs/sessions/session/index.md
koverholt: Let's structure this so it's a clean, copy-pasteable block that others can repeat when adding future errors. Put the section-wide intro under the `## Troubleshoot session errors heading`, then give each error its own `###` subheading with a consistent set of fields (description, common causes, how to resolve). As in:

> ## Troubleshoot session errors
>
> During execution, ADK can raise specific exceptions to help you identify
configuration or state issues.
>
> ### `SessionNotFoundError`
>
> Raised when a runner attempts to access or execute a session that does not
exist in the active session store. Inherits from `ValueError` for backward
compatibility.
>
> - **Common causes:** an invalid, expired, or missing `session_id`; running a
  session before it has been created.
> - **How to resolve:** ensure the session exists first via `create_session(...)`,
  or construct the `Runner` with `auto_create_session=True`.

That way, adding the next error later is just copying the `### ErrorName` block.

Note this also fixes the current sentence ordering, the general "ADK can raise specific exceptions…" intro should lead, then the per-error description, rather than being interleaved as it is now.

koverholt (changes requested):
Thanks for the PR and updates! The content is accurate and I verified it against `adk-python`. Two main requests:

- Restructure into a repeatable template (per-error `###` subheading).
- Drop the internal class-definition snippet in favor of resolution prose.

I left inline comments with the details.

docs/sessions/session/index.md
zyantw: Thanks for this. Taking note.

koverholt (approved):
Thanks for the edits, content LGTM. I made a commit to clean up a bunch of things on the page unrelated to the original PR.

Some changes that were related to the original PR include: fixed line wrapping, fixed sentence case in a bullet list label, and removed a trailing space on the `SessionNotFoundError` heading. Merging.

Issue #1438, opened 2026-03-16
Found docs updates needed from ADK python release v1.26.0 to v1.27.0


xuanyang15:
### 1. **Document new A2A capabilities: Reasoning, Long-Running Tools, and Artifacts**

**Doc file**: docs/a2a/intro.md

**Current state**:
> The introduction covers the "What", "When", and "How" of A2A but does not detail the specific capabilities supported by the protocol implementation, such as reasoning models or artifact streaming.

**Proposed Change**:
> Add a new section "Supported Capabilities" that outlines the advanced features supported by the ADK A2A implementation:
> 1. **Reasoning Models**: Support for transmitting "thought" traces from reasoning models (like Gemini 2.0 Flash Thinking).
> 2. **Long-Running Tools**: Built-in handling for tools that require significant time to execute.
> 3. **Artifact Streaming**: Support for streaming large data or files as artifacts between agents.

**Reasoning**:
The new converter files (`part_converter.py`, `long_running_functions.py`, `from_adk_event.py`) introduce support for these advanced capabilities. Users building complex agent systems need to know that the A2A protocol in ADK supports these features.

**Reference**: src/google/adk/a2a/converters/part_converter.py

### 2. **Document the new A2A agent configuration and interception capabilities**

**Doc file**: docs/a2a/quickstart-consuming.md

**Current state**:
> The consuming quickstart shows basic usage of `RemoteA2aAgent` with just a name, description, and agent card URL. It does not mention any advanced configuration or interception capabilities.

**Proposed Change**:
> Add an 'Advanced Configuration' section that explains how to use `A2aRemoteAgentConfig` to customize the agent's behavior. This should include:
> 1. Importing `A2aRemoteAgentConfig` and `RequestInterceptor`.
> 2. Creating an interceptor to modify requests or responses (e.g., adding headers).
> 3. Passing the configuration to `RemoteA2aAgent`.

**Reasoning**:
The new `A2aRemoteAgentConfig` and `RequestInterceptor` classes provide powerful capabilities for customizing A2A interactions, such as adding authentication headers or logging. Users need to know how to use these features.

**Reference**: src/google/adk/a2a/agent/config.py

### 3. **Create documentation for the new Skills framework and SkillToolset**

**Doc file**: docs/tools/built-in-tools/skills.md

**Current state**:
> (New file)

**Proposed Change**:
> # Skills
> 
[...]

PR #1912, opened 2026-06-30
#1438 - 1 A2A capabilities: Reasoning, Long-Running Tools, and Artifacts

zyantw:
- Rendered view: https://deploy-preview-1912--adk-docs-preview.netlify.app/a2a/intro/#supported-capabilities-in-a2a
- Agent Pull: https://github.com/google/adk-docs/pull/1439
- Original Issue: https://github.com/google/adk-docs/issues/1438


docs/a2a/intro.md
```diff
@@ -83,15 +83,26 @@ works:
...
+## Supported capabilities in A2A
+
+The A2A implementation provides three core features for complex agentic systems. 
+Each feature is powered by a specific module that delivers the necessary functionality to handle data transformation and state management.
+
+| Feature | Module | Functionality | In simpler terms |
+| :---- | :---- | :---- | :---- |
+| **Reasoning & Artifacts** | `part\_converter.py` | Encodes and decodes thought traces and file artifacts between A2A and the model SDK, ensuring data integrity | The Translator: Resolves language barriers between internal agent logic and the AI model SDK. |
+| **Long-Running Tools** | `long\_running\_functions.py` | Manages asynchronous state for tool calls that exceed standard response times to prevent timeouts. | The Manager: Coordinates asynchronous operations, maintaining conversation flow while waiting for external input. |
+| **Event Ingestion** | `from\_adk\_event.py` | Transforms raw platform events into structured A2A objects, serving as the ingestion point for all incoming agent commands. | The Receptionist: Orchestrates incoming traffic, transforming raw signals into actionable commands.|
```
koverholt: Replace this table with a simple "Supported capabilities" list of the three features from the issue: Reasoning, Long-Running Tools, and Artifacts. Give each a one-line description of what it enables for the user, and drop the module information (there are internal SDK files users don't import) and the "In simpler terms" analogies (too informal for reference docs). A short list keeps it accurate, in scope, and user-focused.

Note: the current "Event Ingestion" row also describes `from_adk_event.py` backwards, it converts ADK events *into* A2A events (via `convert_event_to_a2a_events()`), it's not an ingestion point for incoming commands. Another reason to drop that row and stick to the three requested capabilities. There are also other issues in the table such as the backslashes for the module names should be removed, and there is trailing whitespace, but these issues can be ignored since we should remove the table and that content.

docs/a2a/intro.md
```diff
@@ -61,14 +61,14 @@ Consider using **A2A** when:
   require independent deployment or complex state management, a simple function
   or class within the same agent is more appropriate than a separate A2A agent.
 
-## The A2A Workflow in ADK: A Simplified View
+## The A2A workflow in ADK: a simplified wiew
```
koverholt: Typo: `wiew` should be `view`.

koverholt (changes requested):
Thanks! Verified against `adk-python`. Main ask: replace the capabilities table with a simple list of the three features from the issue (Reasoning, Long-Running Tools, Artifacts), dropping the internal module column and the analogies. Plus a typo fix inline. Once that's in I'll re-review.

docs/a2a/intro.md
zyantw: Understood! Thanks for helping me notice these nuances. Taking note, and I will keep my updates simpler.

koverholt (approved):
Thanks for working through the issues. I pushed a commit to clean up the section a bit: dropped the leftover "powered by a specific module" sentence (leftover from the old table), fixed the bullets and wrapping to match the rest of the page, and reworded the descriptions to be a bit more user-facing.

All content LGTM, merging!

PR #1917, opened 2026-07-01
Update API reference docs for ADK Python 2.3.0

koverholt:
Generated using `bash tools/python-api-docs/generate.sh` in this repo.


docs/api-reference/python/py-modindex.html
```diff
@@ -205,7 +205,7 @@
       
       <div class="sidebar-sticky"><a class="sidebar-brand" href="index.html">
   
-  <span class="sidebar-brand-text">Agent Development Kit 2.0.0 documentation</span>
+  <span class="sidebar-brand-text">Agent Development Kit 2.3.0 documentation</span>
```
ToniCorinne: Just a general nit that we might want to consider - specifically adding "Python" to this text. Just to really highlight what language ref page users ended up on.

ToniCorinne (approved):
lgtm - one small suggestion but definitely not a blocker

docs/api-reference/python/py-modindex.html
koverholt: Good suggestion, I added this to the generation script and built output, so the sidebar and page title now have `(Python`) displayed.

PR #1921, opened 2026-07-01
Update ADK Kotlin to 0.4.0

koverholt:
Updates ADK Kotlin references for the [0.4.0 release](https://github.com/google/adk-kotlin/releases/tag/v0.4.0).

- Bump versions from 0.2.0 to 0.4.0 across docs and examples
- Update Kotlin snippets to the 0.4.0 `App` API (plugins and resumability config now live on `App`, replacing the deprecated `InMemoryRunner` parameters)
- Update API reference docs using `bash tools/kotlin-api-docs/generate.sh 0.4.0` in this repo


examples/kotlin/snippets/runtime/RunConfigExample.kt
ToniCorinne, line 1: General question - when introducing new packages like this, would we want to bump the language support tag on the docs page as well?

examples/kotlin/snippets/observability/SetupExample.kt
ToniCorinne, line 1: Same language support tag question here

ToniCorinne (commented):
I do have a best practices question wrt samples embedded in docs pages. 

Also, I think this might need to be it's own issue/PR to fix, but just an FYI: when looking at the deploy preview, the Kotlin reference page originally was redirecting back to the live site version (seems to be working now, so maybe it just wasn't quite rendered when the deploy link was live?) The /api-reference/kotlin//api-reference/kotlin/google-adk-kotlin-examples-android/ page is rendering as a 404 however. Unclear if those two things might be related.

examples/kotlin/snippets/runtime/RunConfigExample.kt
koverholt: I've had the same question as I've been updating code samples for new API. For now I'll stick with the convention that the language support version tracks the version a feature first shipped, not the API a snippet uses.

RunConfig/observability have existed since Kotlin v0.1.0, so no bump. This PR only swaps the deprecated `InMemoryRunner(...)` params for the current `App` API. (New 0.4.0 features like litert-lm are correctly tagged v0.4.0.)

examples/kotlin/snippets/observability/SetupExample.kt
koverholt: I've had the same question as I've been updating code samples for new API. For now I'll stick with the convention that the language support version tracks the version a feature first shipped, not the API a snippet uses.

RunConfig/observability have existed since Kotlin v0.1.0, so no bump. This PR only swaps the deprecated `InMemoryRunner(...)` params for the current `App` API. (New 0.4.0 features like litert-lm are correctly tagged v0.4.0.)

koverholt:
Thanks for the review!

> Also, I think this might need to be it's own issue/PR to fix, but just an FYI: when looking at the deploy preview, the Kotlin reference page originally was redirecting back to the live site version (seems to be working now, so maybe it just wasn't quite rendered when the deploy link was live?) The /api-reference/kotlin//api-reference/kotlin/google-adk-kotlin-examples-android/ page is rendering as a 404 however. Unclear if those two things might be related.

Yes I've also noticed these breakages that seem to be that Dokka-generated docs don't play well specifically with the Netlify preview and causes those duplicated paths and 404s only in Netlify preview. Might have something to do with differences in root path or relative path behavior in mkdocs vs. Netlify?

I've seen similar behavior in previous Kotlin API ref doc updates, but I can confirm that in those cases and this PR that everything works locally as expected and in the built version as expected (you can double check after merging and see that the Netlify preview behavior cannot be reproduced in prod).

koverholt:
@ToniCorinne, preview build problem solved!

Root cause was not Dokka's root URLs but rather Netlify's Pretty URLs post-processing rewriting Dokka's relative `navigation.html` links into absolute ones on the preview; Dokka's `navigation-loader.js` then prepended its `pathToRoot`, producing the doubled `/api-reference/kotlin//api-reference/kotlin/...` and the 404.

Prod (GitHub Pages) and local mkdocs serve serves the files untouched, which is why it only broke on Netlify previews.

So I disabled that URL rewriting in `netlify.toml`:

```
[build.processing.html]
  pretty_urls = false
```

Verified the Kotlin API-ref navigation now works as expected in the Netlify deploy previews.

koverholt:
@ToniCorinne, this is ready for re-review/approval when you are able. Thanks!

ToniCorinne (approved):
Preview  looks great - thanks for fixing that!

PR #1859, opened 2026-06-23
chore: update ADK Java doc to version 1.5.0

tilgalas:

PR #1922, opened 2026-07-01
Update ADK Java dependency versions to 1.5.0

koverholt:
Bumps all ADK Java dependency versions to 1.5.0 across docs and example projects to reflect the latest ADK Java release.

Related to #1859.


docs/integrations/firestore-session-service.md
```diff
@@ -33,7 +33,7 @@ ADK provides a native integration for managing persistent agent session states u
 
     Ensure you use the same version for both `google-adk` and `google-adk-firestore-session-service` to guarantee compatibility.
 
-Add the following dependencies to your `pom.xml` (Maven) or `build.gradle` (Gradle), replacing `1.4.0` with your target ADK version:
+Add the following dependencies to your `pom.xml` (Maven) or `build.gradle` (Gradle), replacing `1.5.0` with your target ADK version:
```
ToniCorinne: nit - this seems like we're encouraging folks to downgrade, rather than use the most-up-to-date version of java adk. Is this line necessary here, or would it make more sense as a note? Reading this page it seems like this is more related to the firestore compatibility note above.

ToniCorinne (approved):
One nit on page clarity, but not a blocker for this.

docs/integrations/firestore-session-service.md
koverholt: Thanks for the feedback. Agree, I had been meaning to clean up this language. Fixed in 819a834.

PR #1925, opened 2026-07-02
docs(a2a): document A2A SDK 0.3/1.0 compatibility

mykytanetipa:
This PR adds a new documentation page, docs/a2a/a2a-sdk-compatibility.md, explaining how ADK's A2A integration remains compatible with both major versions of the A2A Python SDK (a2a-sdk 0.3.x and 1.x.x) without requiring changes to ADK application code.


koverholt (commented):
Thanks @mykytanetipa for the PR and additions. The technical content is accurate against a2a-sdk 1.x, and the deprecation framing is clear.

Before we review, I'd like to understand a few things:

**Release timing / does the described behavior exist in a released ADK yet?**

As far as I can tell from public adk-python:

- `_compat.py` isn't in main or any released ADK Python version
- `pyproject.toml` still pins `a2a-sdk>=0.3.4,<0.4`
- Issue https://github.com/google/adk-python/issues/5056 is still open and PR https://github.com/google/adk-python/pull/5917 was closed.


Is this blocked by some future release of ADK Python? And do you know the timing?

**Audience / scope: is this user documentation or an internal design doc?**

The page documents the private module `_compat.py` which is never imported by application code. There is no action in user's code, and much of this doc overlaps with the existing a2a-python v1.0 migration guide.

Would it make more sense to have only a short user-facing note on the existing A2A overview page that links to the migration guide rather than a new top-level page in ADK docs that we need to maintain, which mainly focuses on a2a-sdk internals?

I have many other comments on the proposed doc as well, but those only really apply if we need to publish / maintain these a2a-sdk docs in the ADK docs, which I would recommend against and would instead recommend just including a callout and link on an existing A2A page in these docs.

mykytanetipa:
Thanks @koverholt. Quick update on the two blocking points you raised:

Released / version constraint: Both are now resolved in ADK Python 2.5.0:
- _compat.py is now in the source code: src/google/adk/a2a/_compat.py (https://github.com/google/adk-python/blob/v2.5.0/src/google/adk/a2a/_compat.py)
- The a2a-sdk pin has been lifted from >=0.3.4,<0.4 to >=0.3.4,<2 (https://github.com/google/adk-python/blob/v2.5.0/pyproject.toml), so both 0.3.x and 1.x.x are now supported.

Scope: I took your recommendation. Rather than a standalone page documenting the a2a-sdk internals, I've reduced this PR to a small informational callout on the existing Introduction to A2A (docs/a2a/intro.md) page that notes the 0.3.x/1.x.x compatibility and links out to the a2a-python v1.0 migration guide. The standalone a2a-sdk-compatibility.md page has been removed.

koverholt (approved):
Thanks for the update and rescoping this to a concise compatibility note and pointing users to the migration guide for the full details. I confirmed the dual-version support has since shipped in ADK 2.5.0, so the note is accurate.

I made a commit to move the note to the two A2A Python quickstart pages, since that's where users will interact with the code, and the intro page is conceptual and not tied to any particular language SDK (Python, Go, Java). LGTM, merging!

PR #1929, opened 2026-07-05
Docs/1492 visual builder security

vivianagamezm-write:
Main issue: https://github.com/google/adk-docs/issues/1492
Rendered preview: https://deploy-preview-1929--adk-docs-preview.netlify.app/visual-builder/


koverholt (approved):
Thanks for the PR and contribution! I verified the security behavior against adk-python and it's accurate.

I pushed a few edits on top of your changes:

- Security section: consolidated it into a single paragraph and reframed the gate as "available only when the web UI is served." The earlier wording implied `adk deploy cloud_run` always disables the builder endpoints, but it depends on whether the web UI is served, so a deployment with `--with_ui` would still expose them. The new wording is correct for every deployment mode.
- Generated project structure: removed the redundant subheading and the per-file description list, which was more detail than this page focused on Visual Builder needs.
- Other formatting and page cleanup unrelated to the original issue/PR

LGTM, merging!

vivianagamezm-write:
Thanks so much for the review and for the clarifying edits. The distinction around the web UI makes sense, and I appreciate the cleanup to keep the page focused. Glad to contribute!

PR #1936, opened 2026-07-07
docs: update more Go documentation and examples to ADK Go v2

hyangah:
This PR updates the documentation to use ADK Go v2.

### Changes
- Updated Go import paths from `google.golang.org/adk/...` to `google.golang.org/adk/v2/...` across 22
  documentation files to align with the v2 release.
- Updated Go installation commands to reference `v2`.
- Fixed `session.NewEvent` API calls in `docs/runtime/event-loop.md` to include the required `context.Context`
  argument as introduced in v2.
- Verified that all Go examples in the repository compile and pass tests with ADK Go v2.

### Exceptions Preserved
- Kept references to `v1` in the 2.0 migration guide, installation tips for v1 users, and reference links to v1
  API docs

### DIFF
--- a/docs/_includes/homepage/_hero.md
+++ b/docs/_includes/homepage/_hero.md
@@ -40,7 +40,7 @@ agent = <span class="fn">Agent</span>(
 
 </pre></div>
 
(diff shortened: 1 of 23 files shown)


koverholt (approved):
Thanks for the follow-up! Verified against adk-go v2 and everything is correct and complete. Good catch on `NewEvent` as well. LGTM, merging.

PR #1942, opened 2026-07-09
Enhance BigQuery Agent Analytics with new options

haiyuan-eng-google:
Added options for OpenTelemetry correlation, custom metadata capture, and payload column projection in BigQuery Agent Analytics configuration.

### DIFF
--- a/docs/integrations/bigquery-agent-analytics.md
+++ b/docs/integrations/bigquery-agent-analytics.md
@@ -569,6 +569,9 @@ account) under which the agent is running needs these Google Cloud roles:
     | `auto_schema_upgrade` | `bool` | `True` | Automatically add new columns to existing tables (additive only) |
     | `create_views` | `bool` | `True` | Create per-event-type BigQuery views (1.27.0+) |
     | `view_prefix` | `str` | `"v"` | Avoid view-name collisions when multiple plugins share a dataset (e.g., `"v_staging"`) |
+    | `enable_otel_correlation` | `bool` | `False` | Capture the ambient OpenTelemetry span context into `attributes.otel.{span_id, trace_id}` as a best-effort Cloud Trace join key (v2.4.0+) |
+    | `custom_metadata_allowlist` | `Optional[List[str]]` | `None` | Capture selected `event.custom_metadata` keys into `attributes.custom_metadata.*` — exact keys or `"prefix*"` patterns (v2.4.0+) |
+    | `payload_column_denylist` | `Optional[List[str]]` | `None` | Project payload columns (`content`, `content_parts`, `attributes`, `latency_ms`) out of the table at write time (v2.4.0+) |
 
 
     The following code sample shows how to define a configuration for the BigQuery
@@ -618,6 +621,42 @@ account) under which the agent is running needs these Google Cloud roles:
         project_id="my-project",
         dataset_id="my_dataset",
         config=config,
+    )
+    ```
+
+    ### Trace correlation, metadata capture, and column projection (v2.4.0+)
+
+    Three options control what extra context lands in `attributes` — and whether
+    payload columns are written at all:
+
+    - **`enable_otel_correlation`** — when `True`, each row captures the ambient
+      OpenTelemetry span context at emission time into `attributes.otel.span_id`
+      and `attributes.otel.trace_id`. Use it to join `agent_events` rows against
+      Cloud Trace spans. This is a best-effort correlation key, not a foreign
+      key; when disabled (the default) no `attributes.otel` is written.
+    - **`custom_metadata_allowlist`** — captures selected keys from an event's
+      `custom_metadata` into `attributes.custom_metadata.*`. Entries are exact
+      keys or prefix patterns ending in `*` (for example `"exp:*"`). Captured
+      values pass the same safety pipeline as all other logged content
+      (truncation, sensitive-key redaction, circular-reference handling).
+      Leaving it unset preserves the previous behavior, where only the built-in
+      `a2a:*` capture runs.
+    - **`payload_column_denylist`** — projects payload columns out of the table
+      at write time, for deployments that must not persist payloads. Only
+      `content`, `content_parts`, `attributes`, and `latency_ms` may be listed;
+      identity and correlation columns are protected and raise `ValueError` if
+      listed. The projection is applied schema-first, so the table schema, the
+      written rows, and the auto-created views stay consistent — views drop
+      derived columns that depend on a denied column. Note that denying
+      `attributes` also disables `attributes.otel` and
+      `attributes.custom_metadata`, and combining it with a non-empty
+      `custom_metadata_allowlist` is rejected at construction.
+
+    ```python
+    config = BigQueryLoggerConfig(
+        enable_otel_correlation=True,                      # join key against Cloud Trace
+        custom_metadata_allowlist=["ticket_id", "exp:*"],  # capture selected custom_metadata keys
+        # payload_column_denylist=["content_parts"],       # don't persist multimodal payloads
     )
     ```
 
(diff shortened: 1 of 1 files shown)


docs/integrations/bigquery-agent-analytics.md
```diff
@@ -621,6 +624,42 @@ account) under which the agent is running needs these Google Cloud roles:
     )
     ```
 
+    ### Trace correlation, metadata capture, and column projection (v2.4.0+)
```
koverholt: Calling out specific version numbers directly in headings and doc text is not a good convention and is not very maintainable. Use language support tags, which are already used as the convention for specifying when a feature was added to a certain library version. So under this header, use:

```
  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v2.4.0</span>
</div>
```

Side note, does this section need to exist at all given that these options are already documented in the table above? None of the other options have separate sections like this with duplicate entries/explanations.

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -823,6 +862,11 @@ builds that include it (not in `v2.2.0`). On a 1.27.0+ release without that
 support, these four views are not created and `v_tool_completed` does not have
 the `pause_kind` / `function_call_id` columns.
 
+‡ Added in v2.4.0: extracted from `attributes.usage_metadata.thoughts_token_count`
+and `attributes.usage_metadata.tool_use_prompt_token_count`, so thinking-model
+reasoning tokens and tool-use prompt tokens can be tracked (and costed)
+separately from ordinary prompt/completion tokens.
+
```
koverholt, line 852: This PR introduces a bunch of version specific callouts in notes like this and throughout the page, and this convention has not been used before on this page and is not maintainable in the long term. Suggest removing this non-conventional footnote and instead using language support tags in the appropriate section to indicate when a feature first became available.

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -823,6 +862,11 @@ builds that include it (not in `v2.2.0`). On a 1.27.0+ release without that
 support, these four views are not created and `v_tool_completed` does not have
 the `pause_kind` / `function_call_id` columns.
```
koverholt: See other comment about removing specific version callouts / unconventional footnotes like this: https://github.com/google/adk-docs/pull/1942/changes#r3553126426

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -569,6 +569,9 @@ account) under which the agent is running needs these Google Cloud roles:
     | `auto_schema_upgrade` | `bool` | `True` | Automatically add new columns to existing tables (additive only) |
     | `create_views` | `bool` | `True` | Create per-event-type BigQuery views (1.27.0+) |
     | `view_prefix` | `str` | `"v"` | Avoid view-name collisions when multiple plugins share a dataset (e.g., `"v_staging"`) |
+    | `enable_otel_correlation` | `bool` | `False` | Capture the ambient OpenTelemetry span context into `attributes.otel.{span_id, trace_id}` as a best-effort Cloud Trace join key (v2.4.0+) |
+    | `custom_metadata_allowlist` | `Optional[List[str]]` | `None` | Capture selected `event.custom_metadata` keys into `attributes.custom_metadata.*` — exact keys or `"prefix*"` patterns (v2.4.0+) |
+    | `payload_column_denylist` | `Optional[List[str]]` | `None` | Project payload columns (`content`, `content_parts`, `attributes`, `latency_ms`) out of the table at write time (v2.4.0+) |
```
koverholt: The three options are accurate, but the config table now has per-row version tags across several eras. Suggest removing the version callouts from this table, which is not maintainable in the long term, and instead using language support tags w/ versions when a feature was first supported in a given version of ADK SDKs: https://github.com/google/adk-docs/pull/1942/changes#r3553126426.

docs/integrations/bigquery-agent-analytics.md
koverholt: This block is propagating the version callout problem the other comments flag, and it's now stale (the "latest release is v2.2.0" claim is wrong since v2.4.0 shipped). Since the whole note is version callouts, suggest removing it and instead indicating minimum versions with language support tags in each relevant section: https://github.com/google/adk-docs/pull/1942/changes#r3553126426.

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -870,6 +914,34 @@ instructions.
...
+        "name": "list_dataset_ids",
+        "description": "Fetches BigQuery dataset ids present in a GCP project.",
+        "parameters": {
+          "type": "object",
+          "properties": {"project_id": {"type": "string"}},
+          "required": ["project_id"]
+        }
+      }
+    ]
+    ```
```
koverholt: Same version callout problem as the other comments. Here "post-v2.4.0" / "newer than v2.4.0" is referring to a version that has not been released yet, so it's confusing and can't be verified by a user today. If you want to document this code change, wait until it's in a release and show the new tools shape under a language support tag block with that specific version (e.g. Python vX.Y.Z) rather than a "post-X.Y.Z" note: https://github.com/google/adk-docs/pull/1942/changes#r3553126426

koverholt (changes requested):
Thanks for the update. The additions are technically accurate, but there are a a few convention issues (details inline): version numbers are hard-coded throughout instead of using language support tags, and one note documents "post-v2.4.0", a version that isn't out yet.

docs/integrations/bigquery-agent-analytics.md
haiyuan-eng-google: Done in a026dbb5. Converted the page from ad-hoc version callouts to the repo-standard `language-support-tag` convention throughout:

- Removed this `‡ Added in v2.4.0` footnote and its `‡` markers. `usage_thinking_tokens` / `usage_tool_use_tokens` stay in the view column list without a version callout (they're in v2.4.0).
- Minimum versions are now expressed with `language-support-tag` blocks on the relevant sections instead of inline suffixes/footnotes: Automatically Created Views → Python v1.27.0, ADK 2.0 workflow events → Python v2.3.0, Trace correlation/metadata/projection → Python v2.4.0.
- Removed the two stale "Version Requirement" notes (top of page and ADK 2.0 section) and the `†` "not in v2.2.0" footnote, and de-versioned the intro prose.
- Removed the "post-v2.4.0" structured-tool-declarations note since it documents unreleased behavior (details in the thread on that block).

The other five threads point back here.

docs/integrations/bigquery-agent-analytics.md
haiyuan-eng-google: Done in a026dbb5. Dropped `(v2.4.0+)` from the header and added a `language-support-tag` (Python v2.4.0) block under it, matching the convention.

On the side note about the section duplicating the table: I kept it but trimmed it so it no longer restates what each option does. It now carries only the cross-option rules the flat table can't express — protected identity/correlation columns raising `ValueError`, schema-first projection keeping table/rows/views consistent, denying `attributes` also disabling `attributes.otel`/`attributes.custom_metadata`, and `payload_column_denylist` + a non-empty `custom_metadata_allowlist` being rejected at construction — plus the combined config example. Happy to fold it entirely into the table and drop the section if you'd prefer.

docs/integrations/bigquery-agent-analytics.md
haiyuan-eng-google: Done in a026dbb5. Removed the `†` footnote and its markers. The four workflow views and the `pause_kind`/`function_call_id` columns now reference the ADK 2.0 section, which carries a `language-support-tag` (Python v2.3.0).

For context: this footnote was pre-existing (not added by this PR) and was also stale — the ADK 2.0 workflow events are no longer "on main only". They shipped in v2.3.0 (2026-06-18), which is why the tag reads v2.3.0.

docs/integrations/bigquery-agent-analytics.md
haiyuan-eng-google: Done in a026dbb5. Removed the `(v2.4.0+)` suffixes from the three rows (and `(1.27.0+)` from `create_views`). The v2.4.0 minimum is now shown once via a `language-support-tag` on the "Trace correlation, metadata capture, and column projection" section that documents these three options, rather than per-row in the table.

docs/integrations/bigquery-agent-analytics.md
haiyuan-eng-google: Done in a026dbb5. Removed this note. You're correct that it was stale — v2.4.0 shipped on 2026-07-07, so "v2.2.0, the latest release" is no longer true, and the ADK 2.0 workflow support it described as unreleased is in v2.3.0+.

The minimum-version info it carried now lives in section-level `language-support-tag` blocks: Automatically Created Views → Python v1.27.0, ADK 2.0 workflow events → Python v2.3.0, Trace correlation/metadata/projection → Python v2.4.0. I also de-versioned the intro prose that duplicated it.

(Note: this note was pre-existing rather than introduced by this PR, but it's cleaned up here since it's the same version-callout pattern.)

docs/integrations/bigquery-agent-analytics.md
haiyuan-eng-google: Done in a026dbb5 — removed the note. You're right that it documents unreleased behavior. I verified it against the tagged source: in v2.4.0 the plugin emits `attributes["tools"] = list(llm_request.tools_dict.keys())` (a list of tool names); the structured `{name, description, parameters}` object shape only exists on `main`. I'll re-add it under a `language-support-tag` with the concrete version once it lands in a tagged release.

koverholt (approved):
Updates LGTM, thanks for working through the issues. Merging.

PR #1943, opened 2026-07-09
Add agent skills for creating and reviewing integrations

koverholt:
Adds agent skills to this repo under `.agents/skills/`, plus contributor-facing references in `CONTRIBUTING.md`. These skills help draft and review integration pages in `docs/integrations/` following our established catalog conventions and best practices.


.agents/skills/integration-create/SKILL.md
```diff
@@ -0,0 +1,293 @@
...
+<div class="language-support-tag">
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python</span><span class="lst-typescript">TypeScript</span>
+</div>
+```
+
+---
+
+## Skeleton: MCP tool
+
+Mirror `github.md` / `notion.md`.
```
ToniCorinne: does this work with just the filename here, rather than the relative path? I tend to get agent spin when I do this so just curious if you've tested it.

.agents/skills/integration-create/SKILL.md
```diff
@@ -0,0 +1,293 @@
...
+- [<Platform> Documentation](<url>)
+- [<Platform> Repository](<url>)
+````
+
+A single copy-pasteable end-to-end code block is acceptable and often better
+than splitting Setup and Observe.
+
+## Skeleton: Plugin
+
+Mirror `daytona.md` / `goodmem.md`.
```
ToniCorinne: same question as above

.agents/skills/integration-review/SKILL.md
```diff
@@ -0,0 +1,272 @@
...
+  files that only show in the full diff.
+- Check whether "Allow edits from maintainers" is enabled
+  (`maintainerCanModify`) so fixes can be pushed directly if requested later.
+- Confirm the CLA is signed (the `google-cla` bot). An unsigned CLA is a hard
+  blocker.
+- If reviewing a local file, run `git status` and `git diff` to see the change.
+
+### Step 2: Read peer pages
+
+Open 5 or more existing pages of the **same category** as the page under review
```
ToniCorinne: Would it be more robust for the agent to explicitly call out the `catalog_tags: ["<tag>"]` here?

.agents/skills/integration-create/SKILL.md
koverholt: Good catch, not only that, but I included the relative path in the section at the top of this file, so I'll just remove all of these `Mirror X / Y` lines for simplicity. Thanks! Fixed in 28df6c2.

.agents/skills/integration-create/SKILL.md
koverholt: Good catch, not only that, but I included the relative path in the section at the top of this file, so I'll just remove all of these `Mirror X / Y` lines for simplicity. Thanks! Fixed in 28df6c2.

.agents/skills/integration-review/SKILL.md
ToniCorinne: Or would it be worthwhile for us to add in a list.txt file that the agent could scan quickly to identify which files would fulfill this requirement? rather than having to open/read each file until it has 5 or more?

.agents/skills/integration-review/SKILL.md
```diff
@@ -0,0 +1,272 @@
...
+  on a card. It must **not repeat the product name** (the title already shows
+  it). No verbose lists of technologies; describe what the integration does. No
+  overclaims.
+- **`catalog_icon`**: `/integrations/assets/<slug>.png` (or `.svg` or `.jpg`).
+  The referenced asset file must actually exist in `docs/integrations/assets/`
+  and be a real image.
+- **`catalog_tags`**: a JSON array of lowercase tags. **Use only tags that
+  already exist in the catalog; never invent a new one.** Tags combine (e.g.
+  `["data", "mcp"]`). Any page that uses MCP in its body must carry the `mcp`
+  tag. To see the valid tags, scan the `catalog_tags` field across
```
ToniCorinne: related to comment above.

.agents/skills/integration-review/SKILL.md
```diff
@@ -0,0 +1,272 @@
...
+
+---
+
+## Feedback tone (for PR comments you draft)
+
+- Short and direct. A couple of sentences per point.
+- Diplomatic and actionable; give the contributor an out (e.g. "Do you have a
+  sample agent in a repo? Otherwise you can remove this link and this page will
+  serve as the sample.").
+- Use precise terminology ("catalog title" for the frontmatter field, not
```
ToniCorinne: small nit --> "**rename the catalog title**  for the frontmatter field, not "rename the page"" to be more explicit.

CONTRIBUTING.md
```diff
@@ -55,6 +55,22 @@ We review contributions for integrations based on the following criteria:
...
 - **Use consistent formatting** with existing documentation pages
 
+## AI-Assisted Development
+
+This repo includes built-in skills for AI coding agents to help with ADK
+documentation contributions:
+
+- **`integration-create`:** Draft a new integration page under
+  `docs/integrations/`: gathers details, picks the right category template (MCP
+  tool, observability, or plugin), and follows the catalog conventions.
```
ToniCorinne: nit -> "etc." I'm assuming we aren't going to be limited to just those three.

ToniCorinne (approved):
Overall, lgtm. 

A few points for clarification, but no blockers.

CONTRIBUTING.md
koverholt: Fixed, thanks!

.agents/skills/integration-review/SKILL.md
koverholt: Fixed, thanks!

.agents/skills/integration-review/SKILL.md
koverholt: Good call. I reworked Step 2 to help the agent find candidates in one pass w/ `grep`, match the tag under review, widen with other signals (domain, template, language), then read those pages in full. Also made the `catalog_tags` frontmatter bullet use the same explicit grep.

I'll skip the `list.txt` manifest. The grep is already a one-shot scan of frontmatter, so the agent never opens every file just to find candidates, and a separate manifest would duplicate the frontmatter and drift as pages are added or retagged.

.agents/skills/integration-review/SKILL.md
koverholt: Fixed, thanks! Details in previous comment.

koverholt:
Thanks for the detailed review and walking through these!

PR #1945, opened 2026-07-10
Add Sprites integration

aezell:
Adds a Sprites integration page to the catalog.

[Sprites](https://sprites.dev) are persistent, stateful Linux sandboxes from Fly.io. Unlike ephemeral sandboxes, a Sprite keeps its filesystem, installed packages, and running processes between sessions, and it can **checkpoint and restore** its whole state — so an ADK agent can snapshot the environment before a risky change and roll back if it goes wrong. The [`sprites-adk`](https://github.com/superfly/sprites-adk) plugin exposes this to agents as tools.

## What's included
- `docs/integrations/sprites.md` — follows the standard integration format (frontmatter + Use cases, Prerequisites, Installation, Use with agent, Available tools, Additional resources).
- `docs/integrations/assets/sprites.png` — square catalog icon.

## Notes for reviewers
- The plugin is published and installable: `pip install sprites-adk` ([PyPI](https://pypi.org/project/sprites-adk/)).
- The `Use with agent` example and tool set are from the working package; the checkpoint/restore flow is exercised against a real Sprite.


docs/integrations/sprites.md
```diff
@@ -0,0 +1,85 @@
...
+- **Persistent development environments**: A named Sprite is reused across
+  sessions — packages and files from earlier runs are still there, so
+  long-running projects don't rebuild the environment from scratch each time.
+
+- **Secure code execution**: Run agent-generated Python, JavaScript, or bash in
+  an isolated microVM instead of on the host machine.
+
+- **Fearless experimentation**: Checkpoint the whole environment before package
+  upgrades, migrations, or bulk edits, then restore it if the change breaks
+  things.
```
koverholt: Small style nit: "Fearless experimentation" reads a little marketing-toned next to the other three functional headings ("Persistent development environments", "Secure code execution", "File workflows"). Consider "Safe experimentation" or "Checkpoint and roll back" or similar for consistency. Not blocking, your call.

docs/integrations/sprites.md
```diff
@@ -0,0 +1,85 @@
...
+plugin = SpritesPlugin(
+  # token="your-sprites-token"  # Or set the SPRITES_TOKEN environment variable
+)
+
+root_agent = Agent(
+    model="gemini-flash-latest",
+    name="sandbox_agent",
+    instruction="Run code and commands in the Sprite sandbox, not locally.",
+    tools=plugin.get_tools(),
+)
```
koverholt, line 67: There is one thing worth adding here: `SpritesPlugin` is a `BasePlugin`, so `plugin.get_tools()` wires up the tools but the plugin's lifecycle callbacks and cleanup only run if it's registered on the runner. Your README recommends the same. You can keep the example as-is and just add these two lines at the appropriate places:

```python
from google.adk.runners import InMemoryRunner

runner = InMemoryRunner(agent=root_agent, plugins=[plugin])
```

Optional and I'll defer to you, but it makes the code sample fully functional.

koverholt (commented):
Thanks for the integration and clean PR. I verified the package (sprites-adk on PyPI), the `SpritesPlugin` API, and all tool names against the source. Everything checks out.

I left a couple of optional comments if you want to update the code sample or use cases bullet, both are minor, non-blocking, and I'll defer to your preference. Once those are resolved we can re-review and move towards merging. Thanks!

docs/integrations/sprites.md
aezell: Good call — changed it to **Safe experimentation** so it matches the functional tone of the other three. Thanks!

docs/integrations/sprites.md
aezell: Added it — imported `InMemoryRunner` and register the plugin with `plugins=[plugin]` at the end of the example. Agreed it makes the sample fully functional and matches the lifecycle/cleanup behavior. Thanks!

koverholt (approved):
Updates LGTM, thanks for the quick followup edits. Everything LGTM, and thanks for contributing to the ADK ecosystem! Merging.

aezell:
Thanks for the quick review and the merge. I appreciate it! Have a great weekend.

PR #1950, opened 2026-07-10
Expose page Markdown at page.md instead of page/index.md

haranrk:
## Summary

Changes the per-page raw Markdown exposure from `page/index.md` to a flat `page.md` (e.g. `/agents/config/index.md` → `/agents/config.md`). HTML URLs and pretty directory URLs are unchanged; this is a full switch with no `page/index.md` output remaining.

## How it works

`use_directory_urls: true` means `mkdocs-llmstxt` writes each page's Markdown next to its HTML output (`site/agents/config/index.md`), and the plugin's output path isn't configurable. Rather than flip `use_directory_urls` (which would rewrite every HTML URL) or fork the plugin, a small post-build hook runs after it and flattens the result. This also works on GitHub Pages, which serves static files only (no redirects).

## Changes

- **`scripts/flatten_md.py`** (new):
  - `flat_md_target()` / `rewrite_index_links()` — pure helpers mapping `<dir>/index.md` → `<dir>.md` and rewriting site-internal `.../index.md` link targets to the flat form.
  - `on_post_build` hook (`@event_priority(-100)`, so it runs after `mkdocs-llmstxt`) that renames the generated Markdown files and rewrites links in the generated `*.md` files plus `llms.txt` / `llms-full.txt`.
- **`mkdocs.yml`** — registers the hook via top-level `hooks:`.
- **`overrides/partials/actions.html`** — the "view as Markdown" link and "copy page as Markdown" button now point at the flat `../<page>.md`.
- **`scripts/test_flatten_md.py`** (new) — unit tests for the helpers plus a `tmp_path` integration test for `on_post_build`.

## Behavior

| Pretty URL | Before | After |
|---|---|---|
| `/agents/config/` | `/agents/config/index.md` | `/agents/config.md` |
| `/agents/` (section index) | `/agents/index.md` | `/agents.md` |

## Testing

- `pytest scripts/test_flatten_md.py` — 11/11 passing.
- `mkdocs build --strict` — exits 0; hook renamed 217 Markdown files, rewrote links in 44 files.
- Verified: flat files exist, old `index.md` files gone, no stale `.../index.md` link targets remain, and buttons resolve to `../config.md` / `../agents.md`.

### DIFF
--- a/scripts/flatten_md.py
+++ b/scripts/flatten_md.py
@@ -69,6 +69,10 @@ def on_post_build(config, **kwargs) -> None:  # noqa: ARG001
         renamed += 1
 
     # Pass 2: rewrite site-internal .../index.md links in all text outputs.
+    # Load-bearing assumption: internal .md links are absolute {site_url}/... URLs
+    # because mkdocs-llmstxt normalizes them via its _convert_to_absolute_link, so
+    # rewrite_index_links only handles absolute links; relative .../index.md links
+    # are not expected here.
     rewritten = 0
     targets = list(site_dir.rglob("*.md"))
     for extra in ("llms.txt", "llms-full.txt"):
--- a/scripts/flatten_md.py
+++ b/scripts/flatten_md.py
@@ -7,8 +7,11 @@ Registered via the top-level ``hooks:`` list in ``mkdocs.yml``. Runs after the
 
 from __future__ import annotations
 
+import logging
 import re
-from pathlib import PurePosixPath
+from pathlib import Path, PurePosixPath
+
+import mkdocs.plugins
 
 
 def flat_md_target(rel_path: str) -> str | None:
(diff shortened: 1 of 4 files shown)


koverholt (commented):
This introduces a lot of complex machinery into our build, which risks broken builds in the future, or silently producing broken converted Markdown during the build.

The benefit seems small: `mkdocs-llmstxt` already emits working per-page `.md` links in `llms.txt` and `llms-full.txt` today, and an agent consuming those retrieves identical content whether the link ends in `page/index.md` or `page.md`. This only changes the link form, not what agents can retrieve, and then we have to add a link-rewriting pass over `llms.txt`/`llms-full.txt` specifically to keep those already-working links consistent after the rename.

One of the nice things about the current setup is that we depend entirely on the external plugin, with no custom build steps, config, or post-processing. This might make more sense as part of the plugin functionality itself and an option rather than maintaining as post-processing scripts.

I'll wait for @joefernandez to review as well since this is a big change to the build process, but in general I am -1 on this change since it gives us 200+ more lines to maintain and added risk to the docs build.

zeroasterisk:
Would this be an upstream change, for all mkdocs users

scripts/flatten_md.py
```diff
@@ -0,0 +1,93 @@
...
+    p = PurePosixPath(rel_path)
+    if p.name != "index.md":
+        return None
+    parent = p.parent
+    if parent in (PurePosixPath("."), PurePosixPath("")):
+        return None  # root site/index.md stays as-is
+    return f"{parent.as_posix()}.md"
+
+
+def rewrite_index_links(text: str, site_url: str) -> str:
```
joefernandez, line 36: rewriting the links of the markdown file is unnecessary. The links as provided work just fine, and you are  introducing complexity, dependency on this URL re-writing script, and creating potential link failures by doing this. Recommend removing this.

scripts/flatten_md.py
```diff
@@ -0,0 +1,93 @@
...
+def on_post_build(config, **kwargs) -> None:  # noqa: ARG001
+    """Rename generated ``index.md`` files to flat ``page.md`` and fix links.
+
+    Priority ``-100`` ensures this runs after the ``mkdocs-llmstxt`` plugin's
+    own ``on_post_build`` (lower priority runs later).
+    """
+    site_dir = Path(config["site_dir"])
+    site_url = config["site_url"]
+
+    # Pass 1: rename <dir>/index.md -> <dir>.md (root index.md is left alone).
```
joefernandez, line 61: You are restructuring the documentation file serving structure on the fly for the sole purpose of providing a new access method (**/page.md URL pattern). This solution not appropriate to the outcome and will likely create serving and maintenance issues in the future.

Since this functionality is purely meant to provide a new access pattern (**/page.md), and since we already provide a markdown output (e.g., https://adk.dev/get-started/index.md ), I recommend you create a extension of the redirect functionality that runs *after* all configured redirects in mkdoc.yaml to capture requests for the **/page.md URL pattern, such that a request for https://adk.dev/get-started.md is redirected to https://adk.dev/get-started/index.md , without rewriting the existing page source.

joefernandez (changes requested):

PR #1973, opened 2026-07-17
Updating skilltoolset page per issue #1350 - 4

zyantw:
- Added a note so the developer checks for the files in their working directory.
- Removed redundant and not useful information for the developer on the agent's original pr.
- Small formatting to headers.

Links
- Rendered page: https://deploy-preview-1973--adk-docs-preview.netlify.app/skills/
- Agent's PRs: https://github.com/google/adk-docs/pull/1357 and https://github.com/google/adk-docs/pull/1354
- Original issue: #1350


docs/skills/index.md
```diff
@@ -24,6 +25,10 @@ Use the `SkillToolset` class to make one or more Skills available to your agent.
 You can define [skills in code](#inline-skills) or load
 [skills from a filesystem](#filesystem-skills).
 
+!!! note "Check your working directory"
```
joefernandez: the assumption we should make with "Getting started" sections is that they cater to intermediate and expert-level users who will figure out little details like this on their own. So tips like this should included *after* the copy-paste code.

move this note to after the code example

docs/skills/index.md
```diff
@@ -24,6 +25,10 @@ Use the `SkillToolset` class to make one or more Skills available to your agent.
...
+
+        Ensure that 'skills/weather_skill' exist in your current working directory and contains the necessary files.
```
joefernandez: make this generic. /weather_skill is just an example, and the way this is presented makes it seem like this is important for all implementations. reword:

Ensure that 'skills/' directory exists in your current working directory and contains the sub-directories for the Skills you want to use in your agent.

docs/skills/index.md
```diff
@@ -123,6 +128,40 @@ three levels:
         documentation, templates, or examples.
     -   `scripts/`: Executable scripts supported by the agent runtime.
 
+### Skill processing and validation
```
joefernandez: 1. order: This information is not critical to building skills, so it should be moved to the end of the page, just ahead of Next Steps section.

2. heading hierarchy: bump the heading levels for this section UP one level, so this heading is at the same level as "Get started" and "Understand Skills"

docs/skills/index.md
joefernandez: Skill structure

(be more specific about the content of this section)

joefernandez (approved):
Thanks for update. Approved with comments.

PR #1976, opened 2026-07-18
fix: add missing await to create_session calls in callback examples

Syed-Umer-Ali:
InMemorySessionService.create_session is async def, but the before_agent_callback.py and after_agent_callback.py examples call it without await. This returns an unawaited coroutine — the session is never created, and subsequent runner.run_async() fails with SessionNotFoundError.
Fixed by adding await to both create_session calls in both files.


koverholt (approved):
Good catch, thanks for contributing this fix, which would cause a SessionNotFoundError. I verified the fix against the latest ADK Python SDK. LGTM, merging!

PR #1977, opened 2026-07-18
fix: correct undefined variable names in state output_key example

Syed-Umer-Ali:
Variables app_name and user_id were defined lowercase but referenced as APP_NAME and USER_ID (uppercase), causing NameError at runtime.## Summary

This PR fixes undefined variable references in the Python `output_key` example on the State documentation page.

The example defines `app_name` and `user_id` but later references `APP_NAME` and `USER_ID`, which are not defined and result in a `NameError` when the example is executed.

This change replaces the undefined uppercase variables with the correctly defined lowercase variables, making the example consistent and executable.


koverholt (approved):
Good catch, thanks for contributing this fix, which would cause a NameError. I verified the fix. LGTM, merging!

PR #1979, opened 2026-07-19
docs: fix type annotation error and missing imports in function tools snippet

Syed-Umer-Ali:
## Why

Two content bugs in the `human_in_the_loop.py` snippet used by the Function Tools docs page:

1. **Type error** — `reimburse()` annotated `-> str` but returns `{'status': 'ok'}` (a `dict`). Type checkers will flag this.

2. **Import error** — `from typing import Any` and `from google.adk.tools import LongRunningFunctionTool` are placed above the `[start:define_long_running_function]` snippet tag. The docs site only renders lines inside the tag, so the code block shown to readers is missing these imports. Copy-pasting the example causes `NameError`.

## Fix

- Changed `-> str` to `-> dict[str, Any]` on `reimburse()`
- Moved the two import lines inside the snippet region so they appear in the rendered code block

## Verification

- The example is syntactically valid Python 3.10+
- Total change: +4/−3 lines in one file. No behavioral impact.

### DIFF
--- a/examples/python/snippets/tools/function-tools/human_in_the_loop.py
+++ b/examples/python/snippets/tools/function-tools/human_in_the_loop.py
@@ -13,15 +13,16 @@
 # limitations under the License.
 
 import asyncio
-from typing import Any
 from google.adk.agents import Agent
 from google.adk.events import Event
 from google.adk.runners import Runner
-from google.adk.tools import LongRunningFunctionTool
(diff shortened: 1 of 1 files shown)


koverholt (approved):
Good catch, thanks for contributing this fix. I verified the fixes against the latest ADK Python SDK/release and everything checks out. LGTM, merging!

PR #1988, opened 2026-07-21
docs: remove Agent Config page

haranrk:
## Summary
- Removes the Agent Config documentation page (`docs/agents/config.md`), served at https://adk.dev/agents/config/
- Removes its navigation entry from `mkdocs.yml`

## Notes
The following files still contain links to `/agents/config/` and are **not** modified by this PR (they belong to other sections whose narrative depends on the Agent Config concept):
- `docs/visual-builder/index.md` (3 links)
- `tools/feature-matrix/start.md` (1 link)

Let me know if these references should also be cleaned up.


koverholt:
Is this related to the changes in google/adk-python@44cd1167 and google/adk-python@d7d73174? Those commits deprecate the Agent Config classes and the `from_config` loader, but the deprecation messages state that "config is now loaded via reflection so the separate config class is no longer needed", which seems like an internal refactor rather than a removal of the YAML feature. Users can still define an Agent Config YAML, and the adk web Visual Builder depends on it.

Given that, it seems only the [run-programmatically](https://adk.dev/agents/config/#run-programmatically) section is affected since it documents the deprecated `config_agent_utils.from_config`. Can you confirm whether the intent of those changes is to remove the entire Agent Config page, or just update the affected subsection?

PR #1994, opened 2026-07-22
Docs/fix template variable in instruction examples

Syed-Umer-Ali:
## What

Instruction strings across all 5 language examples in docs/agents/llm-agents.md
and the corresponding snippet files contain {country} as a documentation
placeholder inside the example query.

## Why

ADK's template engine treats {var} as a state variable injection token. The
docs themselves state (line 127-128): "If the state variable ... does not exist,
the agent will raise an error." Since country is never set as a state variable
in any of these examples, running the code as-written fails with a template
resolution error.

## Fix

Replace {country} with {{country}} in all instruction strings across Python,
TypeScript, Java (inline in docs/agents/llm-agents.md), Go
(examples/go/snippets/agents/llm-agents/snippets/main.go), and Kotlin
(examples/kotlin/snippets/agents/llm-agent/CapitalAgent.kt). Double-braces
escape ADK's template engine, rendering a literal {country} in the LLM
instruction — which is what the example intends.

## Verification

Confirmed that {var} is documented as state variable syntax on the same page
(lines 123-128). The fix preserves the displayed output while preventing template
engine errors. Python f-strings containing {country} in actual code (e.g., line
239) were left untouched.


Syed-Umer-Ali:
Hi maintainers 👋 Following up on this PR. It fixes the template resolution error ({country} → {{country}}) across the Python, TS, Java, Go, and Kotlin doc examples.

CLA is covered and CI checks have passed. Happy to update the branch or make any adjustments if needed. Thanks for your time!

joefernandez:
Thanks for the contribution. We actually want these examples to raise errors for undefined variables, rather than failing quietly because they are escaped.

Closing as working as intended.

PR #2000, opened 2026-07-24
docs: add documentation for OpenAI model support in ADK agents

hanorik:
Added documentation for the experimental OpenAI model support in ADK Go.


karolpiotrowicz:
A few suggestions to round this out:

**1. Link the runnable example.** `examples/openai/` is a great resource and linking repo examples is the norm on other pages. Suggest adding to the end of "Get started":
```markdown
For a complete, runnable sample, see [examples/openai/](https://github.com/google/adk-go/tree/main/examples/openai) in the ADK Go repository.
```

**2. Add a short "Supported features" and "Limitations" section.** For an experimental, text-focused integration, it helps to set expectations up front:
```markdown
## Supported features

- Text generation (streaming and non-streaming)
- Function (tool) calling
- Structured output via `OutputSchema` (JSON schema)
- Reasoning models (e.g. o-series), including reasoning-token accounting
- Token logprobs

## Limitations

- **Text only** — multimodal input (images, audio, files) is not supported.
- **Function tools only** — built-in tools (Google Search, code execution, etc.) are not supported.
- **Structured output uses OpenAI strict mode** — every field declared in an `OutputSchema` is treated as required.
- Some `GenerateContentConfig` options return an error rather than being silently ignored: `TopK`, stop sequences, multiple candidates, frequency/presence penalties, request labels, and safety settings.
```

**3. Admonition type.** `!!! example "Experimental"` renders as a green "Example" box; a caveat reads better (and stands out more) as a warning:
```markdown
!!! warning "Experimental"
```

**4. Minor polish.**
- Heading case: "Configuration Options" → "Configuration options" (the other headings use sentence case).
- The snippet uses `ctx` and `log` without showing where they come from — worth a one-line note (e.g. "assuming a `context.Context` named `ctx`") or including the `context`/`log` imports so it's copy-paste-ready.
- There's a stray tab on the blank line after the model instantiation.

With the example link and the capabilities/limitations added, this will be a really solid page. Thanks again!

karolpiotrowicz:
One more suggestion, on how this page relates to LiteLLM. Since it sits right next to the LiteLLM page — and ADK Python reaches OpenAI *through* LiteLLM — a reader landing here could read "OpenAI = experimental, Go-only" and miss that OpenAI is fully usable from Python today via LiteLLM.

The [Claude page](/agents/models/anthropic/) already handles this nicely by making the intro language-aware. Suggest doing the same here — replace the single-line intro with:

```markdown
You can use OpenAI models (such as GPT-4o and GPT-4o Mini) with ADK. How you connect depends on the language:

- **Go — native support (this page):** ADK Go provides a direct `openaimodel` package that implements the `model.LLM` interface, targeting the OpenAI Responses API. It is currently experimental.
- **Python — via LiteLLM:** ADK Python accesses OpenAI models (and many other providers) through the LiteLLM connector. See [LiteLLM](/agents/models/litellm/).

The rest of this page covers the Go-native integration.
```

This keeps the page focused on the Go-native model while making it clear, up front, that Python users have a stable path via LiteLLM — so no one mistakes "experimental, Go" for "OpenAI isn't well supported."

(Optional, separate follow-up: a one-liner on the LiteLLM page's OpenAI mention pointing Go users back to this native page, for symmetry.)

hanorik:
@karolpiotrowicz thanks for the comments! I've added suggested sections and text adjustments except the `Example` box. Actually it renders as a purple box and seems informative enough for the purpose. This box is also used in the `Model routing` article so to keep it more consistent I'd prefer to leave it as it is.

docs/agents/models/openai.md
```diff
@@ -0,0 +1,100 @@
...
+<div class="language-support-tag">
+   <span class="lst-supported">Supported in ADK</span><span class="lst-go">Go v2.1.0</span><span class="lst-preview">Experimental</span>
+</div>
+
+!!! example "Experimental"
+
+    The `openaimodel` package is experimental and its behavior may change or be removed in the future. We welcome your
+		[feedback](https://github.com/google/adk-go/issues/new?template=feature_request.md)!
+
+You can use OpenAI models (such as GPT-4o and GPT-4o Mini) with ADK. How you connect depends on the language:
```
joefernandez: remove "(such as GPT-4o and GPT-4o Mini)"

since the version numbers will date this document

docs/agents/models/openai.md
```diff
@@ -0,0 +1,100 @@
...
+
+- **Go — native support (this page):** ADK Go provides a direct `openaimodel` package that implements the `model.LLM` interface, targeting the OpenAI Responses API. It is currently experimental.
```
joefernandez: 1. remove: "(this page)"

generally avoid self-referential language in docs.

2. remove: "It is currently experimental."

state this once at the beginning and don't mention it again, so that taking it out of experimental status is a trivial update

3. Add: `[Get started](#get_started).` as last sentence

docs/agents/models/openai.md
```diff
@@ -0,0 +1,100 @@
...
+- **Python — via LiteLLM:** ADK Python accesses OpenAI models (and many other providers) through the LiteLLM connector. See [LiteLLM](/agents/models/litellm/).
+
+The rest of this page covers the Go-native integration.
```
joefernandez: remove. This is obvious from the language tags, and just creates something we have to update later when language coverage increases.

docs/agents/models/openai.md
hanorik: Done.

docs/agents/models/openai.md
hanorik: Done.

docs/agents/models/openai.md
hanorik: Done.

joefernandez (approved):

PR #2001, opened 2026-07-24
docs: clarify TypeScript API key environment variable differs from other quickstarts

MuhammadHamidRaza:
## Summary

The TypeScript quickstart currently instructs users to configure `GEMINI_API_KEY`, while the Python, Go, Java, and Kotlin quickstarts all use `GOOGLE_API_KEY`.

This difference is currently undocumented, which can confuse users moving between ADK languages.

## Problem

The TypeScript ADK's `GeminiParams.apiKey` resolves its API key from:

- `GOOGLE_GENAI_API_KEY`
- `GEMINI_API_KEY`

It does **not** read `GOOGLE_API_KEY`.

Meanwhile:

- Python examples use `GOOGLE_API_KEY`
- Go examples use `GOOGLE_API_KEY`
- Java examples use `GOOGLE_API_KEY`
- Kotlin examples use `GOOGLE_API_KEY`

A user who follows one of those quickstarts first and later tries the TypeScript quickstart may reasonably expect the same environment variable to work. Instead, authentication fails because the TypeScript ADK never reads `GOOGLE_API_KEY`, and the documentation doesn't explain the difference.

## What this PR changes

This PR is documentation-only.

- Adds a note to the TypeScript quickstart explaining that the TypeScript ADK currently reads `GEMINI_API_KEY` or `GOOGLE_GENAI_API_KEY`, not `GOOGLE_API_KEY`.
- Adds a cross-reference note to the other language quickstarts to alert users switching to TypeScript.

## Verification

Verified against:

- All five ADK quickstart guides in `google/adk-docs`
- TypeScript `GeminiParams.apiKey` API reference
- Python `google.genai.Client`
- Go SDK (`os.Getenv("GOOGLE_API_KEY")`)
- Kotlin SDK (`System.getenv("GOOGLE_API_KEY")`)

## Notes

A longer-term improvement would be for the TypeScript ADK to also support `GOOGLE_API_KEY`, bringing it in line with the other ADK language SDKs. That behavior change is outside the scope of this documentation-only PR.

### DIFF
--- a/docs/get-started/typescript.md
+++ b/docs/get-started/typescript.md
@@ -92,19 +92,19 @@ to set environment variables:
 === "MacOS / Linux"
 
     ```bash title="Update: my-agent/.env"
-    echo 'GEMINI_API_KEY="YOUR_API_KEY"' > .env
+    echo 'GOOGLE_API_KEY="YOUR_API_KEY"' > .env
     ```
 
 === "Windows PowerShell"
 
     ```console title="Update: my-agent/.env"
-    echo 'GEMINI_API_KEY="YOUR_API_KEY"' > .env
+    echo 'GOOGLE_API_KEY="YOUR_API_KEY"' > .env
     ```
 
 === "Windows Command Prompt"
 
     ```console title="Update: my-agent/.env"
-    echo GEMINI_API_KEY="YOUR_API_KEY" > .env
+    echo GOOGLE_API_KEY="YOUR_API_KEY" > .env
     ```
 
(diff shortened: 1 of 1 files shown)


koverholt:
This PR would break the ADK JS quickstart. The PR body states that "[adk-js] does not read `GOOGLE_API_KEY`", but then swaps `GEMINI_API_KEY` for `GOOGLE_API_KEY`, which won't work with adk-js. Closing since the ADK JS quickstart is correct as is, and no need to add further notes.

PR #1772, opened 2026-05-19
ADK Python 2.0.0 GA docs update

joefernandez:

PR #2009, opened 2026-07-27
Stop publishing homepage partials as standalone pages

haranrk:
## Problem

The homepage is assembled from partials in `docs/_includes/homepage/` via the macros plugin (`docs/index.md`). `mkdocs.yml` listed `/_includes/**` under **`not_in_nav`**, which only suppresses the "page exists but is not in nav" warning — it does not exclude the file from the build.

So all ten partials are built, published, and indexed. They currently have live URLs:

- https://adk.dev/_includes/homepage/_eval/
- https://adk.dev/_includes/homepage/_hero/
- https://adk.dev/_includes/homepage/_faq/
- …and seven more

All ten are in `sitemap.xml` and the site search index.

These pages are broken in isolation, because the partials use paths relative to the site root. On `/_includes/homepage/_eval/`, for example:

| markup in partial | resolves to | result |
| --- | --- | --- |
| `href="evaluate/"` | `/_includes/homepage/_eval/evaluate/` | 404 |
| `src="assets/adk-eval-case.gif"` | `/_includes/homepage/_eval/assets/…` | broken image |

`_eval.md` is the most visible case since nothing links to it at all — it was dropped from `index.md` in #1772 but the file was left behind. `_ai-dev-tools.md` was dropped in the same PR and is likewise unreferenced.

## Changes

1. Move `/_includes/**` from `not_in_nav` to `exclude_docs` so the partials are skipped by the build. The macros plugin reads them from disk through its own Jinja loader, so the homepage is unaffected.
2. Delete `_eval.md` and `_ai-dev-tools.md`, which are no longer included anywhere.

## Verification

Local `mkdocs build` before and after:

- `site/index.html` is **byte-identical** — the homepage renders exactly as before.
- The only difference across the entire output tree is the removal of the ten `_includes/` pages (3284 → 3274 HTML files).
- `_includes` entries in `sitemap.xml`: 10 → 0.
- `_includes` entries in the search index: 10 → 0.
- No new build warnings or errors.


joefernandez (approved):
Thanks for the cleaning up this serving messiness.

PR #1956, opened 2026-07-12
Add Unstructured Transform MCP integration

simoncoombes:
Adds an integration catalog page for the [Unstructured Transform MCP server](https://docs.unstructured.io/transform/overview) — Unstructured's hosted document-parsing service (PDFs, Office documents, images, and 60+ file types into structured, AI-ready output).

## What's included

- `docs/integrations/unstructured.md` — catalog frontmatter, use cases, prerequisites, a complete working agent example (remote streamable-HTTP connection with bearer auth), the 4-tool reference table, and resources.
- `docs/integrations/assets/unstructured.png` — 400×400 logo.

## Verification

The example was verified live against ADK 2.4.0 (Python) and the production server on 2026-07-12: `McpToolset` + `StreamableHTTPConnectionParams` with the `Authorization: Bearer` header authenticates at the `initialize` handshake, all 4 tools enumerate under `tool_filter`, and a Gemini-driven agent completed the full flow autonomously (start job → paced status polling → fetch results) against a real PDF.

The page documents two practical findings from that verification: the `[mcp]` extra is required for ADK's MCP classes to import, and the default 5s connection timeout is too short for a remote handshake (the example sets 30s).

## Note on draft status

Opened as draft while the companion installation guide on the Unstructured docs site (linked from the page) finishes publishing — will mark ready once that link is live. Authored and maintained by Unstructured (I work there); we'll keep the page current as ADK and the server evolve.

PR #2010, opened 2026-07-28
docs: update Unstructured Transform MCP tool names and cover the extraction tools

simoncoombes:
The Unstructured Transform MCP server renamed three of its tools after the integration page landed (#1956), so the sample agent on that page no longer works. The `tool_filter` list is the breaking part: it names tools the server no longer exposes, so the toolset resolves to nothing and the agent gets no Transform tools at all.

- `transform_files` is now `start_transform_job`
- `check_transform_status` is now `check_job_status`
- `get_transform_results` is now `get_job_results`

`request_file_upload_url` is unchanged, and the tool behaviour and arguments are the same.

The same release added structured data extraction, which the page did not cover, so I have included it here rather than sending a second PR over the same file. Two new tools, `suggest_extraction_schema_for_file` and `start_extraction_job`, return named fields as JSON matching a schema. Because `tool_filter` is an allowlist, they have to be listed there or an agent cannot reach them at all, which is the same failure mode as the stale names above.

The one thing worth knowing when reviewing: extraction runs on the element JSON that a parse produces, not on a raw file, so it always follows a transform job and is driven by the `output_ref` that `get_job_results` returns for each file. The agent instruction, the note, and the tool table now say that. Extraction results also come back inline rather than behind a pre-signed download URL.

I verified both the names and the extraction flow against the live server (Unstructured Transform 0.7.2) rather than working from a changelog, including running a parse and then an extraction end to end. They match the rename on the Unstructured side (Unstructured-IO/docs#1004). I maintain the linked Unstructured guide, so if anything else on the page needs updating later I am happy to follow up.

### DIFF
--- a/docs/integrations/unstructured.md
+++ b/docs/integrations/unstructured.md
@@ -96,11 +96,11 @@ the agent pause between status checks, because parsing jobs run asynchronously:
             name="transform_agent",
             instruction=(
                 "You parse documents with the Unstructured Transform MCP server. "
-                "Pass public https:// file URLs straight to transform_files. It "
-                "returns a job_id; poll with check_transform_status, calling "
+                "Pass public https:// file URLs straight to start_transform_job. It "
+                "returns a job_id; poll with check_job_status, calling "
                 "wait_seconds(30) between checks (jobs take 30 seconds to a few "
-                "minutes). When the job completes, call get_transform_results and "
-                "report the parsed content back to the user. transform_files "
+                "minutes). When the job completes, call get_job_results and "
+                "report the parsed content back to the user. start_transform_job "
                 "accepts an optional stages config; it auto-selects a parse "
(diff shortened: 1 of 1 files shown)


simoncoombes:
I pushed a second commit that adds the server's structured data extraction tools (`suggest_extraction_schema_for_file` and `start_extraction_job`) to the `tool_filter`, the agent instruction, and the tool table, and updated the PR description. They shipped in the same server release as the rename, and since `tool_filter` is an allowlist they would otherwise be unreachable from this page's agent, so it seemed better than a second PR touching the same file. Happy to split it back out if you would rather review the rename on its own.

koverholt (approved):
Thanks for the PR and the clean updates to the Unstructured integration page! All content LGTM, merging.

PR #2011, opened 2026-07-28
docs(deploy): correct adk deploy CLI options for Cloud Run and GKE

GWeale:
Small slice of the docs audit covering the deploy pages and the Python get-started page.

## Changes

- **`docs/deploy/cloud-run.md`** — documented an `--agent_engine_id` option that `adk deploy cloud_run` does not accept; it now documents the real `--session_service_uri`, `--artifact_service_uri` and `--memory_service_uri` options with their accepted URI forms, plus a note that leaving them unset falls back to in-memory services and loses sessions and artifacts when an instance is recycled.
- **`docs/deploy/gke.md`** — the `--log_level` row omitted `critical` from the accepted values; added.
- **`docs/get-started/python.md`** — the project layout omitted the `.gitignore` that keeps `.env` out of source control; added to the tree.

## How this was produced


docs/get-started/python.md
```diff
@@ -59,6 +59,7 @@ file containing the main control code for the agent.
 my_agent/
     agent.py      # main agent code
     .env          # API keys or project IDs
+    .gitignore    # keeps .env out of source control
```
joefernandez: remove. This assumes all readers are using Git for source control

joefernandez:
### Technical Verification & Code Citations:

1. Removal of --agent_engine_id and Addition of Service URIs:

In cli_tools_click.py:L2137-L2158 ([GitHub Link](https://github.com/google/adk-python/blob/main/src/google/adk/cli/cli_tools_click.py#L2137-L2158)), the cli_deploy_cloud_run command definition does not accept agent_engine_id as an option. Instead, it is decorated with adk_services_options ([GitHub Link](https://github.com/google/adk-python/blob/main/src/google/adk/cli/cli_tools_click.py#L558-L644)), which adds the --session_service_uri, --artifact_service_uri, and --memory_service_uri options.

2. In-Memory Fallback Behavior:

In cli_deploy.py:L645-L668 ([GitHub Link](https://github.com/google/adk-python/blob/main/src/google/adk/cli/cli_deploy.py#L645-L668)), the to_cloud_run deployment logic receives these options. As shown in cli_deploy.py:L706-L708 ([GitHub Link](https://github.com/google/adk-python/blob/main/src/google/adk/cli/cli_deploy.py#L706-L708)), if use_local_storage is False (which is the default for Cloud Run) and the ADK version is >= 1.3.0, both session_service_uri and artifact_service_uri default to 'memory://' if not explicitly set:

```python
if parse(adk_version) >= parse('1.3.0') and not use_local_storage:
  session_service_uri = session_service_uri or 'memory://'
  artifact_service_uri = artifact_service_uri or 'memory://'
```

This matches the warning added in the documentation.

3. GKE Deployment Documentation (docs/deploy/gke.md)
Proposed Changes: Update the --log_level options to include critical.

Verification & Code Citations:
In cli_tools_click.py:L2657-L2677 ([GitHub Link](https://github.com/google/adk-python/blob/main/src/google/adk/cli/cli_tools_click.py#L2657-L2677)), the cli_deploy_gke command defines --log_level with the type LOG_LEVELS. The LOG_LEVELS ([GitHub Link](https://github.com/google/adk-python/blob/main/src/google/adk/cli/cli_tools_click.py#L50-L53)) choice object is defined as:

```python
LOG_LEVELS = click.Choice(
    ["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
    case_sensitive=False,
)
```

Thus, critical (case-insensitive) is indeed a valid option, making this documentation change correct.

joefernandez (approved):
Thanks for update. Verified and approved.

PR #2017, opened 2026-07-28
docs(sessions): fix async examples, imports and state claims

GWeale:
Corrects the sessions, state, memory, artifacts and events pages, where the Python examples and behavioral claims had drifted from the v2.5.0 source.

## Changes
- **`docs/artifacts/index.md`** — the `list_user_files_py` example `await`ed `tool_context.list_artifacts()` from a plain `def`; it is now `async def`. Two places said Python has separate `CallbackContext`/`ToolContext` types — in Python these are aliases of the unified `Context`. Also documents that the list/get artifact-version methods return `ArtifactVersion` metadata (not the payload), that in-memory versions hold that metadata alongside the `types.Part`, and that listing from a session also returns the user's `user:`-prefixed filenames.
- **`docs/sessions/state.md`** — `from google.adk.agents import CallbackContext` does not resolve; now imported from `google.adk.agents.callback_context`. Dropped the blanket "ensures thread-safety" claim: `DatabaseSessionService` takes a per-session lock, `InMemorySessionService` is not safe for multi-threaded use. Callback names `on_before_agent_call` / `on_after_agent_call` corrected to `before_agent_callback` / `after_agent_callback`, and `get_user_state()` documented for reading `user:` state before a session exists.
- **`docs/events/index.md`** — the `is_final_response()` rules were wrong: `skip_summarization` alone is enough in Python (no tool result required), and the two examples labelled `is_final_response() == False` (state/artifact-update-only, loop escalation) are in fact `True`. The Python accessor is the `event.content` attribute, not `event.content()`.
- **`docs/sessions/session/index.md`** — install hints `pip install vertexai` and `google-adk[vertexai]` corrected to `google-adk[gcp]`; `DatabaseSessionService` now notes it needs the `db` extra; a snippet that reassigned `session_service` to the result of `create_session` now assigns `session`.
- **`docs/sessions/memory.md`** — `VertexAiRagMemoryService` is not conditionally exported; it always imports, but constructing it raises `ImportError` without the Agent Platform SDK.

## How this was produced


joefernandez:
## Technical review report

## Verdict: Approve — all 15 technical claims verified correct

I verified each claim two ways: reading `adk-python` main ([`2353dde`](https://github.com/google/adk-python/commit/2353dde8e9081439d8d9f8358469ddf44ce997fc)) and *executing* the behavior against a clean `google-adk==2.6.3` install.

**Highest-value fixes:**

- **`state.md:571` is a hard crash.** `from google.adk.agents import CallbackContext` raises `ImportError` — `CallbackContext` is in neither `__all__` nor `_LAZY_MEMBERS` ([`agents/__init__.py#L41-L80`](https://github.com/google/adk-python/blob/2353dde8e9081439d8d9f8358469ddf44ce997fc/src/google/adk/agents/__init__.py#L41-L80)). It's an alias assignment in `callback_context.py` ([L21](https://github.com/google/adk-python/blob/2353dde8e9081439d8d9f8358469ddf44ce997fc/src/google/adk/agents/callback_context.py#L21)).
- **`artifacts/index.md:936` is a `SyntaxError`** — `await` in a plain `def`; [`list_artifacts` is a coroutine](https://github.com/google/adk-python/blob/2353dde8e9081439d8d9f8358469ddf44ce997fc/src/google/adk/agents/context.py#L768-L776).
- **`is_final_response()` docs were wrong in 4 places.** The helper is 9 lines ([`event.py#L288-L304`](https://github.com/google/adk-python/blob/2353dde8e9081439d8d9f8358469ddf44ce997fc/src/google/adk/events/event.py#L288-L304)) and short-circuits on `skip_summarization` *or* `long_running_tool_ids` alone. I built the actual truth table — state-delta-only and loop-escalation events return `True`, not `False`. Notably the author did **not** blanket-flip labels: the transfer-signal example genuinely is `False` and was correctly left alone.
- **`google-adk[vertexai]` is not a real extra.** Confirmed against the distribution's declared extras; `gcp` and `db` are, and [`missing_extra`](https://github.com/google/adk-python/blob/2353dde8e9081439d8d9f8358469ddf44ce997fc/src/google/adk/utils/_dependency.py#L20-L30) emits `pip install google-adk[gcp]` verbatim.

**Three things to flag to the author:**

1. The PR body says it audited against **v2.5.0**; current is **v2.6.3**. I re-verified — everything still holds, so no rework, but the description should be bumped.
2. `google-adk[vertexai]` also survives at [`express-mode.md:91`](https://github.com/google/adk-docs/blob/d9e930e823ac88103c8dfee0ad754779aaace773/docs/integrations/express-mode.md#L91) and `:120` — out of scope, worth a follow-up.
3. Rule 2 in `events/index.md` still says "check if the `longRunningToolIds` list is **empty**" while the code sample below it checks non-empty. Pre-existing, but the PR rewrites that sentence anyway.

One incidental find: [`event.py#L152`](https://github.com/google/adk-python/blob/2353dde8e9081439d8d9f8358469ddf44ce997fc/src/google/adk/events/event.py#L152) says *"Do not assign the ID. It will be assigned by the session"* — that comment is stale **in adk-python**, contradicted by `model_post_init` 12 lines below. The PR documents actual behavior correctly; upstream is the thing that's wrong.

The diff also applies cleanly to current `adk-docs` main — no rebase needed.

joefernandez (approved):

PR #2022, opened 2026-07-28
docs(graphs): fix graph workflow samples and behavior claims

GWeale:
Corrections to the `docs/graphs/` pages (plus one note on `docs/workflows/patterns.md`) where Python samples could not run as written and documented behavior did not match v2.5.0.

## Changes
- **`docs/graphs/dynamic.md`** — a node that `yield`s is an async generator, so the sample's bare `return code` emitted nothing; it now yields `Event(output=code)`. Also `FunctionNode(func=...)` as a keyword, `@node(rerun_on_resume=True)` on workflow nodes, and the parallel sample no longer takes a `real_node` argument no caller could supply — it defines the worker node and passes `use_sub_branch=True`.
- **`docs/graphs/index.md`** — live streaming was listed as flatly unsupported. It is the default `single_turn` node mode that ignores the live request queue; nodes in `task` or `chat` mode can stream.
- **`docs/graphs/data-handling.md`** — dropped the `{CityTime.time_info}` / `<field from node>` data-selection syntax, which does not exist; `instruction` placeholders are substituted from session state, so an unmatched one is left in the prompt verbatim. Also corrected the multi-`yield` description (each `yield` emits a separate Event and only one may carry `output`) and the state-print / node-input parameter names.
- **`docs/graphs/routes.md`** — `JoinNode` waits on upstream *completion*, not output (completing with no output gives `None`; a node that never runs stalls the join). A workflow may have at most one output-producing terminal node, else `ValueError`. `chat` mode agents are only allowed directly after `START`, and a cycle needs at least one routed edge; the loop sample now actually loops.
- **`docs/graphs/human-input.md`** — replaced the `RequestInput` sample with the tool-confirmation API (`FunctionTool(..., require_confirmation=True)` and `tool_context.request_confirmation()`), and linked the full confirmation page.
- **`docs/workflows/patterns.md`** — added a note that `SequentialAgent`, `ParallelAgent` and `LoopAgent` are deprecated in Python and emit a `DeprecationWarning`.

## How this was produced

### DIFF
--- a/docs/graphs/data-handling.md
+++ b/docs/graphs/data-handling.md
@@ -369,6 +369,8 @@ accepted and produced by any agent node.
     ***BaseModel*** to constrain any agent's input and output:
 
     ```python
+    from datetime import date
+
     from google.adk import Agent
     from pydantic import BaseModel
 
@@ -389,14 +391,14 @@ accepted and produced by any agent node.
         output_schema=FlightSearchOutput,
         tools=[search_flights_api],
         mode="single_turn",
-        ...
+        # ...
     )
 
     assistant = Agent(
         name="assistant",
         instruction="You help users plan trips.",
         sub_agents=[flight_searcher],
-        ...
+        # ...
     )
     ```
 
@@ -466,7 +468,7 @@ accepted and produced by any agent node.
     root_agent = Workflow(
         name="root_agent",
         edges=[
-            (START, city_generator_agent, lookup_time_function, city_report_agent)
+            ("START", city_generator_agent, lookup_time_function, city_report_agent)
         ],
     )
     ```
@@ -500,9 +502,9 @@ accepted and produced by any agent node.
     framework serializes it into `Event.Output`. The successor `AgentNode`
     receives the struct as its user content — the fields are available to the
     agent's `Instruction` without any `{key}` template syntax. This is the
-    direct equivalent of Python's `input_schema=CityTime` with
-    `{CityTime.time_info}` template placeholders: the struct fields are
-    delivered as typed input rather than looked up by name from state.
+    direct equivalent of Python's `input_schema=CityTime`: in both languages
+    the struct fields are delivered as typed input rather than looked up by
+    name from state.
 
     ```go
     --8<-- "examples/go/snippets/graphs/data-handling/main.go:structured-output"
--- a/docs/graphs/data-handling.md
+++ b/docs/graphs/data-handling.md
@@ -365,8 +365,8 @@ accepted and produced by any agent node.
     The curly-brace `{key}` placeholders that an `instruction` supports are
     substituted from *session state*, not from the node input, so a
     placeholder such as `{CityTime.time_info}` is left in the prompt verbatim.
-    To template a value into an instruction, write it to state first — with
-    `output_key` or `Event(state=...)` — and reference it as `{time_info}`.
+    To template a value into an instruction, write it to state first, with
+    `output_key` or `Event(state=...)`, and reference it as `{time_info}`.
 
 === "Go"
 
--- a/docs/graphs/data-handling.md
+++ b/docs/graphs/data-handling.md
@@ -84,8 +84,9 @@ Each step in a workflow produces output for its successor.
     Use the ***return*** syntax when outputting ***Event*** data that does not
     require additional processing. When emitting data that requires additional
     processing, or if you are generating more than one data item, you can use
-    more than one ***yield*** command. Each ***yield*** call adds to a list of
-    data objects on the Event which is passed to the next node of a graph. A
+    more than one ***yield*** command. Each ***yield*** call emits a separate
+    ***Event***, but only one of those events may carry an `output`, and that
+    single value is what the next node of the graph receives. A
     ***return*** or ***yield*** command without a parameter passes a `None` value
     to the next node.
 
@@ -216,7 +217,7 @@ inside tools and callbacks regardless of which agent style you use.
           },
       )
 
-    async def task_attempt_node(node_input: Content, attempts: int):
+    async def task_attempt_node(node_input, attempts: int):
       yield Event(
           state={
               "attempts": attempts + 1,
@@ -224,7 +225,7 @@ inside tools and callbacks regardless of which agent style you use.
       )
 
     async def read_state_node(ctx: Context):
-      print(f"attempts state: {ctx.state}") # attempts state: attempts: 1
+      print(f"attempts state: {ctx.state['attempts']}") # attempts state: 1
 
     root_agent = Workflow(
         name="root_agent",
@@ -328,35 +329,26 @@ accepted and produced by any agent node.
 
 === "Python"
 
-    Use the curly-brace `{ }` syntax to select properties from the input
-    schema, or `< >` to select a property and also qualify it by the name
-    of the source node:
+    Use `input_schema` on an agent to validate the structured output of the
+    preceding node and deliver it as the agent's input. Refer to the fields of
+    that input by name in the instruction:
 
     ```python
     class CityTime(BaseModel):
         time_info: str  # time information
         city: str       # city name
 
-    def lookup_time_function(city: str):
+    def lookup_time_function(node_input: str):
         """Simulate returning the current time in the specified city."""
-        return Event(output=CityTime(time_info='10:10 AM', city=city))
+        return Event(output=CityTime(time_info='10:10 AM', city=node_input))
 
     city_report_agent = Agent(
         name="city_report_agent",
         model="gemini-flash-latest",
         input_schema=CityTime,
-
-        # data selection based on class and parameter
-        # instruction="""
-        #     Return a sentence in the following format:
-        #     It is {CityTime.time_info} in {CityTime.city} right now.
-        # """,
-
-        # more restrictive data selection based on source node name
         instruction="""
-            Return a sentence in the following format:
-            It is <CityTime.time_info from lookup_time_function> in
-            <CityTime.city from lookup_time_function> right now.
+            You receive a CityTime with time_info and city fields. Reply with a
+            single sentence stating that it is that time in that city right now.
         """,
     )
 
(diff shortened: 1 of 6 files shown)


docs/workflows/patterns.md
```diff
@@ -793,7 +802,7 @@ your project requirements before committing to a full implementation.
     * **Interaction:** Can be implemented using a custom **Tool** that pauses execution and sends a request to an external system (e.g., a UI, ticketing system) waiting for human input. The tool then returns the human's response to the agent.
     * **Workflow:** Could use **LLM-Driven Delegation** (`transfer_to_agent`) targeting a conceptual "Human Agent" that triggers the external workflow, or use the custom tool within an `LlmAgent`.
     * **State/Callbacks:** State can hold task details for the human; callbacks can manage the interaction flow.
-    * **Note:** ADK doesn't have a built-in "Human Agent" type, so this requires custom integration.
+    * **Note:** ADK has no "Human Agent" agent type, but in Python you do not have to build the pause yourself: yield a `RequestInput` event from a [graph workflow node](/graphs/human-input/), or set `require_confirmation=True` on a `FunctionTool` for a yes/no approval. To supply a custom hint, call `tool_context.request_confirmation()` inside the tool.
```
joefernandez, line 805: L0 RISK REVIEW (Low): This bullet is language-neutral, but the new note scopes the guidance to Python. Tool confirmation is documented for Go and Java as well ([`confirmation.md` L87-L119](https://github.com/google/adk-docs/blob/f5bd694f49d0fb40333cc6ac8ef8e58a1aacdec4/docs/tools-custom/confirmation.md#L87-L119)), and graph human-input nodes are marked supported in Python, TypeScript, and Go ([`human-input.md` L3-L5](https://github.com/google/adk-docs/blob/f5bd694f49d0fb40333cc6ac8ef8e58a1aacdec4/docs/graphs/human-input.md#L3-L5)). Suggest dropping the Python framing and letting the linked pages carry per-language support, which also avoids naming Python-only syntax here.

```suggestion
    * **Note:** ADK has no *Human Agent* agent type, but you do not have to build the pause yourself: yield a `RequestInput` event from a [graph workflow node](/graphs/human-input/), or enable [tool confirmation](/tools-custom/confirmation/#boolean-confirmation) on a `FunctionTool` for a yes/no approval. To supply a custom hint or collect structured data, use [advanced confirmation](/tools-custom/confirmation/#advanced-confirmation) inside the tool. See those pages for per-language support.
```

joefernandez (commented):
# Content review report

- **Review Score**: 99
- **Summary reasoning**: The change corrects workflow-graph samples and tightens behavior claims in existing pages. No legal, security, privacy, pricing, dated, third-party, customer-name, or unreleased-feature content is added. One capability statement is scoped too narrowly and understates non-Python language support.
- **Recommended action**: Approve with one change — drop the Python-only scoping on the human-in-the-loop note in [`docs/workflows/patterns.md` L805](https://github.com/google/adk-docs/blob/f5bd694f49d0fb40333cc6ac8ef8e58a1aacdec4/docs/workflows/patterns.md#L805). The only risk factor is an accuracy/capability-representation nit in an otherwise language-neutral bullet; everything else in the change reduces risk by removing unsupported claims.
- **Number of lines changed**: M, 249
- **Number of files changed**: 8

## 🟢 Low risk items

  - **Capability claim scoped to Python only.** [`docs/workflows/patterns.md` L805](https://github.com/google/adk-docs/blob/f5bd694f49d0fb40333cc6ac8ef8e58a1aacdec4/docs/workflows/patterns.md#L805) says "in Python you do not have to build the pause yourself", but tool confirmation is documented for Go and Java too (`RequireConfirmation` / `requireConfirmation`, [`docs/tools-custom/confirmation.md` L87-L119](https://github.com/google/adk-docs/blob/f5bd694f49d0fb40333cc6ac8ef8e58a1aacdec4/docs/tools-custom/confirmation.md#L87-L119)), and graph human-input nodes are listed as supported in Python, TypeScript, and Go ([`docs/graphs/human-input.md` L3-L5](https://github.com/google/adk-docs/blob/f5bd694f49d0fb40333cc6ac8ef8e58a1aacdec4/docs/graphs/human-input.md#L3-L5)). Suggested rewrite is in the inline comment.

joefernandez (commented):
Checked the three flagged items against the released ADK sources; all three hold. Fixes for each are suggested inline.

- `docs/workflows/patterns.md` — the deprecation note is scoped to Python, but adk-js deprecates the same three classes ([`main-v2.0.0`](https://github.com/google/adk-js/blob/main-v2.0.0/core/src/agents/sequential_agent.ts#L42-L50), still present in [`main-v2.1.0`](https://github.com/google/adk-js/blob/main-v2.1.0/core/src/agents/sequential_agent.ts#L47-L50)). I confirmed Go [v2.4.0](https://github.com/google/adk-go/blob/v2.4.0/agent/workflowagents/sequentialagent/agent.go) and Java [v1.10.1](https://github.com/google/adk-java/blob/v1.10.1/core/src/main/java/com/google/adk/agents/SequentialAgent.java) carry no deprecation, so "Python and TypeScript" is the correct scope.
- `docs/graphs/dynamic.md` — the `return code` → `yield Event(output=code)` fix is correct, but `compile_lint_check` in the same snippet still has the state-binding defect this PR fixes elsewhere.
- `docs/graphs/data-handling.md` — `from datetime import date` makes `departure_date` resolve, but `Flight` and `search_flights_api` are still undefined, so the snippet still raises `NameError`.

Everything else in the PR looked accurate on the points I checked, including the single-terminal-output `ValueError` claim in `routes.md` and the `chat`-mode-after-`START` restriction.

One thing I did not change, outside these three: `code_workflow(ctx: Context, user_request: str)` (line 435 here, and `editorial_workflow` at line 280) has the same shape as the `compile_lint_check` defect — a non-`node_input` parameter under the default `state` binding. Worth confirming whether an entry node is special-cased before treating it as a bug.

docs/workflows/patterns.md
```diff
@@ -9,6 +9,15 @@ Agent Development Kit (ADK), including code examples. These patterns are useful
 across a broad set of applications and you should evaluate and test them against
 your project requirements before committing to a full implementation.
 
+!!! note "Python: template workflow agents are deprecated"
```
joefernandez, line 12: `SequentialAgent`, `ParallelAgent` and `LoopAgent` are deprecated in adk-js too, not just Python: each logs a warning on first instantiation ([sequential_agent.ts#L47-L50](https://github.com/google/adk-js/blob/main-v2.0.0/core/src/agents/sequential_agent.ts#L47-L50), [parallel_agent.ts#L47-L50](https://github.com/google/adk-js/blob/main-v2.0.0/core/src/agents/parallel_agent.ts#L47-L50), [loop_agent.ts#L56-L59](https://github.com/google/adk-js/blob/main-v2.0.0/core/src/agents/loop_agent.ts#L56-L59), mechanism at [utils/deprecated.ts#L32-L47](https://github.com/google/adk-js/blob/main-v2.0.0/core/src/utils/deprecated.ts#L32-L47)). Go [v2.4.0](https://github.com/google/adk-go/blob/v2.4.0/agent/workflowagents/sequentialagent/agent.go) and Java [v1.10.1](https://github.com/google/adk-java/blob/v1.10.1/core/src/main/java/com/google/adk/agents/SequentialAgent.java) are not. This page has TypeScript tabs, so scoping the note to Python tells TS readers the opposite of the truth.

```suggestion
!!! note "Python and TypeScript: template workflow agents are deprecated"
```

Lines 14-19 need the same rescoping: `DeprecationWarning` is the Python spelling (TypeScript logs via `logger.warn`), and "which cannot be used as an `LlmAgent` sub-agent" currently attaches to the graph-based workflow link rather than to `Workflow`. Suggested body:

    In Python and TypeScript, `SequentialAgent`, `ParallelAgent`, and
    `LoopAgent` are deprecated: constructing one emits a deprecation
    warning, and the classes are scheduled for removal in a future
    release. The examples below still run. For new code, express these
    patterns with a [graph-based workflow](/graphs/); note that a
    `Workflow` cannot yet be used as an `LlmAgent` sub-agent.

docs/graphs/dynamic.md
```diff
@@ -430,7 +431,7 @@ workflows offer much more flexibility to define the routing logic you need.
         output_schema=str,
     )
 
-    @node # workflow node
+    @node(rerun_on_resume=True) # workflow node
     async def code_workflow(ctx: Context, user_request: str):
       code = await ctx.run_node(coder_agent, user_request)
       check_resp = await ctx.run_node(compile_lint_check, code)
```
joefernandez, line 437: The `return code` → `yield Event(output=code)` fix below is right, but this call still fails. `compile_lint_check` (line 420) is a `FunctionNode` with the default `parameter_binding="state"`, so a parameter not named `node_input` is read from `ctx.state` instead of from the value passed to `run_node` ([_function_node.py#L290-L358](https://github.com/google/adk-python/blob/v2.9.0/src/google/adk/workflow/_function_node.py#L290-L358); the default is set at [L121](https://github.com/google/adk-python/blob/v2.9.0/src/google/adk/workflow/_function_node.py#L121)). On this first call `"code"` is not in state yet, so it raises `ValueError: Missing value for parameter "code"` — the same defect the PR fixes in `city_time_function` and `task_attempt_node`.

Line 420 is outside the diff, so no suggestion block. The fix:

    @node(name="lint_reviewer")
    async def compile_lint_check(ctx: Context, node_input: str):

docs/graphs/data-handling.md
```diff
@@ -368,6 +369,8 @@ accepted and produced by any agent node.
     ***BaseModel*** to constrain any agent's input and output:
 
     ```python
+    from datetime import date
+
     from google.adk import Agent
     from pydantic import BaseModel
```
joefernandez, line 376: Good catch on the missing `date` import, but the snippet still does not run: `Flight` is undefined, so `class FlightSearchOutput(BaseModel): flights: list[Flight]` raises `NameError` at class-definition time, and `search_flights_api` is undefined at `tools=[search_flights_api]`. Adding both here keeps the example runnable:

```suggestion

    class Flight(BaseModel):
        carrier: str
        price: float

    def search_flights_api(
        origin: str, destination: str, departure_date: date, passengers: int
    ) -> list[dict]:
        """Look up flights with your own flight-search provider."""
        ...

```

PR #1761, opened 2026-05-12
Add AGENT_RESPONSE event type and view

haiyuan-eng-google:
## Summary

The BigQuery Agent Analytics plugin now logs an `AGENT_RESPONSE` event when the agent yields a final response to the user, with a corresponding `v_agent_response` view. This PR adds the missing documentation.

- Add `AGENT_RESPONSE` to the captured events summary table
- Add `v_agent_response` view row with columns: `response_text` (STRING), `source_event_id` (STRING), `source_event_author` (STRING), `source_event_branch` (STRING)
- Update auto-created view count from 16 to 17
- Add event type documentation with JSON payload example under "Agent lifecycle & Generic Events"

PR #2028, opened 2026-07-29
Enhance BigQuery Agent Analytics documentation

haiyuan-eng-google:
Updated documentation to clarify Java and Python plugin differences, added details on dropped-event observability, and improved explanations of event types and attributes.


koverholt (changes requested):
This PR re-introduces inline language/version callouts that will become overly complex and stale, and duplicated per-language tables, which were the same two conventions flagged in previous PRs (#1942, #1761). 

The content itself is accurate, but it needs to follow the conventions already established on this page. Please:

- Replace the inline version callouts with language support tags (`<div class="language-support-tag">`), rather than hard-coding versions in prose.
- Collapse the duplicated drop-reasons tables into one. This PR adds a second, parallel "Drop reasons (Java)" table alongside the existing Python one. Consolidate them (or use a single table with a per-language column) so we don't maintain duplicate entries, consistent with the feedback in #1942 about avoiding duplicate sections that restate an existing table.

haiyuan-eng-google:
Done in b362c3c.

**1. Version callouts → language support tags.** No hard-coded versions remain outside `language-support-tag` blocks. Added `Java v1.7.0` to the existing tag on **Agent workflow and pause/resume events (ADK 2.0)**, and new tags (`Python` + `Java v1.7.0`) on **Built-in redaction** and **Dropped-event observability**, then removed the corresponding prose. Also removed `(Java: v1.7.0 or later)` from the intro, `(v1.7.0+)` from the auto-created-views paragraph (it links to `#adk-2-events`, which carries the tag), and `(v1.7.0+)` from the two Java `Public methods` bullets. Dropped the "earlier Java builds created plugin-owned OpenTelemetry spans" sentence in Tracing and observability — the bullet now just states current behavior for both languages. I used a versionless `Python` span on the two new tags rather than guess a Python minimum for features that predate what the page documents; happy to pin exact versions if you'd prefer.

**2. Single drop-reasons table.** Collapsed into one table with a `Language` column. `queue_full` is one row ("Python, Java") carrying both config spellings; the remaining reasons are tagged per language. No duplicate entries.

Also trimmed the duplicated Java scope prose: the intro is one clause instead of two sentences, and the "Java support" note under `#adk-2-events` no longer restates the top-level-`attributes` detail that the "Java attribute location" note already covers.

koverholt (approved):
Thanks for the fixes. The updates LGTM.

PR #2031, opened 2026-07-30
docs(integrations): fix unresolvable imports and stale API claims

GWeale:
18 integration pages documented imports that raise, methods with the wrong
signature, or tool names that do not exist. Verified against
`google/adk-python` @ main — every rewritten import was executed, and every
signature and default introspected. 56 Python blocks, 0 syntax failures after
(2 before).

## Imports and calls that could not work

- `agent-identity.md` — `google.adk.tools.mcp` does not exist; it is
  `google.adk.tools.mcp_tool`. Also adds the missing
  `StreamableHTTPConnectionParams` import.
- `agent-registry.md` — the documented install could not import
  `AgentRegistry`: `agent_registry.py:35` imports agent-identity
  unconditionally at module scope. Now documents
  `pip install "google-adk[a2a,agent-identity]"`.
- `apigee-api-hub.md` — `tools=toolset.get_tools()` passed a coroutine
  (`APIHubToolset.get_tools` is async), raising `TypeError`. Passes the toolset.
- `cloud-trace.md` — `AdkApp` is `vertexai.agent_engines.AdkApp`, not
  `google.adk.apps`; `maybe_set_otel_providers` lives in
  `google.adk.telemetry.setup` and is not re-exported.
- `freeplay.md` — `App` is in `google.adk.apps`, not `google.adk.runners`.
- `skills-registry.md` — `google.adk.integrations.gcp_skill_registry` does not
  exist; the module is `google.adk.integrations.skill_registry`.

## Claims that did not match the code

- `skills-registry.md` — `search_skills` / `get_skill` are `async` and
  keyword-only, and `get_skill` has no `version` parameter.
- `gcs.md` — every documented tool name carried a `gcs_` prefix the real tools
  lack, so the documented `tool_filter` yielded an empty toolset.
- `reflect-and-retry.md` — a tool that *returns* an error is not retried by
  default (only raised exceptions are); `max_retries=3` gives 4 attempts, not 3,
  which the page's own text already said; `TrackingScope` is not exported from
  `google.adk.plugins`.
- `agent-registry.md` — `header_provider` only affects `get_mcp_toolset`;
  `RemoteA2aAgent` has no header field.
- `application-integration.md` — `tool_name_prefix` / `tool_instructions` are
  accepted and then discarded on the `integration=` path.
- `code-exec-agent-runtime.md` — named a class that does not exist, and said
  `tool_outputs` where the executor emits `tool_output`.
- `mlflow-scorers.md` — the class is `LLMRegistry`, not `LlmRegistry`.
- `api-registry.md` — `google.adk.tools.api_registry` is deprecated in favour of
  `google.adk.integrations.api_registry`.
- `windsor-ai.md` — `ADK_ENABLE_JSON_SCHEMA_FOR_FUNC_DECL` is on by default and
  no longer needs setting.
- `temporal.md`, `computer-use.md`, `cisco-ai-defense.md` — top-level `await`,
  unused imports, and `App` used without importing it.

Also adds the `is_final_response()` content guard where it was missing
(`phoenix.md`, `future-agi.md`).

### DIFF
--- a/docs/integrations/cloud-trace.md
+++ b/docs/integrations/cloud-trace.md
@@ -137,7 +137,7 @@ agent using the ADK CLI.
 
 === "Python"
 
-    If you are using the Vertex AI SDK `AdkApp` abstraction, you can enable cloud tracing by adding `enable_tracing=True`:
+    If you are using the Agent Platform SDK `AdkApp` abstraction, you can enable cloud tracing by adding `enable_tracing=True`:
 
     ```python
     from vertexai.agent_engines import AdkApp
--- a/docs/integrations/cloud-trace.md
+++ b/docs/integrations/cloud-trace.md
@@ -137,10 +137,10 @@ agent using the ADK CLI.
 
 === "Python"
 
-    If you are using the `AdkApp` abstraction, you can enable cloud tracing by adding `enable_tracing=True`:
+    If you are using the Vertex AI SDK `AdkApp` abstraction, you can enable cloud tracing by adding `enable_tracing=True`:
 
     ```python
-    from google.adk.apps import AdkApp
+    from vertexai.agent_engines import AdkApp
 
     adk_app = AdkApp(
         agent=root_agent,
@@ -155,14 +155,14 @@ For fully customized agent runtimes, you can enable cloud tracing by using the b
 === "Python"
 
     ```python
-    from google.adk import telemetry
     from google.adk.telemetry import google_cloud
+    from google.adk.telemetry.setup import maybe_set_otel_providers
 
     # Get GCP exporters configuration
     hooks = google_cloud.get_gcp_exporters(enable_cloud_tracing=True)
 
     # Initialize and set global OTel providers
-    telemetry.maybe_set_otel_providers(otel_hooks_to_setup=[hooks])
+    maybe_set_otel_providers(otel_hooks_to_setup=[hooks])
     ```
 
 === "TypeScript"
--- a/docs/integrations/code-exec-agent-runtime.md
+++ b/docs/integrations/code-exec-agent-runtime.md
@@ -72,7 +72,7 @@ example](#advanced-example) or the full
 
 ## How it works
 
-The `AgentEngineCodeExecutor` Tool maintains a single sandbox throughout an
+The `AgentEngineSandboxCodeExecutor` Tool maintains a single sandbox throughout an
 agent's task, meaning the sandbox's state persists across all operations within
 an ADK workflow session.
 
@@ -179,7 +179,7 @@ def base_system_instruction():
       print(df.shape)
       ```
       The output will be presented to you as:
-      ```tool_outputs
+      ```tool_output
       (49, 7)
 
       ```
@@ -189,11 +189,11 @@ def base_system_instruction():
       print(f'{{x=}}')
       ```
       The output will be presented to you as:
-      ```tool_outputs
+      ```tool_output
       x=999751168
 
       ```
-    - You **never** generate ```tool_outputs yourself.
+    - You **never** generate ```tool_output yourself.
     - You can then use this output to decide on next steps.
     - Print just variables (e.g., `print(f'{{variable=}}')`.
 
--- a/docs/integrations/gcs.md
+++ b/docs/integrations/gcs.md
@@ -188,7 +188,7 @@ Tool | Description
 ---- | -----------
 `list_buckets` | List GCS bucket names in a Google Cloud project.
 `create_bucket` | Create a new GCS bucket in a specific location. Requires `Capabilities.READ_WRITE`.
-`update_bucket` | Update properties of a GCS bucket (e.g. versioning or uniform bucket-level access). Requires `Capabilities.READ_WRITE`.
+`update_bucket` | Update properties of a GCS bucket, such as versioning or uniform bucket-level access. Requires `Capabilities.READ_WRITE`.
 `delete_bucket` | Delete a GCS bucket (bucket must be empty first). Requires `Capabilities.READ_WRITE`.
 
 ## Sample agents
--- a/docs/integrations/gcs.md
+++ b/docs/integrations/gcs.md
@@ -175,21 +175,21 @@ The GCS integration split the capabilities into two main toolsets:
 
 Tool | Description
 ---- | -----------
-`gcs_get_bucket` | Get metadata information about a GCS bucket.
-`gcs_list_objects` | List object names in a GCS bucket. Supports optional prefix filtering and pagination.
-`gcs_get_object_metadata` | Get metadata properties of a specific GCS object (blob).
-`gcs_create_object` | Create a new object (blob) in a bucket from in-memory string data or a local file upload.
-`gcs_get_object_data` | Get content of a GCS object as a string, or download it directly to a local file.
-`gcs_delete_objects` | Delete multiple GCS objects (blobs) from a bucket.
+`get_bucket` | Get metadata information about a GCS bucket.
+`list_objects` | List object names in a GCS bucket. Supports optional prefix filtering and pagination.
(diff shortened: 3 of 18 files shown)


docs/integrations/cloud-trace.md
```diff
@@ -137,10 +137,10 @@ agent using the ADK CLI.
 
 === "Python"
 
-    If you are using the `AdkApp` abstraction, you can enable cloud tracing by adding `enable_tracing=True`:
+    If you are using the Vertex AI SDK `AdkApp` abstraction, you can enable cloud tracing by adding `enable_tracing=True`:
```
koverholt: Since the platform rename from "Vertex AI" to "Agent Platform" a few months ago (#1673), we're avoiding adding any new instances of "Vertex AI" to the docs in prose (realizing that `vertexai` will still appear in some literal code and imports), so either replace this with:

> the Agent Platform SDK `AdkApp` abstraction

or revert to:

> the `AdkApp` abstraction

docs/integrations/code-exec-agent-runtime.md
```diff
@@ -179,7 +179,7 @@ def base_system_instruction():
       print(df.shape)
       ```
       The output will be presented to you as:
-      ```tool_outputs
+      ```tool_output
```
koverholt, line 182: This is a great find, and the changes in this PR are correct. No changes need here.

Just a heads-up that the upstream sample this is based on (https://github.com/google/adk-python/blob/main/contributing/samples/code_execution/agent_engine_code_execution/agent.py) still has the same bug in that it uses the ​`tool_outputs` delimiter instead of ​`tool_output`.

docs/integrations/reflect-and-retry.md
```diff
@@ -67,8 +67,11 @@ ADK project's App object, as shown below:
     ```
 
 
-With this configuration, if any tool called by an agent returns an error, the
-request is updated and tried again, up to a maximum of 3 attempts, per tool.
+With this configuration, if any tool called by an agent *raises* an exception,
```
koverholt: Wording nit: "raises an exception" is Python phrasing, but this page has support for both Python and Go. Could we make it language-neutral?

docs/integrations/reflect-and-retry.md
```diff
@@ -79,7 +82,9 @@ The Reflect and Retry Plugin has the following configuration options:
 *   **`throw_exception_if_retry_exceeded`**: (optional) If set to `False`, the
     system does not raise an error if the final retry attempt fails. Default
     value is `True`.
```
koverholt: Not introduced by this PR, but `throw_exception_if_retry_exceeded` defaulting to `True` is the Python default. The Go equivalent `WithErrorIfRetryExceeded` defaults to `false`, so the two SDKs behave differently once retries are exhausted (Python raises the original error; Go returns a guidance message instead). There are other issues on this page that apply only to Python and don't refer to how the Go SDK handles this, so this is not a blocker for this PR, optionally cleanup here or we can address in a future PR.

docs/integrations/reflect-and-retry.md
```diff
@@ -79,7 +82,9 @@ The Reflect and Retry Plugin has the following configuration options:
...
-*   **`tracking_scope`**: (optional)
+*   **`tracking_scope`**: (optional) A `TrackingScope` value, imported from
+    `google.adk.plugins.reflect_retry_tool_plugin` (it is not re-exported from
+    `google.adk.plugins`):
```
koverholt: This parenthetical:

> (it is not re-exported from`google.adk.plugins`)

reads a bit like a review note. Since the bullet already gives the exact import path, the "not re-exported" clause is redundant. Suggest dropping it.

docs/integrations/gcs.md
```diff
@@ -175,21 +175,21 @@ The GCS integration split the capabilities into two main toolsets:
...
 Tool | Description
 ---- | -----------
-`gcs_list_buckets` | List GCS bucket names in a Google Cloud project.
-`gcs_create_bucket` | Create a new GCS bucket in a specific location.
-`gcs_update_bucket` | Update properties of a GCS bucket (e.g. versioning or uniform bucket-level access).
-`gcs_delete_bucket` | Delete a GCS bucket (bucket must be empty first).
+`list_buckets` | List GCS bucket names in a Google Cloud project.
+`create_bucket` | Create a new GCS bucket in a specific location. Requires `Capabilities.READ_WRITE`.
+`update_bucket` | Update properties of a GCS bucket (e.g. versioning or uniform bucket-level access). Requires `Capabilities.READ_WRITE`.
+`delete_bucket` | Delete a GCS bucket (bucket must be empty first). Requires `Capabilities.READ_WRITE`.
```
koverholt: This change is incorrect and should be reverted. It makes the tool tables no longer match the actual tool names, so please restore the `gcs_` prefixes before merging.

These tables list the tools as exposed to the model, and those names keep the `gcs_` prefix. `GCSToolset`/`GCSAdminToolset` set `tool_name_prefix="gcs"`, and `LlmAgent` calls `get_tools_with_prefix()`, so the function declarations the model receives are `gcs_get_bucket`, `gcs_list_buckets`, etc. (matching the [upstream](https://github.com/google/adk-python/tree/main/contributing/samples/integrations/gcs) [sample](https://github.com/google/adk-python/tree/main/contributing/samples/integrations/gcs_admin) READMEs). Removing the prefix here makes the table no longer match what the model sees or what shows up in traces.

I think the confusion is that `tool_filter` matches the *unprefixed* names (filtering happens in `get_tools()` before the prefix is applied), so `tool_filter=['gcs_get_bucket']` yields an empty toolset while `['get_bucket']` works. That's a real issue, but it's about `tool_filter` and this page doesn't use `tool_filter` anywhere.

Could we *restore* the `gcs_` prefixes in both tables, *keep* the new `Requires Capabilities.READ_WRITE` notes (those are correct), and *add* a note/callout below the table to capture the difference:

```
!!! note

    The tool names listed here are the ones exposed to the model (prefixed with `gcs_`).
    When using `tool_filter`, reference the unprefixed names such as `get_bucket`.
```

koverholt (changes requested):
Thanks for the PR and fixes throughout the integration pages. 🎉 There are lots of excellent finds in here, like bad imports, deprecated APIs, and settings that are now default and no longer need to be set explicitly.

I left some inline comments to address/fix: one blocking issue in `gcs.md` and a naming/wording fix in `cloud-trace.md`, plus minor non-blocking comments in `code-exec-agent-runtime.md` and `reflect-and-retry.md`.

docs/integrations/gcs.md
GWeale: done, prefixes are back in both tables and i added your note. kept the READ_WRITE notes. good call on tool_filter matching before the prefix gets applied

docs/integrations/cloud-trace.md
GWeale: done, using "Agent Platform SDK" now

docs/integrations/reflect-and-retry.md
GWeale: done, reads "fails with an error" now

docs/integrations/reflect-and-retry.md
GWeale: took this one too rather than leaving it, noted the Go default and that it returns guidance instead of raising

docs/integrations/reflect-and-retry.md
GWeale: dropped

docs/integrations/code-exec-agent-runtime.md
GWeale: thanks, ill file the upstream sample bug against adk-python

koverholt (approved):
Thanks for working through the issues and review comments. All updated content LGTM.

PR #2032, opened 2026-07-30
docs(tutorials): fix ToolContext, output_key and persistence claims

GWeale:
Corrections to the tutorial, quickstart and grounding pages — the ones a new
user runs first. Verified against `google/adk-python` @ main; all 34 Python
blocks compile and all 3 JSON blocks parse.

## docs/tutorials/agent-team.md

- Said `ToolContext` must be **declared last** in a tool signature. It does not:
  ADK matches the context parameter on its annotation type at any position
  (`utils/context_utils.py:67-79`, `tools/function_tool.py:116`). Proved with
  the parameter first, middle, and renamed.
- A code comment claimed the delegated greeting turn overwrites
  `last_weather_report` via `output_key`. It does not —
  `agents/llm_agent.py:953-963` early-returns when `event.author != self.name`.
  The page's own prose 60 lines later already said so.
- Said persistent session storage "requires custom implementation or future ADK
  integrations". `DatabaseSessionService` and `VertexAiSessionService` already
  ship.
- A helper annotated `-> InMemorySessionService` that returns a `Session`; a
  constant whose name and value disagreed; `pip install litellm` unpinned when
  ADK requires `litellm>=1.84`.

## docs/get-started/streaming/quickstart-streaming.md

Said Callback, LongRunningTool, ExampleTool and Shell agents "will be supported
in future versions of ADK Streaming". All four work on the live path today —
each verified individually. Only **model** callbacks are genuinely absent, since
`_handle_*_model_callback` is reached only from the non-live `_call_llm_async`.
The caveat now states that limitation instead.

## docs/grounding/grounding_with_search.md

The documented `groundingChunks[].document{title,uri,id}` shape does not exist —
`types.GroundingMetadata.model_validate` rejects it (`extra_forbidden`). For
this page's datastore case the real field is `retrievedContext`. The corrected
payload round-trips; the old one fails validation. Also guards
`len(grounding_chunks)`, which raises `TypeError` when the field is `None`.

## Smaller

- `docs/get-started/python.md` — the project layout omitted the `.gitignore`
  that `adk create` writes, and the prose did not mention `adk create` is
  interactive (it prompts for model and backend).
- `docs/evaluate/custom_metrics.md` — prose named
  `check_final_response_match` and `response_match` where the example defines
  `check_final_response_exact_match`.


docs/get-started/python.md
```diff
@@ -50,6 +50,11 @@ Run the `adk create` command to start a new agent project.
 adk create my_agent
 ```
 
+The command is interactive: it prompts you to pick a model for the root agent
```
joefernandez: revert these changes; getting started updates need be limited to critical information only (brevity requirement) and also need to be consistent across languages 

these changes are unnecessary and create inconsistencies across the language versions

docs/get-started/python.md
```diff
@@ -59,6 +64,7 @@ file containing the main control code for the agent.
 my_agent/
     agent.py      # main agent code
     .env          # API keys or project IDs
+    .gitignore    # keeps .env out of version control
```
joefernandez: revert

docs/tutorials/agent-team.md
```diff
@@ -1894,7 +1897,7 @@ Your Weather Bot team is a great starting point. Here are some ideas to further
     *   Use `after_tool_callback` to process or log the results returned by a tool.
     *   Implement `before_agent_callback` or `after_agent_callback` for agent-level entry/exit logic.
 5.  **Error Handling:** Improve how the agent handles tool errors or unexpected API responses. Maybe add retry logic within a tool.
-6.  **Persistent Session Storage:** Explore alternatives to `InMemorySessionService` for storing session state persistently (e.g., using databases like Firestore or Cloud SQL – requires custom implementation or future ADK integrations).
+6.  **Persistent Session Storage:** Swap `InMemorySessionService` for one of ADK's persistent implementations – `DatabaseSessionService`, which is SQLAlchemy-backed and installed with `pip install google-adk[db]`, or `VertexAiSessionService`. See [Session](/sessions/session/) for details.
```
joefernandez: don't use "Swap" language; make it clear there is a decision to be made, here.

joefernandez:
### Technical review report

Verified claims in PR #2032 against `google/adk-python` @ [`d92bb42`](https://github.com/google/adk-python/commit/d92bb42ccfeac7255ad367974d79629ef368db32) (main, 2026-08-07), running the code in a clean venv with ADK 2.6.2 rather than just reading it. Full review is in the artifact.

**Verdict: approve with two wording nits.** All 11 substantive claims are technically correct.

**Strongest confirmations:**

- **`ToolContext` position** — [`context_utils.py#L66-L99`](https://github.com/google/adk-python/blob/d92bb42ccfeac7255ad367974d79629ef368db32/src/google/adk/utils/context_utils.py#L66-L99) iterates `signature.parameters` and matches on annotation; position is never consulted. [`function_tool.py#L140-L142`](https://github.com/google/adk-python/blob/d92bb42ccfeac7255ad367974d79629ef368db32/src/google/adk/tools/function_tool.py#L140-L142) also adds it to `_ignore_params`, confirming the PR's "hidden from the schema" clause. I built tools with the context param first/middle/renamed/last — all four detected, none leaked into `parameters_json_schema`.
- **`output_key`** — [`llm_agent.py#L978-L988`](https://github.com/google/adk-python/blob/d92bb42ccfeac7255ad367974d79629ef368db32/src/google/adk/agents/llm_agent.py#L978-L988) early-returns on `event.author != self.name`. Upstream has an exact regression test for this: [`test_llm_agent_output_save.py#L209-L239`](https://github.com/google/adk-python/blob/d92bb42ccfeac7255ad367974d79629ef368db32/tests/unittests/agents/test_llm_agent_output_save.py#L209-L239). Ran it: 18 passed.
- **`[db]` extra** — the ImportError ADK raises is verbatim the PR's wording: `Please install it by running: pip install google-adk[db]` ([`sessions/__init__.py#L50-L56`](https://github.com/google/adk-python/blob/d92bb42ccfeac7255ad367974d79629ef368db32/src/google/adk/sessions/__init__.py#L50-L56)).
- **Grounding** — the old `document` payload fails `GroundingMetadata.model_validate` with `extra_forbidden`; the PR's `retrievedContext` payload validates and round-trips. I validated the whole JSON block from the PR head, not just the hunk.
- **Both None-guards are real crash fixes**: `grounding_metadata` can be truthy with `grounding_chunks=None` (→ `TypeError`), and `is_final_response()` returns `True` on a content-less event (→ `AttributeError`).

**Two nits worth fixing before merge:**

1. "workflow agents such as `SequentialAgent`" over-generalizes. `SequentialAgent` is the *only* one with live support — [`loop_agent.py#L163-L167`](https://github.com/google/adk-python/blob/d92bb42ccfeac7255ad367974d79629ef368db32/src/google/adk/agents/loop_agent.py#L163-L167) and [`parallel_agent.py#L257-L261`](https://github.com/google/adk-python/blob/d92bb42ccfeac7255ad367974d79629ef368db32/src/google/adk/agents/parallel_agent.py#L257-L261) both `raise NotImplementedError`. "such as" invites a reader into a runtime crash.
2. "ADK only runs them around the non-live `generate_content` call" is imprecise: with `RunConfig.support_cfc=True`, `_call_llm_async` wraps `run_live()` while still firing model callbacks ([`base_llm_flow.py#L1480-L1482`](https://github.com/google/adk-python/blob/d92bb42ccfeac7255ad367974d79629ef368db32/src/google/adk/flows/llm_flows/base_llm_flow.py#L1480-L1482)). The conclusion still holds for this quickstart, but scoping it to "the `run_async` path" would be exact.

Also, the PR *description* cites two stale line numbers (`function_tool.py:116` is now L141; `llm_agent.py:953-963` is now L978-L988) — docs content is unaffected, but worth refreshing.

docs/live/get-started/streaming-python.md
joefernandez, line 1: This file has moved:

- Revert these changes. 
- Additionally, consider adding this change to the new /live/get-started/streaming-python.md page.

docs/get-started/python.md
joefernandez, line 1: revert these changes

docs/get-started/python.md
GWeale: reverted, python.md is out of the pr now

docs/get-started/python.md
GWeale: done.

docs/tutorials/agent-team.md
GWeale: done, reads "consider changing InMemorySessionService to one of ADK's
persistent implementations, such as ..." now

docs/live/get-started/streaming-python.md
GWeale: reverted. rebased onto main so the change follows the rename, it's on live/get-started/streaming-python.md now. also fixed two things in it: only SequentialAgent has live support (LoopAgent and ParallelAgent raise NotImplementedError), and scoped the callback claim to the run_async path since support_cfc=True still fires model callbacks.

docs/get-started/python.md
GWeale: done.

joefernandez (approved):

PR #2039, opened 2026-07-31
docs (python): Python API Reference docs v2.6.0

joefernandez:
- generated using https://github.com/google/adk-docs/tree/main/tools/python-api-docs


koverholt (approved):

koverholt:
I found a bug in the Python API docs ref generation script (or rather, a workaround we had to use for earlier versions of ADK 2.x.x that is now fixed), that is causing duplicate entries for every integration module.

I'll submit a PR to fix this issue in the generation script. And then you can either rerun the script and commit in this PR, or commit the new fixed API docs in a new PR.

Screenshot 2026-07-31 at 13 24 56

koverholt:
Done, the fixed version of the Python API ref generation script is in #2043.

You can use that fixed version to regenerate API ref docs and commit the new generated API docs to this PR (or a new PR if desired).

Duplicate API doc sections for integrations are fixed now (among other fixes/guards):

Screenshot 2026-07-31 at 14 05 41

PR #2043, opened 2026-07-31
Make the Python API reference generation script more robust

koverholt:
Fixes two correctness issues in the Python API reference generator and reduces manual maintenance.

1. Fixes duplicate entries for integrations in API ref docs.

    In `discover_modules.py`:

    - Remove the `NAMESPACE_SUBPACKAGES` workaround now that integrations ships an `__init__.py`, eliminating duplicate automodule entries for every `integrations.*` module.
    - Add `warn_namespace_packages()` to warn (instead of silently skipping) when a subpackage lacks an `__init__.py`, so it can be fixed upstream.

2. Fixes stale extras used for autodoc generation, which could result in empty API doc sections.

    In `generate.sh`:

    - Derive `PIP_EXTRAS` from adk-python's pyproject.toml (all groups minus `dev`,`test`,`benchmark`,`community`) instead of a stale hardcoded list
    - Clarify the `PYTHON_VERSION` comment to explain the pin satisfies adk-python's requires-python and python_version-gated extras.


tools/python-api-docs/discover_modules.py
```diff
@@ -90,20 +84,45 @@ def _collect_modules(path, prefix):
...
-    for ns in NAMESPACE_SUBPACKAGES:
-        mod = importlib.import_module(ns)
-        modules.extend(_collect_modules(mod.__path__, mod.__name__ + "."))
-
     modules.sort()
     return modules
 
 
+def warn_namespace_packages(pkg):
+    """Warn about subpackages lacking __init__.py (skipped by walk_packages).
```
joefernandez, line 103: I'm scratching my head about this warning function.

seems like there is a TODO here? does the script still produce usable output?

joefernandez (approved):
Approved, but see comment

tools/python-api-docs/discover_modules.py
koverholt: Yes! It will produce a warning at API ref docs generation time in case something we were guarding against and working around happens again in ADK 2.0.0. Here is the (long) backstory:

1. In ADK 2.0.0, the `integration/*` subpackage did not have an `__init__.py` file and thus we had to manually list it in `NAMESPACE_SUBPACKAGES` in this script so that those modules would get discovered and included in the generated API docs.
2. It looks like as of ADK 2.3.0, they added an `__init__.py` file [there](https://github.com/google/adk-python/commit/86254e3147106d7f0ade470fdd5267e6a431edc6) so we no longer need this `NAMESPACE_SUBPACKAGES` workaround in this script.
3. In this PR, I wanted to remove the `NAMESPACE_SUBPACKAGES` workaround since it's no longer needed for any other subpackages, but I wanted to guard from this happening in future releases. 

Since the `discover_modules` function crawls ADK subpackages to discover them, I decided to add this warning function in case it happens again in the future, we'll know at API docs generation time, and we can let the adk-python ENG team know that a newly added subpackage does not have an `__init__.py` file and thus won't show up in the API docs. Otherwise the subpackage would silently file to get included in the API docs, so this is an effort to keep it from silently failing in the future.

Note: Other SDK languages include their API docs config in their repos, so they maintain this kind of logic. ADK Python does not include their API docs config in the public-facing repo, which is why we have to keep  this kind of config up-to-date and in sync with subpackages and extras and such in our scripts.

PR #2052, opened 2026-08-03
fix: remove broken relative import in rag_engine example

AAliKKhan:
## Problem

The \
ag_engine.py\ snippet embedded in \docs/integrations/knowledge-engine.md\ cannot run:

1. \rom .prompts import return_instructions_root\ — the \prompts\ module does not exist anywhere in the repo, and a relative import fails when the file is run standalone (as readers do).
2. \instruction=return_instructions_root()\ references an undefined name.

## Fix

Removed the broken import and replaced \
eturn_instructions_root()\ with a self-contained instruction string describing the RAG agent's behavior.

## Why it matters

The code is the complete example on the Knowledge Engine integration page. Every reader who runs it hits \ModuleNotFoundError\.


examples/python/snippets/tools/built-in-tools/rag_engine.py
```diff
@@ -43,7 +42,11 @@
 root_agent = Agent(
     model="gemini-2.0-flash-001",
```
joefernandez:
```suggestion
    model="gemini-flash-latest",
```

joefernandez (approved):
Thanks for the fixes!

examples/python/snippets/tools/built-in-tools/rag_engine.py
```diff
@@ -19,7 +19,6 @@
 from vertexai.preview import rag
 
 from dotenv import load_dotenv
-from .prompts import return_instructions_root
```
joefernandez, line 22: Recording why this line had to go, for anyone who reads the diff later.

`from .prompts import return_instructions_root` was carried over from a multi-file agent layout, where a sibling `prompts.py` supplies `return_instructions_root()`. No such module exists beside this snippet — [`examples/python/snippets/tools/built-in-tools/`](https://github.com/google/adk-docs/tree/2e4d46eb153dbe1e282cc7b18342043b1c2f2898/examples/python/snippets/tools/built-in-tools) has no `prompts.py`, and neither does anywhere else under `examples/` — so the import could not succeed either way it was run:

- as a script (`python rag_engine.py`), the leading dot fails before the name is even looked up: `ImportError: attempted relative import with no known parent package`;
- as part of a package (e.g. `adk run <dir>`), the dot resolves and it then fails with `ModuleNotFoundError: No module named 'prompts'`.

That matters more here than in an ordinary example file, because this file is inlined verbatim into the published page by [`docs/integrations/knowledge-engine.md` L31](https://github.com/google/adk-docs/blob/2e4d46eb153dbe1e282cc7b18342043b1c2f2898/docs/integrations/knowledge-engine.md#L31) (`--8<--`), so a reader copying the Knowledge Engine snippet hit the error on the first line they ran, with nothing on the page telling them what `prompts` was meant to be.

Deleting the import and inlining the instruction as a string literal ([rag_engine.py L45-L49](https://github.com/google/adk-docs/blob/2e4d46eb153dbe1e282cc7b18342043b1c2f2898/examples/python/snippets/tools/built-in-tools/rag_engine.py#L45-L49)) is the right fix rather than adding a `prompts.py`: it keeps the snippet self-contained, so what the docs show is the whole of what a reader needs to run.

PR #2053, opened 2026-08-03
fix: await create_session in agent_search example

AAliKKhan:
## Problem

In \examples/python/snippets/tools/built-in-tools/agent_search.py\ (embedded in \docs/integrations/agent-search.md\), \InMemorySessionService.create_session\ is async but is called without \wait\ at module scope:

\\\python
session_vsearch = session_service_vsearch.create_session(...)
\\\

This produces an un-awaited coroutine (\RuntimeWarning\). The session object is later *not* awaited anywhere, so the code does not behave as intended.

## Fix

Wrapped the call in \syncio.run(...)\, matching the pattern already used in \code_execution.py\.

## Why it matters

Consistency with the rest of the built-in tools examples; readers copying this get a coroutine warning and an incorrect session object.


joefernandez (approved):
Thanks for the fix!

PR #2055, opened 2026-08-03
fix: await create_session and add license header in agent_cli example

AAliKKhan:
## Problem

\examples/python/snippets/tools/auth/agent_cli.py\ (embedded in \docs/tools-custom/authentication.md\) has two issues:

1. \session_service.create_session(...)\ is async but called without \wait\ inside \sync_main\. The result is a coroutine, so \session.id\ later raises \AttributeError\/fails.
2. The file is missing the Apache 2.0 license header, failing the License Header Lint CI check.

## Fix

- Added \wait\ to the \create_session\ call.
- Prepended the standard Google LLC / Apache-2.0 license header.

## Why it matters

This is the complete end-to-end OAuth/OIDC authentication example on the Tool Authentication page. As written, it crashes immediately after session creation.


examples/python/snippets/tools/auth/agent_cli.py
```diff
@@ -1,3 +1,17 @@
+# Copyright 2025 Google LLC
```
joefernandez:
```suggestion
# Copyright 2026 Google LLC
```

joefernandez (approved):
Thanks for the fixes!

PR #2056, opened 2026-08-03
docs: move get_bucket to admin toolset and simplify GCS IAM permissions

danyang-google:
Move get_bucket to admin toolset and simplify GCS IAM permissions


docs/integrations/gcs.md
```diff
@@ -187,6 +186,7 @@ Tool | Description
 Tool | Description
 ---- | -----------
 `gcs_list_buckets` | List GCS bucket names in a Google Cloud project.
+`gcs_get_bucket` | Get metadata information about a GCS bucket.
```
joefernandez, line 189: doesn't look like this command exists
- https://github.com/google/adk-python/blob/main/src/google/adk/integrations/gcs/admin_toolset.py#L38-L46

docs/integrations/gcs.md
```diff
@@ -30,7 +30,7 @@ and read/write objects.
 - **Enable the Google Cloud Storage API** in the target Google Cloud project.
 - **IAM Permissions**: The authenticated principal (Application Default
   Credentials, service account, or user) must have the correct permissions,
-  including `roles/storage.objectAdmin` and `roles/storage.admin`, to perform GCS
+  including `roles/storage.admin`, to perform GCS
```
joefernandez, line 33: Verified OK:  roles/storage.admin is a strict superset of roles/storage.objectAdmin, so this is redundancy removal, not a correctness fix.

docs/integrations/gcs.md
danyang-google: Good catch! The related code changed (https://github.com/google/adk-python/commit/ad9c113f6eeb8ca6d1c0ea304667cda21e071c0c) has been released in [v2.7.0](https://github.com/google/adk-python/releases/tag/v2.7.0) so it is safe to merge this now.

joefernandez (approved):

PR #2070, opened 2026-08-05
 docs: fix GCP skill registry import path in integration page

Syed-Umer-Ali:
## What
The example in "Use with Agent" imports from `google.adk.integrations.gcp_skill_registry`, which does not exist in the installed ADK.

## Why
Running `from google.adk.integrations.gcp_skill_registry import GCPSkillRegistry` raises `ModuleNotFoundError`. The module is actually `google.adk.integrations.skill_registry` (verified: `import google.adk.integrations.skill_registry` resolves and exposes `GCPSkillRegistry`; `gcp_skill_registry` does not exist in `google-adk` 2.6.0).

## Fix
Changed the import on line 61 to `from google.adk.integrations.skill_registry import GCPSkillRegistry`.

## Verification
`python -c "from google.adk.integrations.skill_registry import GCPSkillRegistry"` succeeds.


joefernandez:
Thanks for the update. Import already fixed.

PR #2071, opened 2026-08-05
docs: fix MCP toolset import and add missing StreamableHTTPConnectionParams import

Syed-Umer-Ali:
## What
Two problems in the "Configure tools" example: (1) imports `McpToolset` from `google.adk.tools.mcp`, which does not exist; (2) uses `StreamableHTTPConnectionParams` on line 97 without importing it.

## Why
`from google.adk.tools.mcp import McpToolset` raises `ModuleNotFoundError` — the correct module is `google.adk.tools.mcp_tool` (verified: `McpToolset` and `StreamableHTTPConnectionParams` both exist there). Without the second import the example raises `NameError` at runtime.

## Fix
Changed the import on line 87 to `from google.adk.tools.mcp_tool import McpToolset, StreamableHTTPConnectionParams`.

## Verification
`python -c "from google.adk.tools.mcp_tool import McpToolset, StreamableHTTPConnectionParams"` succeeds.


joefernandez:
Thanks for the update. Import already fixed.

PR #2045, opened 2026-07-31
docs: add Eventarc tool integration page and code snippets

milenvk:
## Summary
Adds official integration documentation and executable code snippets for the Google Cloud Eventarc toolset (`EventarcToolset`).

## Changes
- **New Integration Page (`docs/integrations/eventarc.md`)**:
  - Complete API reference for `EventarcToolset` and the default `publish_message` tool.
  - Dedicated documentation for domain-specific publishing (`EventarcToolset.create_publish_tool`).
  - Comprehensive attribute binding rules and sentinel reference table (`CloudEventAttributesBinding`, `AgentProvided`, `MISSING`, `OMIT`), including a dedicated example section clarifying `MISSING` vs `OMIT`.
  - Explains behavior when `payload_schema` is omitted (`None`), publishing notification-only CloudEvents without a JSON body.
- **Python Snippets (`examples/python/snippets/tools/built-in-tools/`)**:
  - `eventarc.py`: Demonstrates general-purpose `EventarcToolset` initialization and publishing CloudEvents.
  - `eventarc_domain_specific.py`: Demonstrates creating domain-specific publishing tools with static string, dynamic (`AgentProvided`), and runtime lambda attribute bindings, as well as Pydantic payload schemas.
- **Asset Card Icon**:
  - Adds square Eventarc icon at `docs/integrations/assets/eventarc.png`.

Staged: https://deploy-preview-2045--adk-docs-preview.netlify.app/integrations/eventarc/

## Verification
- Tested local documentation build with `mkdocs serve` and verified catalog card rendering.
- Verified Python snippets compile cleanly (`python3 -m py_compile`).
- Followed all `integration-create` and `integration-review` rules (concise verb-led description, canonical imports, experimental language badge, auto-discovered catalog cards).

PR #2074, opened 2026-08-05
docs(eventarc): Address technical verification report feedback

milenvk:
## Summary of Changes

This PR addresses the follow-up documentation items identified in the verification report on [google/adk-docs#2045](https://github.com/google/adk-docs/pull/2045#issuecomment-5184438669), aligned with the latest adk-python Eventarc capabilities:

1. **Prerequisites**: Added Step 4 (`pip install "google-adk[gcp]"`) to install the required GCP extra dependency.
2. **`datacontenttype` Description**: Clarified that `datacontenttype` defaults to `application/json` for dictionary/list payloads and `text/plain` for string payloads. In `examples/python/snippets/tools/built-in-tools/eventarc.py`, updated the prompt string to explicitly include `datacontenttype 'application/json'`.
3. **Runtime Lambda Binding**: Updated the table example from `ctx.id` to `ctx.session_id` and clarified that callables can inspect the event payload, runtime context, or both.
4. **`OMIT` Table Row**: Expanded the list of mandatory attributes that cannot be set to `OMIT` to include `id` and `specversion` (`type`, `source`, `bus`, `id`, `specversion`).

## Testing
- Verified against `google/adk-python` with all 63 Eventarc unit tests passing.


milenvk:
I have ran the integration-review skill and it came back clean with no recommendations for a change,

joefernandez (approved):

PR #2075, opened 2026-08-05
docs(get-started): make the TypeScript quickstart true — Node version, credentials, and expected output

AmaadMartin:
Three factual corrections to the TypeScript quickstart, applied inside the existing page structure.

> **This description was rewritten on 2026-08-10.** The PR originally restructured the page and this body described six fixes. @joefernandez [asked for that revert](https://github.com/google/adk-docs/pull/2075#pullrequestreview-3617374029) because the change broke the structure shared across all five language quickstarts, and he was right. The page was restored from `main` and only the corrections that fit inside the existing structure were re-applied. The old description was left up for four days describing work that is no longer in the diff — apologies to anyone who read it.

## What is actually in this diff

| File | Change |
|---|---|
| `get-started/typescript.md:6-7` | `Node.js 24.13.0` → `Node.js 22`, `npm 11.8.0` → `npm 9` |
| `get-started/typescript.md:46` | `npm install @google/adk` → `npm install @google/adk zod` |
| `get-started/typescript.md:63` | `name: 'get_current_time'` → `'getCurrentTime'` |
| `get-started/installation.md:54` | same `zod` addition |
| `get-started/index.md:16` | `:fontawesome-brands-js:` → `:simple-typescript:` |

**+6/−6.** The heading structure is byte-identical to `main`, including line numbers:

```
$ diff <(git show main:docs/get-started/typescript.md | grep -n '^#\{1,3\} ') \
       <(grep -n '^#\{1,3\} ' docs/get-started/typescript.md)
$ echo $?
0
```

### The three corrections

**Node version.** `@google/adk@1.5.0` declares no `engines` field, and neither does `@google/adk-devtools`. The whole quickstart runs on v22.22.2 / npm 9.2.0, and every TypeScript example in this repo pins `@types/node ^20`. 22 is the active LTS line and is what I tested — it's a judgement call, so change it if you'd prefer a different floor.

**`zod`.** The page's own sample imports it but never installs it. It currently works by accidental npm hoisting and fails under pnpm with `✘ [ERROR] Could not resolve "zod"`. Verified before and after.

**Tool name.** The code registered `get_current_time` while the instruction referred to `getCurrentTime`. Honesty note: this is a consistency fix, not a crash fix — I tested `main` unmodified and the model tolerates the mismatch. Drop it if you'd rather.

**`index.md`** gives the TypeScript card its own brand mark; Python, Go, Java and Kotlin all use theirs. Verified it renders (material ships `simple/typescript.svg`; `simple-kotlin` is already used on that page).

## Not in this PR

Other problems I hit while testing could not be fixed without adding structure the sibling quickstarts don't have, so they were dropped rather than forced in. They're enumerated in [this comment](https://github.com/google/adk-docs/pull/2075#issuecomment-5209157668) — the most significant being that **`GOOGLE_API_KEY`, which the other four quickstarts and `agents/models/google-gemini.md` all tell readers to use, is not accepted by the TypeScript runtime.** This page is the correct one and the cross-language pattern is wrong, so I changed nothing and left the call to you.

Verified: quickstart followed end to end from a clean directory on both credential paths, `zod` fix confirmed under pnpm, `mkdocs build` zero warnings.

### DIFF
--- a/docs/get-started/typescript.md
+++ b/docs/get-started/typescript.md
@@ -1,24 +1,58 @@
 # TypeScript Quickstart for ADK
 
-In this quickstart you build an agent that answers a question by calling a
-TypeScript function you wrote, and then talk to it from your terminal. Four
-steps, one file.
+This guide shows you how to get up and running with Agent Development Kit
+for TypeScript. Before you start, make sure you have the following installed:
 
-**Prerequisites**
+*   Node.js 22 or later
+*   Node Package Manager (npm) 9 or later
 
-*   **Node.js 22 or later.** Check with `node --version`. This guide was
-    verified end to end on Node.js v22.22.2 with npm 9.2.0.
-*   **npm, pnpm, or yarn.** Every command below is given for all three.
-*   **A Gemini API key.** Create one in Google AI Studio on the
-    [API Keys](https://aistudio.google.com/app/apikey) page. If you would rather
-    authenticate with Google Cloud, [step 3](#3-set-your-credentials) covers
-    that too.
+## Create an agent project
 
-## 1. Write the agent
+Create an empty `my-agent` directory for your project:
 
-Create a `my-agent` directory, and inside it an `agent.ts` file containing a
-[Function Tool](/tools-custom/function-tools/) named `getCurrentTime` and an
-agent that uses it:
+```none
+my-agent/
+```
+
+??? tip "Create this project structure using the command line"
+
+    === "MacOS / Linux"
+
+        ```bash
+        mkdir -p my-agent/
+        ```
+
+    === "Windows"
+
+        ```console
+        mkdir my-agent
+        ```
+
+### Configure project and dependencies
+
+Use the `npm` tool to install and configure dependencies for your project,
+including the package file, ADK TypeScript main
+library, and developer tools. Run the following commands from your
+`my-agent/` directory to create the `package.json` file and install the
+project dependencies:
+
+```console
+cd my-agent/
+# initialize a project as an ES module
+npm init --yes
+npm pkg set type="module"
+npm pkg set main="agent.ts"
+# install ADK libraries
+npm install @google/adk zod
+# install dev tools as a dev dependency
+npm install -D @google/adk-devtools
+```
+
+### Define the agent code
+
+Create the code for a basic agent, including a simple implementation of an ADK
+[Function Tool](/tools-custom/function-tools/), called `getCurrentTime`.
+Create an `agent.ts` file in your project directory and add the following code:
 
 ```typescript title="my-agent/agent.ts"
 import {FunctionTool, LlmAgent} from '@google/adk';
@@ -37,7 +71,7 @@ const getCurrentTime = new FunctionTool({
 });
 
 export const rootAgent = new LlmAgent({
-  name: 'helloTimeAgent',
+  name: 'hello_time_agent',
   model: 'gemini-flash-latest',
   description: 'Tells the current time in a specified city.',
   instruction: `You are a helpful assistant that tells the current time in a city.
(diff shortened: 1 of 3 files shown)


docs/get-started/typescript.md
joefernandez, line 1: This change does not follow the pattern of ADK getting started guides, which need to be consistent across languages. Rewriting the TypeScript getting started sample like this is not acceptable.  

Revert the changes to this file, otherwise I have to reject the whole PR.

AmaadMartin:
You're right, and I've reverted it. Thanks for the clear call.

I checked the other four quickstarts after your comment and the pattern is unambiguous — `## Create an agent project` → `### Define the agent code` → `### Configure project and dependencies` → `### Set your API key` → `## Run your agent` → `## Next: Build your agent`. My version renumbered every heading, flattened the H2/H3 hierarchy, and added a `## Troubleshooting` section no other language has. That's a template break, not an improvement, and someone reading Python and TypeScript side by side would have seen two different documents.

`docs/get-started/typescript.md` is now restored from `main` and the PR is **+6/−6** across three files. The heading structure is byte-identical to `main` — including line numbers, so nothing moved:

```
$ diff <(git show main:docs/get-started/typescript.md | grep -n '^#\{1,3\} ') \
       <(grep -n '^#\{1,3\} ' docs/get-started/typescript.md)
$ echo $?
0
```

**What's left, all inside the existing structure:**

| File | Change |
|---|---|
| `typescript.md:6-7` | `Node.js 24.13.0` → `Node.js 22`, `npm 11.8.0` → `npm 9` |
| `typescript.md:46` | `npm install @google/adk` → `npm install @google/adk zod` |
| `typescript.md:63` | `name: 'get_current_time'` → `'getCurrentTime'` |
| `installation.md:54` | same `zod` addition |
| `index.md:16` | `:fontawesome-brands-js:` → `:simple-typescript:` |

On the Node version: `@google/adk@1.5.0` declares no `engines` field, and neither does `@google/adk-devtools`. The whole quickstart runs on v22.22.2 / npm 9.2.0, and every TypeScript example in this repo pins `@types/node ^20`. 22 is the active LTS line and is what I actually tested — but it's my judgement call, so change it if you'd rather state something else.

On `zod`: the page's own sample imports it but never installs it. It currently works by accidental npm hoisting and fails under pnpm with `✘ [ERROR] Could not resolve "zod"`. Verified both before and after.

On the tool name: honesty note — this is a consistency fix, not a crash fix. I tested `main`'s version unmodified and the model tolerates the mismatch and still calls the tool. Drop it if you'd prefer.

**`index.md`** I kept, since giving TypeScript its own brand mark makes it *more* consistent with the other cards, not less — Python, Go, Java and Kotlin all use their own. Verified the icon renders (material ships `simple/typescript.svg`; `simple-kotlin` is already used on the same page). Happy to drop it if you disagree.

**`installation.md`** I cut back hard. My earlier version added nested npm/pnpm/yarn tabs that no other language tab on that page uses and took the TypeScript tab from 6 lines to 36 — the same mistake you flagged, in a second place. It's now the one-word `zod` fix.

---

### Things still wrong on that page that I did *not* fix

I found these while testing but none can be fixed without adding structure, so I've left them. Listing them so they're on the record — happy to open a separate issue, or a PR against whatever shape you'd want:

1. **`GOOGLE_API_KEY` doesn't work on TypeScript, and it's what we tell people to use.** The Python, Go, Java and Kotlin quickstarts all use `GOOGLE_API_KEY`, and so does `docs/agents/models/google-gemini.md`, which this page links to for model setup. The TypeScript runtime reads only `GOOGLE_GENAI_API_KEY` and `GEMINI_API_KEY`; `GOOGLE_API_KEY` fails with `Error: API key must be provided via constructor or GOOGLE_GENAI_API_KEY or GEMINI_API_KEY environment variable.` This page is the one that's *correct* and the cross-language pattern is the one that's wrong, so I deliberately changed nothing — it's your call which way to reconcile it.
2. **A bad API key produces no error and exit code 0.** One INFO line, no answer, nothing on stderr. I reproduced the identical silence from four distinct causes: bad key, unauthorized project, a model that doesn't exist in the chosen region, and a project without `aiplatform.googleapis.com` enabled. That's an SDK bug more than a docs bug, but it's the thing most likely to strand a reader who follows this page.
3. **The page's success state is a screenshot** (`adk-run.png`), so there's no text a reader can compare against, and following the page literally doesn't reach it.
4. **`.env` vs shell precedence is unstated, and the shell wins** — so the page can appear to work for the wrong reason if the reader has ambient `GOOGLE_CLOUD_*` set.

If a Troubleshooting section is something you'd want on *all* the quickstarts rather than just this one, I'm glad to do that as a separate, consistent pass across all five languages rather than bolting it onto TypeScript alone.

joefernandez:
None of the remaining changes are necessary, and the suggested Node.js version changes are against what the TS team specified in this change:
- https://github.com/google/adk-docs/commit/bd3f42410ee59ce8d086f721a432ac9f02fdc94c

This PR is an no-op. I am closing it.

PR #2076, opened 2026-08-05
docs(live): add a streaming quickstart for ADK TypeScript

AmaadMartin:
`adk.dev` has a streaming quickstart for Python and for Java. It has none for TypeScript. This adds one.

> **Scope note:** `CONTRIBUTING.md` asks for an issue before new documentation. Happy to open one and split or reshape this — the page is offered as a concrete starting point rather than a fait accompli.

## What it teaches, and why not what you might expect

The obvious page to write would mirror the Python one. That would have been wrong, because **`runLive` is not implemented in the TypeScript SDK**:

- `Runner.runLive` does not exist (`typeof runner.runLive === 'undefined'`; there is a TODO at `runner.ts:580`)
- `LlmAgent.runLiveFlow` throws at `llm_agent.ts:760`
- `StreamingMode.BIDI` is referenced nowhere, so passing it **silently degrades to no streaming at all**

So this page teaches what TypeScript actually does and what a JavaScript developer actually wants: **token streaming an agent's response into a browser UI over SSE.** Live audio is mentioned once, honestly, as not yet available in TypeScript. The `BIDI` trap is documented.

## The three things that decide whether a reader succeeds

1. **Streaming is off by default.** `StreamingMode.NONE` is the default (`run_config.ts:124`). Without `runConfig: { streamingMode: StreamingMode.SSE }` you get one blob and conclude streaming is broken. Measured: **6 events at 221/442/663/884/1105 ms with SSE, versus 1 event at 1109 ms without.**
2. **The final event repeats the entire answer.** A naive `text += chunk` renders it twice — measured at exactly **2.000×**. The page protects against this at all three consumer sites and explains why.
3. **Errors are events, not exceptions.** `runAsync` does not throw; a text-only loop renders silence.

## Verified, not composed

Every snippet was executed against a real `gemini-2.5-flash`. The reviewer who graded it ran the page's literal browser client and watched the DOM grow `37 → 400 → 765 → 1082 → 1162` chars across four frames, with `rendered twice: false`. Five error strings in Troubleshooting reproduce **verbatim**, including the `[400]` JSON block.

One bug this surfaced and fixes in the shipped example: Express 5 registers the callback passed to `app.listen(port, cb)` as an `'error'` listener too, so on a busy port the server **prints its success banner and exits 0** while never binding. The example now attaches `'listening'` and `'error'` separately and exits non-zero with an actionable message.

Includes runnable examples under `examples/typescript/snippets/live/streaming/`, wired in via `--8<--`; `mkdocs build --strict` is clean.

## How this was measured

Every page in this series was graded by an evaluator who **did the task using only the page** — installed, ran the code, hit the errors, and timed it with a stopwatch. The metric that mattered most was **page exits**: the number of times they had to leave the page to make progress (another docs page, GitHub, npm, reading `node_modules`, or guessing from types).

| Task | Before | After | Page exits before → after | Rubric |
|---|---|---|---|---|
| Get to a running agent with my own tool | 3.2 min¹ | **2.4 min** | **3 → 0** | 5/13 → **13/13** |
| Stream an agent's response into a UI | *no page existed* | **8.5 min**² | — → **1**³ | **13/13** |
| Expose my agent over A2A | *no page existed* | **6.7 min** | — → **0**⁴ | **13/13** |
| Call a remote A2A agent | *no page existed* | **9.1 min** | — → **0**⁴ | **12/13** |

¹ The old page was a **DNF** on its own terms: the evaluator only finished because they were handed a credential recipe that appeared in no document they were allowed to read. All three page exits were credentials — one of them to a linked page whose recipe does not work.
² 2.9 min to streaming in the console; 8.5 to a streaming browser UI.
³ The one remaining exit was the evaluator's own GCP project id, which no page can supply.
⁴ Both A2A pages exited once for a missing Vertex/ADC credential path; that tab was added and the exit is now closed.

For calibration the same protocol was run against **Mastra's getting-started page** on the identical task: **7.5 minutes, 6 page exits**, and a DNF from the page alone — it contains no agent or tool code and points at a GUI. The rubric itself is distilled from 15 pages of Mastra and Vercel AI SDK documentation, which are the bar this series is aiming at.

## Honest limitations

- There is no AI Studio API key on the machine this was verified on. Everything was executed end-to-end against a real `gemini-2.5-flash` over Vertex AI ADC. The API-key path was verified as far as credential resolution and the literal `API key not valid` failure; the successful-call transcript on that path is assembled from observed lines rather than a single capture. This is called out wherever it applies.
- Windows was not tested.


joefernandez:
## Technical verification review
Verdict: technically accurate

Verified against `google/adk-js` @ [`2c07ad3`](https://github.com/google/adk-js/commit/2c07ad3741cd84788d0b30a793587d5dd4b46106) (HEAD of `main`), `@google/adk` 1.5.0 (also npm `latest`).

**The three claims the page is built on all check out, and two of them are exact line hits:**

| Claim | Citation |
|---|---|
| `Runner.runLive` doesn't exist; TODO at `runner.ts:580` | [runner.ts#L580-L581](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/src/runner/runner.ts#L580-L581) — it's the last line of the class body |
| `LlmAgent.runLiveFlow` throws at `llm_agent.ts:760` | [llm_agent.ts#L755-L761](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/src/agents/llm_agent.ts#L755-L761) |
| Default is `StreamingMode.NONE` at `run_config.ts:124` | [run_config.ts#L119-L129](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/src/agents/run_config.ts#L119-L129), applied unconditionally at [runner.ts#L237](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/src/runner/runner.ts#L237) |
| `BIDI` silently degrades | streaming is gated on `=== StreamingMode.SSE` only: [llm_agent.ts#L1095-L1100](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/src/agents/llm_agent.ts#L1095-L1100) |
| Final event repeats the whole answer with `partial: false` | [streaming_utils.ts#L472](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/src/utils/streaming_utils.ts#L472) — and the SDK's own test asserts the duplicate: [sse_agent_test.ts#L142-L154](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/tests/integration/streaming_sse/sse_agent_test.ts#L142-L154) |
| Errors are events, not exceptions; `[400]` prefix | [llm_agent.ts#L1248-L1270](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/src/agents/llm_agent.ts#L1248-L1270) |

Also verified verbatim: the `Sending out request … stream: false` log line ([google_llm.ts#L168-L170](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/src/models/google_llm.ts#L168-L170)), the API-key error string ([#L118-L122](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/src/models/google_llm.ts#L118-L122)), `GOOGLE_API_KEY` genuinely not being read ([#L396-L401](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/src/models/google_llm.ts#L396-L401)), the `adk` bin living in `@google/adk-devtools` ([dev/package.json#L21-L23](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/dev/package.json#L21-L23)), and the Express 5 `listen`-callback-as-error-listener bug ([express@v5.1.0 application.js#L598-L606](https://github.com/expressjs/express/blob/v5.1.0/lib/application.js#L598-L606)).

## Issues found

1. **Substantive — `server.ts` loses text once the agent has tools.** Forwarding only `event.partial` drops the consolidated text event the aggregator emits when one chunk carries both text and a `functionCall` ([streaming_utils.ts#L301-L320](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/src/utils/streaming_utils.ts#L301-L320)); the repo's test shows that text arriving *only* as `partial: false` ([sse_agent_test.ts#L98-L112](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/tests/integration/streaming_sse/sse_agent_test.ts#L98-L112)). Harmless for the tool-free agent built here, but "Next steps" points readers straight at tools.
2. **"`runAsync` does not throw" is absolute where the code isn't** — it throws on `Session not found` ([runner.ts#L263-L270](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/src/runner/runner.ts#L263-L270)) and re-throws non-`Error` failures. Say "model failures do not throw".
3. **Wrong cause for the CJS/top-level-await error** — `@google/adk` is dual-published with a CJS build ([core/package.json#L20-L33](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/package.json#L20-L33)); the error comes from the examples' own top-level `await`. The prescribed fix is still right.
4. Nit: `errorCode` is set for any non-`STOP` finish reason, so `[${errorCode}] ${errorMessage}` can render `[MAX_TOKENS] undefined` ([streaming_utils.ts#L465-L469](https://github.com/google/adk-js/blob/2c07ad3741cd84788d0b30a793587d5dd4b46106/core/src/utils/streaming_utils.ts#L465-L469)).
5. Nit: "Node.js 22 or later" is the author's choice — adk-js declares no `engines` field.

Docs-side checks also pass: all seven internal links resolve on `adk-docs@main`, `lst-typescript` exists in `docs/stylesheets/language-tags.css`, and the `--8<--` paths follow the existing `examples/typescript/snippets/` convention (with `check_paths: true` enforcing it).

AmaadMartin:
Thank you for this — the line-level citations made every item directly actionable. All five are fixed in `5dfb2e9`. Issue 1 was a real bug and you were right that "Next steps" walks readers straight into it.

### 1. Text loss with tools — fixed, and it was worse than latent

You're right about the mechanism. When one chunk carries both text and a `functionCall`, the aggregator emits that text **only** in the consolidated `partial: false` event and never as a delta, so forwarding only `event.partial` drops it.

The awkward part is that dropping non-partial text is exactly what prevents the double-render, so a naive fix trades one bug for a worse one. I added a small `TurnText` helper that appends deltas and, on a consolidated event, emits only the tail beyond what has already been shown, then resets on the same boundary the aggregator does.

Measured, where 1.000× is "renders exactly the answer":

| Case | Before | After |
|---|:-:|:-:|
| Tool agent, the mixed-chunk shape your test asserts | **0.662×** — lost `"Let me show the forecast. "` | **1.000×** |
| Tool agent, live Vertex model | **0.832×** — lost 57 chars | **1.000×** |
| Tool-free agent (the one this page builds) | 1.000× | **1.000×** |
| Tool-free, naive "forward everything" fix | — | 2.000× ✗ |

End-to-end over HTTP through the page's own `index.html` parser: 221 chars / 2 sentences before, 272 / 3 after, still 6 incremental updates, no duplication.

One caveat I want on the record rather than buried: probing the raw stream, **Vertex on `global` and `us-central1` never actually merged text and a `functionCall` into one chunk**, so on that backend the bug is latent. To get a live repro I moved the chunk boundary to the shape `sse_agent_test.ts:98-112` asserts. I have no AI Studio key, so I couldn't check whether that backend chunks differently in practice. Your citation is what established the shape is real; my live test only confirms the handling.

### 2. "does not throw" — narrowed

Confirmed both: `Session not found` throws from `runner.ts:269`, and the non-`Error` re-throw at `llm_agent.ts:1279`. Reworded to "model failures do not throw" everywhere it appeared.

### 3. CJS explanation — corrected, remedy kept

You're right, it's dual-published: `require('@google/adk')` resolves to `dist/cjs/index.js` and works. The error is from the example's own top-level `await`. While fixing the explanation I found my text was wrong a second way — it said "these files" use top-level await, but only `stream-to-console.ts` does, so `npm start` actually survives the misconfiguration. Both corrected.

### 4. `[MAX_TOKENS] undefined` — reproduced and fixed

Obtained it with a low `maxOutputTokens`: `errorCode="MAX_TOKENS"`, `errorMessage=undefined`. Now prints `[MAX_TOKENS]` with no trailing `undefined`. I also moved the error check to *after* the text write, since that event carries the last of the answer — writing the error first discarded it.

### 5. Node 22 — reworded as the tested floor

Agreed, it's a recommendation, not something read off the package. Phrased as the tested/supported floor rather than a requirement.

---

`tsc --noEmit` clean at both 1.5.0 and a clean-install 1.6.0; `npm ls` clean; `mkdocs build` 0 warnings.

Two things worth your judgement, since this page has no counterpart in the other languages:

- **It teaches SSE token streaming, not `runLive`.** Given `Runner.runLive` doesn't exist and `BIDI` silently degrades, mirroring the Python page's shape would have documented something that doesn't work. So this page deliberately doesn't parallel `streaming-python.md`. If that's a structural problem the way the getting-started rewrite was in #2075, tell me now and I'll reshape it before you spend more review time — I'd rather match the house pattern than win the argument.
- **It ends in a browser UI** rather than `adk web`, because the task a JS developer arrives with is "stream this into my own frontend". That's also a departure from the Python page, and the same question applies.

kazunori279:
Thank you for this, and I'm sorry to be arriving with this answer after the amount of care you put in.

I re-verified the whole page before deciding — adk-js cloned at HEAD (`9360bf2`, past the `2c07ad3` Joe checked against), `@google/adk` 1.6.0 resolved from your `^1.5.0`, Node 25, real `gemini-2.5-flash` over Vertex. It all held: SSE gave 6 events where the default gave 1, naive appending measured exactly 2.000×, `TurnText` passed the `sse_agent_test.ts` mixed-chunk shape plus the adversarial cases I threw at it, the browser client grew the DOM `0 → 140 → 497 → 859 → 1183 → 1414` with no double render, the `[400]` and `[MAX_TOKENS]` output reproduced verbatim, the Express 5 `listen` trap reproduced exactly as you described it, and `flushHeaders()` was 15 ms versus 2844 ms. `mkdocs build --strict` clean, `tsc --noEmit` clean at 1.6.0. Your write-up was accurate throughout.

**After consulting the stakeholders, we've decided not to take this page, and I'd like to ask you to close the PR.** The reason is placement, not quality.

The page teaches `models.generateContentStream()` — ordinary HTTP token streaming of a text reply. That is a different feature from what `docs/live/` documents, which is the bidirectional Live API over `live.connect()` / `BidiGenerateContent`: audio and video in both directions, interruption, `responseModalities`. `live/index.md` defines the section that way and carries a Python-only support tag.

You identified the underlying reason yourself: the TypeScript SDK has no official Live API support. `Runner.runLive` does not exist and `LlmAgent.runLiveFlow` throws. So there was no TypeScript slot for you to fill under Live, and the page landed in the nearest one. Publishing it beside the Python and Java live quickstarts would tell readers that TypeScript has live agents — which is why your page has to spend a paragraph taking that back. We'd rather not put readers in a position where the navigation and the page disagree.

That leaves the page without a home in the current information architecture, and reshaping it into something that fits would mean either documenting `runLive`, which does not run, or rebuilding it around `adk web`, which `get-started/typescript.md` already covers and which would cost the parts of your page that are genuinely worth having.

Two things I want on the record, because they're not consolation:

- When `runLive` ships in adk-js, the TypeScript live quickstart will need writing, and I would be glad if you wrote it. Please ping me when you see that land.
- Per `CONTRIBUTING.md`, an issue first is the right path for a new page. If you'd like to pursue a home for the SSE text-streaming material outside `live/`, please open one and we can settle the placement before you write, rather than after.

Thanks again. The failure-mode coverage here — the 2.000× duplication trap, `BIDI` degrading with no warning, the Express 5 banner-and-exit-0 bug — is better than most of what we ship, and none of that reflects on the decision.

AmaadMartin:
@kazunori279 — that's the right call, and thank you for re-verifying the whole thing at HEAD before making it. Closing as asked.

The reasoning is one I'd rather have than a merge: `docs/live/` is scoped to the bidirectional Live API, TypeScript has no Live API support, and a page that needs a paragraph to walk back what its own navigation implies is a page in the wrong place. I'd noticed the tension — it's why I asked Joe whether the departure from `streaming-python.md` was a structural problem — but I framed it as "does this page match the template" when the real question was "does this section's subject exist in this language". That's the more useful test and I'll apply it first from now on.

Yes to writing the TypeScript live quickstart when `runLive` lands. I'll watch `Runner.runLive` and `LlmAgent.runLiveFlow` in adk-js and ping you.

On the SSE material: I'd like to find it a home, and I'll follow the issue-first path this time rather than arriving with a finished page. Before I open anything, one question so I propose against your IA rather than my guess — is text streaming something you'd want as its own topic (something like `runtime/streaming.md`, next to the other runtime concerns), or as a section inside the existing runtime docs, or is it out of scope for the docs site altogether? Happy with any of those answers, including the last one.

One thing worth flagging regardless of what happens to the page, since it'll outlive it: `StreamingMode.BIDI` is accepted by the API, referenced nowhere in the implementation, and silently degrades to non-streaming with no warning. A TypeScript reader who reasons "bidi is a superset of SSE" gets strictly worse behaviour than if they'd asked for SSE, and nothing tells them. That's an SDK issue rather than a docs one — happy to file it against adk-js if it's useful.

kazunori279:
Thanks for closing it, and for the reframing — "does this section's subject exist in this language" is exactly the test, and it's a better articulation of it than mine.

## On where the SSE material should go

Of your three options, I'd go with the first: its own topic under **Run Agents → Agent Runtime**, alongside `cancel.md` and `runconfig.md`.

Two things make that the cleanest fit:

- `runtime/cancel.md` is already tagged TypeScript-only, so a single-language page in that section is an established precedent rather than an exception. Python, Go and Java tabs can be added later without moving anything.
- `runtime/runconfig.md#enable-streaming` already exists and already has a TypeScript tab. So the *reference* is covered — what's missing is the task-oriented walkthrough, which is a different genre and shouldn't be crammed into a config reference page.

Which means the new page should assume `runconfig.md` for the "what are the modes" material and link to it, rather than restate it. Concretely, I'd expect something noticeably shorter than #2076: the SSE loop, the duplication trap and `TurnText`, the error-as-event handling, and the SSE-over-HTTP plumbing — and drop the project scaffolding and credentials setup, which `get-started/typescript.md` covers.

Please open the issue with that shape and we'll settle scope there before you write.

## On `StreamingMode.BIDI`

Yes, please file it against adk-js. Your description is right, and I reproduced it while reviewing: `BIDI` yields one event with no error and no warning, the enum member appears nowhere in the implementation outside its own doc comment, and streaming is gated solely on `=== StreamingMode.SSE`. A reader who assumes bidi ⊇ sse gets strictly worse behaviour than if they had asked for less. Silent degradation is the part worth emphasising in the issue — an unsupported-mode error would be fine.

You also surfaced a docs bug on our side while you were at it. `runconfig.md#enable-streaming` tells readers that for bidirectional streaming they should "use `runner.run_live()` instead" — which is true for Python and not true for TypeScript, on a page that carries a TypeScript tab. We'll fix that regardless of what happens with the new page. Thank you.

AmaadMartin:
Both filed, thank you — the placement reasoning made the scoping question easy to answer.

- **google/adk-js#676** — `StreamingMode.BIDI` silent degradation. Led on silence being the defect rather than the missing feature, and listed throw / remove / warn in that order, quoting your "an unsupported-mode error would be fine". Credited your independent repro.
- **#2124** — the streaming page proposal, shaped as you described: `runtime/`, assumes `runconfig.md` for the modes and `get-started/typescript.md` for setup, keeps the SSE loop, the duplication trap, the tool-chunk text loss, error-as-event, and minimal plumbing. Three open questions in there, including how far to take the HTTP plumbing. Not writing anything until scope is agreed.

I verified both of your precedent claims before leaning on them: `runtime/cancel.md:3-5` is indeed TypeScript-only with no other tag, and `runconfig.md:106` does carry the `run_live()` line on a page with TypeScript tabs.

One adjacent nit on that page while you're in it: `runconfig.md:97` says *"set the `streaming_mode` parameter"* in language-neutral prose above the tabs, where the TypeScript property is `streamingMode`. The code tabs themselves are correct (`:34`, `:134`) — it's only the prose. Same shape as the `run_live()` line: text written in one language's conventions sitting above multi-language tabs. Happy to fold both into a one-line PR if that's less work than doing it yourselves; equally happy to leave it with you.

PR #2077, opened 2026-08-05
docs(a2a): add A2A exposing and consuming quickstarts for ADK TypeScript

AmaadMartin:
The A2A quickstarts exist for Python, Go and Java. The sidebar **enumerates those three languages and omits TypeScript** — so the gap is visible to every reader — even though `@google/adk` ships a complete, fully-exported A2A implementation. This adds both TypeScript quickstarts.

> **Scope note:** `CONTRIBUTING.md` asks for an issue before new documentation. Happy to open one and split this into two PRs — they are combined here because they cross-link and share one example project, and splitting them would leave a broken intermediate state.

## Contents

- `docs/a2a/quickstart-exposing-typescript.md` — serve an ADK agent over A2A, proven with a real agent-card fetch and a real `message/send`
- `docs/a2a/quickstart-consuming-typescript.md` — call a remote A2A agent, both **directly** and composed as a **`subAgent`** so the model routes to it
- `examples/typescript/a2a_basic/` — one runnable project serving both pages, mirroring the Go example's layout
- nav entries in house order (Python, TypeScript, Go, Java)

## What running it actually taught us

These pages are built from a working server and client, not from reading the source. Each of the following silently breaks an unwarned reader, and each is now documented with its literal symptom:

- **`toA2a()` returns an Express app that does not listen.** You call `app.listen()`.
- **`toA2a()` throws unless you pass `allowUnauthenticated: true`.** The library fails closed, which is good; the quickstart says so, says why, and links to the authenticated version.
- **`agentCard` is a base URL, not the card URL.** Passing the card URL yields a doubled `/.well-known/.well-known/…` **and crashes the process with an uncaught rejection**, because `await this.init()` sits outside the `try` in `runAsyncImpl`. This is the single most likely reader failure and it leads the troubleshooting.
- **`port` in the options only decorates the agent card — it does not start a server.** A mismatch with `app.listen()` surfaces as a bare `fetch failed`.
- **Auth rejections return HTTP 500 / JSON-RPC `-32603`, not 401**, and the agent-card endpoint is always public.
- **`basePath` behaves differently from its JSDoc**, and a client base URL needs a trailing slash or `new URL()` silently drops the segment.
- Express and `@a2a-js/sdk` are direct dependencies, not peers — nothing extra to install.

## Verification

Agent card fetched over HTTP is **byte-identical** to the JSON pasted in the page. `message/send` returns `"state": "completed"` with the real artifact chain. The authenticated server was run three ways: no token → `-32603`, wrong token → `-32603`, correct token → completed. The `subAgent` routing case was run **four times**, with the remote process logging the tool call every time. `npx tsc --noEmit` clean; `mkdocs build` **zero warnings**.

One thing worth flagging as a finding rather than a fix: the reviewer's first pass at the authenticated example shipped a *fragment* labelled as a complete file. It compiled in the author's head and not in `tsc`, and at runtime it returned `-32603` to **correctly**-authenticated requests — which the page's own troubleshooting then misattributed to a bad token. A reader who took the security advice would have ended up worse off than one who ignored it. It is now the real file, included via `--8<--`, with `crypto.timingSafeEqual` and the secret read from the environment, and it was run.

Also: the example's original routing instruction handed off to the remote agent only **2 of 3 times**. The instruction was tightened to 8/8, and the page documents the failure mode rather than hiding it.

## How this was measured

Every page in this series was graded by an evaluator who **did the task using only the page** — installed, ran the code, hit the errors, and timed it with a stopwatch. The metric that mattered most was **page exits**: the number of times they had to leave the page to make progress (another docs page, GitHub, npm, reading `node_modules`, or guessing from types).

| Task | Before | After | Page exits before → after | Rubric |
|---|---|---|---|---|
| Get to a running agent with my own tool | 3.2 min¹ | **2.4 min** | **3 → 0** | 5/13 → **13/13** |
| Stream an agent's response into a UI | *no page existed* | **8.5 min**² | — → **1**³ | **13/13** |
| Expose my agent over A2A | *no page existed* | **6.7 min** | — → **0**⁴ | **13/13** |
| Call a remote A2A agent | *no page existed* | **9.1 min** | — → **0**⁴ | **12/13** |

¹ The old page was a **DNF** on its own terms: the evaluator only finished because they were handed a credential recipe that appeared in no document they were allowed to read. All three page exits were credentials — one of them to a linked page whose recipe does not work.
² 2.9 min to streaming in the console; 8.5 to a streaming browser UI.
³ The one remaining exit was the evaluator's own GCP project id, which no page can supply.
⁴ Both A2A pages exited once for a missing Vertex/ADC credential path; that tab was added and the exit is now closed.

For calibration the same protocol was run against **Mastra's getting-started page** on the identical task: **7.5 minutes, 6 page exits**, and a DNF from the page alone — it contains no agent or tool code and points at a GUI. The rubric itself is distilled from 15 pages of Mastra and Vercel AI SDK documentation, which are the bar this series is aiming at.

## Honest limitations

- There is no AI Studio API key on the machine this was verified on. Everything was executed end-to-end against a real `gemini-2.5-flash` over Vertex AI ADC. The API-key path was verified as far as credential resolution and the literal `API key not valid` failure; the successful-call transcript on that path is assembled from observed lines rather than a single capture. This is called out wherever it applies.
- Windows was not tested.

### DIFF
--- a/docs/a2a/index.md
+++ b/docs/a2a/index.md
@@ -20,6 +20,7 @@ Navigate through the guides below to learn about ADK's A2A capabilities:
 
   *   **[A2A Quickstart (Consuming) for Python](./quickstart-consuming.md)**
       * **[A2A Extension - V2 Implementation](./a2a-extension.md)**
+  *   **[A2A Quickstart (Consuming) for TypeScript](./quickstart-consuming-typescript.md)**
   *   **[A2A Quickstart (Consuming) for Go](./quickstart-consuming-go.md)**
   *   **[A2A Quickstart (Consuming) for Java](./quickstart-consuming-java.md)**
 
--- a/docs/a2a/index.md
+++ b/docs/a2a/index.md
@@ -1,7 +1,7 @@
 # ADK with Agent2Agent (A2A) Protocol
 
 <div class="language-support-tag">
-  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python</span><span class="lst-go">Go</span><span class="lst-java">Java</span><span class="lst-preview">Experimental</span>
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python</span><span class="lst-typescript">TypeScript</span><span class="lst-go">Go</span><span class="lst-java">Java</span><span class="lst-preview">Experimental</span>
 </div>
 
 With Agent Development Kit (ADK), you can build complex multi-agent systems where different agents need to collaborate and interact using [Agent2Agent (A2A) Protocol](https://a2a-protocol.org/)! This section provides a comprehensive guide to building powerful multi-agent systems where agents can communicate and collaborate securely and efficiently.
@@ -13,6 +13,7 @@ Navigate through the guides below to learn about ADK's A2A capabilities:
   Start here to learn the fundamentals of A2A by building a multi-agent system with a root agent, a local sub-agent, and a remote A2A agent. The following guides cover how do I expose your agent so that other agents can use it via the A2A protocol:
 
   *   **[A2A Quickstart (Exposing) for Python](./quickstart-exposing.md)**
+  *   **[A2A Quickstart (Exposing) for TypeScript](./quickstart-exposing-typescript.md)**
   *   **[A2A Quickstart (Exposing) for Go](./quickstart-exposing-go.md)**
   *   **[A2A Quickstart (Exposing) for Java](./quickstart-exposing-java.md)**
 
--- a/docs/a2a/intro.md
+++ b/docs/a2a/intro.md
@@ -265,4 +265,4 @@ concerns and easy integration of specialized agents.
 Now that you understand the "why" of A2A, let's dive into the "how."
 
 - **Continue to the next guide:**
-  Quickstart: Exposing Your Agent, [in Python](./quickstart-exposing.md), [in Go](./quickstart-exposing-go.md), [in Java](./quickstart-exposing-java.md)
+  Quickstart: Exposing Your Agent, [in Python](./quickstart-exposing.md), [in TypeScript](./quickstart-exposing-typescript.md), [in Go](./quickstart-exposing-go.md), [in Java](./quickstart-exposing-java.md)
--- a/docs/a2a/quickstart-consuming-typescript.md
+++ b/docs/a2a/quickstart-consuming-typescript.md
@@ -1,94 +1,49 @@
 # Quickstart: Consuming a remote agent via A2A
 
 <div class="language-support-tag">
-  <span class="lst-supported">Supported in ADK</span><span class="lst-typescript">TypeScript v1.5.0</span><span class="lst-preview">Experimental</span>
+  <span class="lst-supported">Supported in ADK</span><span class="lst-typescript">TypeScript</span><span class="lst-preview">Experimental</span>
 </div>
 
-An A2A agent runs somewhere else — another process, another machine, another team's
-service — and speaks the [Agent2Agent protocol](intro.md) over HTTP. This page answers the
-first question you hit when you meet one: **how do I call it from my ADK TypeScript agent
-and get a real answer back?**
+This quickstart covers the most common starting point for any developer: **"There is a remote agent, how do I let my ADK agent use it via A2A?"**. This is crucial for building complex multi-agent systems where different agents need to collaborate and interact.
 
-You will build two things, because they are different jobs: a script that calls a remote
-agent **directly**, and a local agent that has the remote agent as a **sub-agent**, so the
-model decides when to route to it. You will also start the remote agent yourself, so
-nothing on this page depends on someone else's server being up.
+## Overview
 
-Here is the whole direct client. This is the complete file — the only thing missing is a
-remote agent to point it at, which you start in step 4.
+This sample demonstrates the **Agent2Agent (A2A)** architecture in the Agent Development Kit (ADK), showcasing how multiple agents can work together to handle complex tasks. The sample implements an agent that can roll dice and check if numbers are prime.
 
-```typescript title="examples/typescript/a2a_basic/direct_client.ts"
---8<-- "examples/typescript/a2a_basic/direct_client.ts:full"
+```text
+┌─────────────────┐    ┌──────────────────┐    ┌────────────────────┐
+│   Root Agent    │───▶│   Roll Agent     │    │   Remote Prime     │
+│  (Local)        │    │   (Local)        │    │   Agent            │
+│                 │    │                  │    │  (localhost:8001)  │
+│                 │───▶│                  │◀───│                    │
+└─────────────────┘    └──────────────────┘    └────────────────────┘
 ```
 
-Let's take a look at what is happening in this code:
+The A2A Basic sample consists of:
 
-1. **`new RemoteA2AAgent({...})`** is the client. It is imported from `@google/adk` — there
-   is no `@google/adk/a2a` subpath — and it extends `BaseAgent`, so anywhere ADK accepts an
-   agent, it accepts this.
-2. **`agentCard: 'http://localhost:8001'` is a base URL, not the URL of the card.**
-   `RemoteA2AAgent` appends `.well-known/agent-card.json` itself. Passing the card URL is
-   the most common way to break this page; see [Troubleshooting](#troubleshooting).
-3. **`name` and `description`** are required by ADK, not fetched from the remote card. The
-   `description` is what a parent agent's model reads when deciding whether to route here,
-   so write it for that audience.
-4. **`Runner`** drives the agent exactly as it drives a local one: `runAsync()` returns an
-   async iterable of events. There is no separate "A2A client" API to learn.
-5. **`event.errorMessage`** is how remote failures arrive. They are *not* thrown, and the
-   process still exits `0`. A loop that only reads `event.content` shows you nothing but
-   your own prompt when the remote agent fails.
-6. **`event.partial`** marks the streamed chunks of an answer that is still being written.
-   Skipping them prints the finished sentence once. Drop that line and the same run prints
-   `Yes, 7 is a`, then ` prime number.`, then `Yes, 7 is a prime number.` again.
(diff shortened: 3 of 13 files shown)


AmaadMartin:
Applying your #2075 feedback here proactively, before you have to spend review time saying it twice.

You were right that these pages have to be consistent across languages, and I'd made the same mistake here — worse, in fact. I'd invented my own structure *and* my own anchor slugs, when Python, Go and Java all deliberately share theirs (`{ #getting-the-sample-code }`, `{ #start-the-remote-a2a-agent-server }`, and so on). Those are clearly stable cross-language link targets and I broke them.

Both pages are now rebuilt on the house template.

**Anchors now match exactly:**

```
quickstart-exposing-typescript   #getting-the-sample-code #check-that-your-remote-agent-is-running #run-the-main-consuming-agent
quickstart-exposing-go           #getting-the-sample-code #check-that-your-remote-agent-is-running #run-the-main-consuming-agent
quickstart-exposing (python)     #getting-the-sample-code #check-that-your-remote-agent-is-running #run-the-main-consuming-agent

quickstart-consuming-typescript  #getting-the-sample-code #start-the-remote-prime-agent-server #look-out-for-the-required-agent-card-of-the-remote-agent #run-the-main-consuming-agent
quickstart-consuming-go          #getting-the-sample-code #start-the-remote-prime-agent-server #look-out-for-the-required-agent-card-of-the-remote-agent #run-the-main-consuming-agent
```

**Spine now matches too** — `## Overview` → `## Exposing the Remote Agent with the `toA2a()` function` → the four numbered steps → `## Example Interactions` → `## Next Steps`. The consuming page is now structurally identical to the Go page, section for section. I also added `## Example Interactions`, which both my pages were missing and every sibling has, and fixed `## Next steps` → `## Next Steps`.

**Length: 599 → 299 and 625 → 289.** For reference Python is 283/306 and Go is 130/156, so these are now in the same band rather than double it.

### What I cut, honestly

The `## Troubleshooting` H2 is gone from both — that's the section you rejected on #2075 and no A2A page has one. The highest-value entries survive as inline `!!! warning` admonitions at the step where a reader actually hits them (matching the admonition style the sibling pages already use); the rest are dropped. Also cut: an ~80-line raw task-JSON dump, a ~30-line provider error dump, and duplicated code includes.

Two cuts I want to flag rather than bury, because they removed genuinely useful and correct content:

1. **The `Authorization`-header snippet for `RemoteA2AAgent`** is now undocumented anywhere. It compiles and was tested. Say the word and I'll put it back somewhere appropriate.
2. **The `.well-known/.well-known` 404** is the single most likely first failure — `agentCard` takes a *base* URL, and passing the card URL crashes the process with an uncaught rejection because `await this.init()` sits outside the `try` in `runAsyncImpl`. It survives as a one-line `!!! warning` in step 4 of the consuming page rather than a full entry.

### What I deliberately kept, and why

**Credential setup**, relocated into step 1 on both pages rather than removed. Worth calling out because it's a real divergence from the siblings: neither the Go nor the Python A2A quickstart mentions credentials at all, yet both start agents that make model calls. When an evaluator ran these pages cold, that was the only thing that blocked them. Both the API-key and Vertex tabs are there, plus a note on which of the three processes actually needs the key. If you'd rather these pages stay silent on credentials to match the others, I'll pull it — but I think the others have a gap rather than this having an excess.

Everything was re-run after the restructure: dice server + card fetch + `message/send`, and prime server + direct client + subAgent routing. Every transcript on the pages is a verbatim capture from today's runs, and each `jq` command shown is the exact command that produced the block under it. `tsc --noEmit` clean, `mkdocs build` 0 warnings.

One process note: per `CONTRIBUTING.md` these are new pages and should have had an issue first. Happy to open one retroactively, or to close this and re-propose whatever shape you'd prefer.

AmaadMartin:
@joefernandez — requesting a review on this one when you have time. (I don't have permission to add reviewers from a fork, hence the mention.)

Since I opened it I've restructured both pages onto the shared A2A template, applying your #2075 feedback without waiting to be told twice — details in the [comment above](https://github.com/google/adk-docs/pull/2077#issuecomment-5209299635). Short version:

- **Anchors now byte-match the siblings** (`#getting-the-sample-code`, `#start-the-remote-prime-agent-server`, `#look-out-for-the-required-agent-card-of-the-remote-agent`, `#run-the-main-consuming-agent`). I'd invented my own, which broke what look like deliberate cross-language link targets.
- **Spine matches**: `## Overview` → `## Exposing the Remote Agent with …` → the four numbered steps → `## Example Interactions` → `## Next Steps`. The consuming page is now section-for-section identical to the Go page.
- **599 → 299 and 625 → 289 lines** (Python is 283/306). The `## Troubleshooting` H2 you rejected on #2075 is gone; the highest-value entries survive as inline admonitions at the step where a reader hits them.
- Just merged latest `main` in, so it carries the `TypeScript ADK` nav casing from #2079 rather than reverting it. `mkdocs build` exit 0, zero warnings; CLA, header-check and the Netlify preview are all green.

Both flows were re-run end to end after the restructure — every transcript on the pages is a verbatim capture, and each `jq` command shown is the one that produced the block beneath it.

Two things I'd genuinely like your call on rather than guessing:

1. **Credentials.** I kept the API-key and Vertex setup inside step 1 on both pages. That's a deliberate divergence — neither the Go nor the Python A2A quickstart mentions credentials at all, yet both start agents that make model calls, and it was the only thing that blocked someone running these cold. I think the siblings have a gap, but if you'd rather these match and stay silent, I'll pull it.
2. **Process.** Per `CONTRIBUTING.md` these are new pages and should have had an issue first. Happy to open one retroactively, or to close this and re-propose in whatever shape you'd prefer — including splitting exposing and consuming back into two PRs if that reviews more easily. They're combined here only because they cross-link and share one example project.

No rush, and no objection if the answer is "reshape it first."

AmaadMartin:
@kazunori279 closed #2076 on a placement test that I should apply here myself rather than leave for a reviewer to discover:

> *"`live/index.md` defines the section that way and carries a Python-only support tag... Publishing it beside the Python and Java live quickstarts would tell readers that TypeScript has live agents — which is why your page has to spend a paragraph taking that back. We'd rather not put readers in a position where the navigation and the page disagree."*

I ran that test against this PR. **I believe A2A passes where Live failed, but the decision is yours and I'd rather hand you the evidence than the conclusion.**

### The section's subject exists in TypeScript

| | scope stated by the section index | TypeScript status |
|---|---|---|
| `docs/live/` | *"low-latency **bidirectional voice and video** interaction"* | `Runner.runLive` absent, `runLiveFlow` throws → **no slot** |
| `docs/a2a/` | *"agents ... collaborate and interact using **Agent2Agent (A2A) Protocol**"* | fully implemented and exported |

All five A2A symbols resolve from the **published** `@google/adk@1.5.0` root, not just from source:

```
toA2a                EXPORTED (function)
RemoteA2AAgent       EXPORTED (function)
A2AAgentExecutor     EXPORTED (function)
getA2AAgentCard      EXPORTED (function)
AGENT_CARD_PATH      EXPORTED (string)
```

Everything on both pages was run against that published version — a real agent card over HTTP, a completed `message/send`, and a model routing to a remote agent as a sub-agent.

### The tell Kaz identified is absent here

The signal that #2076 was misplaced was that the page had to take back what the nav implied. Grepping these two pages for `not supported`, `not yet`, `not implemented`, `does not exist`, `does not work` returns **nothing**. There's no caveat paragraph because there's no gap to apologise for.

### Two gaps I found, disclosed rather than buried

I went looking for reasons this *shouldn't* land, and found two. One is a non-issue, one is real:

1. **Live mode over A2A is unimplemented — in both languages.** TypeScript's `A2ARemoteAgent` throws `Live mode is not supported in A2ARemoteAgent yet.`; adk-python's `remote_a2a_agent.py:834` raises `_run_live_impl for ... via A2A is not implemented.` **Parity, so not a differentiator.**
2. **Task cancellation is a genuine gap.** TypeScript's `A2AAgentExecutor.cancelTask()` throws `Task cancellation is not supported yet.`, where adk-python implements `async def cancel(...)` in `a2a_agent_executor_impl.py:69`. **TypeScript is behind Python here.**

Neither quickstart uses either feature, and no A2A quickstart in any language mentions cancellation — so the pages themselves are accurate. But #2 does bear on **the index badge**, which this PR changes:

```diff
-  <span class="lst-python">Python</span><span class="lst-go">Go</span><span class="lst-java">Java</span>
+  <span class="lst-python">Python</span><span class="lst-typescript">TypeScript</span><span class="lst-go">Go</span><span class="lst-java">Java</span>
```

An unqualified `TypeScript` there asserts A2A support at the same level as the others, and on cancellation that isn't quite true. Options, your call: leave it as-is (the section is already tagged Experimental for everyone), add a version qualifier like the other badges use elsewhere, or drop the badge change and let the two quickstarts stand on their own. **I'll do whichever you prefer, including dropping it.**

### And if the answer is still no

If TypeScript A2A isn't considered released in the sense that matters for the docs site — the same standard that closed #2076 — then say so and I'll close this too, with no argument. I'd rather learn the bar than litigate one page.

### Process

Both you and Joe have now pointed at issue-first for new pages, and you're right that I've had it backwards twice. If you'd like, I'll open an issue for these two pages and we can settle placement before you spend more review time; I left them open rather than opening one unprompted because the pages already exist and I didn't want to double your notifications. Say the word.

kazunori279:
@joefernandez — could you take this one? You already reviewed #2075, so the context is yours, and the three decisions Amaad is asking for are all yours to make rather than mine: the `a2a/index.md` badge, whether the credential setup stays or matches the silent siblings, and the issue-first process question.

@AmaadMartin — one thing I can answer, since it was my test: **applying the #2076 placement criterion here yourself, unprompted, was the right instinct, and I think your reading is correct.** The reason #2076 failed was that `live/index.md` scopes its section to bidirectional voice and video, TypeScript has no such entry point, and the page therefore had to spend a paragraph retracting what the navigation promised. Neither half of that holds here: A2A is implemented and exported in TypeScript, and your grep for retraction language coming back empty is the right way to check for the tell. I have no objection on the grounds that closed #2076.

That is the narrow question, though, not the review. The badge in particular is a real question and you were right to surface `cancelTask()` rather than let it be found later — but where that lands is Joe's call, not mine.

Thanks for the disclosure of the two gaps, and for checking parity with adk-python on both instead of only reporting the TypeScript side. That is the useful shape for this kind of report.

PR #2078, opened 2026-08-05
fix(docs): the Grounding Lite TypeScript MCP sample names a transport that does not exist

AmaadMartin:
The TypeScript sample for the Grounding Lite MCP server **cannot connect**. Not "is slightly wrong" — it never opens a transport.

## The finding

```
error TS2820: Type '"SseConnectionParams"' is not assignable to type
'"StdioConnectionParams" | "StreamableHTTPConnectionParams"'
```

`SseConnectionParams` **has never existed in the TypeScript SDK.** All 21 published versions of `@google/adk` (0.1.0 → 1.5.0) were unpacked and grepped: zero occurrences, in any release, including `0.2.0` — the version this page's own badge names. There is no "it used to work" defence.

Runtime is worse than the type error suggests. `mcp_session_manager.js` switches on `type` with exactly two cases, so `"SseConnectionParams"` falls through to a **no-op `default`** and `client.connect()` is never called. Observed directly:

```
SseConnectionParams          → createSession() RESOLVED, transport=undefined  (NO TRANSPORT)
StreamableHTTPConnectionParams → THREW: fetch failed                          (real attempt)
```

The agent starts, has no transport, and reports nothing.

`headers` is not a valid field either — `TS2561: Did you mean to write 'header'?` — so the API key is dropped even if the type is corrected in isolation.

## Why it looks the way it does

This is a transliteration. `SseConnectionParams` is a real class in the Python SDK, and the Python sample directly above this one on the same page carries the identical `headers` keys and values. A scan of all 41 TypeScript MCP code fences in `docs/` found **15 remote samples in `docs/integrations/` that all already use `transportOptions.requestInit.headers`** — `mcp-tools.md:484` is the sole outlier.

## The fix

`type: "StreamableHTTPConnectionParams"` with `transportOptions: { requestInit: { headers: { "X-Goog-Api-Key": … } } }`, verified `tsc --noEmit` **exit 0** under `strict`.

The `Accept` and `Content-Type` headers are also dropped: the MCP SDK overwrites both (`streamableHttp.js:298-299`), so the old comment was justifying an inert line. The replacement comment describes what the endpoint actually does — a single-URL POST that may answer with either JSON or an event stream — which is accurate and requires no reference to any other language's SDK.

One caveat worth knowing: `transportOptions` exists from **0.3.0** onward. That is consistent with all 15 sibling samples; the alternative `header:` field is `@deprecated`.

The dead symbol appears nowhere else in a TypeScript context. Its 13 remaining occurrences across 3 files are all Python samples, where it is a real class, and are left alone.

`mkdocs build`: exit 0, zero warnings.

### DIFF
--- a/docs/tools-custom/mcp-tools.md
+++ b/docs/tools-custom/mcp-tools.md
@@ -501,14 +501,19 @@ export const rootAgent = new LlmAgent({
     description: "A helpful assistant for planning travel.",
     tools: [
         new MCPToolset({
-            // Using SseConnectionParams to connect to the remote Grounding Lite service,
-            // mirroring Python's StreamableHTTPConnectionParams.
-            type: "SseConnectionParams",
+            // Grounding Lite is a remote MCP server reached over streamable HTTP,
+            // so the connection is a StreamableHTTPConnectionParams. Custom headers
+            // travel on every request via transportOptions.requestInit; the
+            // transport sets Accept and Content-Type itself, so only the API key
+            // needs to be supplied here.
+            type: "StreamableHTTPConnectionParams",
             url: "https://mapstools.googleapis.com/mcp",
-            headers: {
-                "X-Goog-Api-Key": googleMapsApiKey,
-                "Content-Type": "application/json",
-                "Accept": "application/json, text/event-stream"
+            transportOptions: {
+                requestInit: {
+                    headers: {
+                        "X-Goog-Api-Key": googleMapsApiKey
+                    }
+                }
             }
         })
     ],


AmaadMartin:
@joefernandez a small nudge on this one, on severity rather than age — it's a 12-line fix and I think it's the highest-impact item of the set I opened, but it's queued behind two much larger pages.

The TypeScript sample on `docs/tools-custom/mcp-tools.md` cannot connect. Not "is slightly off" — it opens no transport at all:

```
SseConnectionParams            → createSession() RESOLVED, transport=undefined  (NO TRANSPORT)
StreamableHTTPConnectionParams → THREW: fetch failed                            (real attempt)
```

`mcp_session_manager.js` switches on `type` with two cases; `"SseConnectionParams"` matches neither and falls through to a no-op `default`, so `client.connect()` is never called. The agent starts, silently does nothing, and reports no error — which is the worst shape a broken sample can have, because the reader has nothing to search for.

The symbol has **never existed** in the TypeScript SDK. All 21 published versions of `@google/adk` (0.1.0 → 1.5.0) unpacked and grepped: zero occurrences, including `0.2.0`, the version this page's own badge names. It's a Python class transliterated into a TypeScript sample — the Python example directly above it on the same page carries the identical `headers` keys.

One command to confirm the whole thing:

```
$ npx tsc --noEmit   # on the sample as published
error TS2820: Type '"SseConnectionParams"' is not assignable to type
'"StdioConnectionParams" | "StreamableHTTPConnectionParams"'
```

The fix uses `StreamableHTTPConnectionParams` with `transportOptions.requestInit.headers`, matching the **15 other remote MCP samples** already in `docs/integrations/` — this page is the sole outlier. `tsc --noEmit` exit 0, `mkdocs build` 0 warnings.

No urgency beyond that, and happy to keep waiting if the queue is the queue. Flagging it only because it's cheap to review and currently ships code that cannot run.

joefernandez:
# Technical review — adk-docs PR #2078

- **Review Score**: 98 / 100
- **Summary reasoning**: The change's central claim is correct and I verified it against the current release of `google/adk-js` (`adk-v2.0.0`) and the MCP TypeScript SDK it binds to. `SseConnectionParams` genuinely does not exist in adk-js, and the replacement — `StreamableHTTPConnectionParams` with `transportOptions.requestInit.headers` — is the documented, non-deprecated API. The explanatory comment's factual claims are also accurate. Only cosmetic nits remain inside the diff.
- **Recommended action**: Approve. The one-line-of-code fix is a real bug fix: the previous sample failed both compilation and runtime. The only things I'd consider are trimming the five-line comment and, separately from this PR, the pre-existing defects in the neighbouring Java sample and the missing peer-dependency note.

I worked **without a configured brief** for this level — no level-specific instructions were supplied in the job. I used the [L2-review-technical-accuracy skill](https://source.corp.google.com/piper///depot/google3/experimental/cloud-ai-docs/adk-docs-skills/L2-review-technical-accuracy/SKILL.md), which I was able to reach, and reviewed against the current GitHub releases of the ADK libraries.

## Verification performed

Everything below was checked against tag [`adk-v2.0.0`](https://github.com/google/adk-js/releases/tag/adk-v2.0.0) (the latest `google/adk-js` release, published 2026-08-21) and MCP SDK [`v1.26.0`](https://github.com/modelcontextprotocol/typescript-sdk/tree/v1.26.0), the version adk-js declares as its peer.

**1. The premise is right — `SseConnectionParams` does not exist.** A repo-wide search of adk-js at `adk-v2.0.0` returns zero matches for `SseConnectionParams`. The union is only two members:

> `export type MCPConnectionParams = StdioConnectionParams | StreamableHTTPConnectionParams;`
> — [mcp_session_manager.ts#L71-L73](https://github.com/google/adk-js/blob/adk-v2.0.0/core/src/tools/mcp/mcp_session_manager.ts#L71-L73)

The old code was broken twice over, not merely stylistically wrong:
- **Compile time**: `type: "SseConnectionParams"` is not assignable to the union, and `headers` is an excess property — the (deprecated) field is singular `header`, per [mcp_session_manager.ts#L56-L61](https://github.com/google/adk-js/blob/adk-v2.0.0/core/src/tools/mcp/mcp_session_manager.ts#L56-L61).
- **Runtime**: the `switch` in `createSession()` would fall through to `default:`, so `client.connect(transport)` is never called, the unconnected client is added to `activeSessions`, and the first `session.listTools()` throws. See [mcp_session_manager.ts#L104-L146](https://github.com/google/adk-js/blob/adk-v2.0.0/core/src/tools/mcp/mcp_session_manager.ts#L104-L146).

**2. The replacement API is correct.** `StreamableHTTPConnectionParams.transportOptions?: StreamableHTTPClientTransportOptions` ([L65](https://github.com/google/adk-js/blob/adk-v2.0.0/core/src/tools/mcp/mcp_session_manager.ts#L65)), and that SDK type has `requestInit?: RequestInit` ([streamableHttp.ts#L100](https://github.com/modelcontextprotocol/typescript-sdk/blob/v1.26.0/src/client/streamableHttp.ts#L100)). The library's own JSDoc names exactly this path as the successor to the deprecated field: *"Use transportOptions.requestInit.headers instead."*

**3. The comment's claims hold up.** "Custom headers travel on every request" — `_commonHeaders()` merges `this._requestInit?.headers` via `normalizeHeaders`, which handles a plain object ([streamableHttp.ts#L198-L203](https://github.com/modelcontextprotocol/typescript-sdk/blob/v1.26.0/src/client/streamableHttp.ts#L198-L203), [transport.ts#L9-L21](https://github.com/modelcontextprotocol/typescript-sdk/blob/v1.26.0/src/shared/transport.ts#L9-L21)), and it is used by the POST send path, the GET SSE stream, and DELETE terminate. "The transport sets Accept and Content-Type itself" — confirmed, and it in fact *overrides* any caller-supplied values, because the merged `Headers` is assigned after the `...this._requestInit` spread:

> ```ts
> const headers = await this._commonHeaders();
> headers.set('content-type', 'application/json');
> headers.set('accept', 'application/json, text/event-stream');
> const init = { ...this._requestInit, method: 'POST', headers, ... };
> ```
> — [streamableHttp.ts#L466-L476](https://github.com/modelcontextprotocol/typescript-sdk/blob/v1.26.0/src/client/streamableHttp.ts#L466-L476)

So dropping `Content-Type`/`Accept` is not just tidier, it removes two settings that never had any effect. `X-Goog-Api-Key` survives untouched.

**4. Surrounding sample still type-checks.** `MCPToolset` is re-exported from the package root ([core/src/index.ts#L50](https://github.com/google/adk-js/blob/adk-v2.0.0/core/src/index.ts#L50)), so `import {LlmAgent, MCPToolset} from "@google/adk"` is valid, and `MCPToolset extends BaseToolset` is accepted in `tools:` since `ToolUnion = BaseTool | BaseToolset | BaseNode` ([llm_agent.ts#L283](https://github.com/google/adk-js/blob/adk-v2.0.0/core/src/agents/llm_agent.ts#L283)).

**5. No unreleased-code risk.** `transportOptions` is present in [`adk-v1.6.0`](https://github.com/google/adk-js/blob/adk-v1.6.0/core/src/tools/mcp/mcp_session_manager.ts) as well as `adk-v2.0.0`, so the sample works on the current and previous releases.

**6. Consistent with the rest of the doc set.** The new shape matches the pattern already used in every other TypeScript streamable-HTTP MCP sample: [windsor-ai.md#L87-L98](https://github.com/google/adk-docs/blob/a1b155152974cc4576b8231e37b8a826888423eb/docs/integrations/windsor-ai.md#L87-L98), [supermetrics.md#L88-L99](https://github.com/google/adk-docs/blob/a1b155152974cc4576b8231e37b8a826888423eb/docs/integrations/supermetrics.md#L88-L99), [stripe.md#L152-L163](https://github.com/google/adk-docs/blob/a1b155152974cc4576b8231e37b8a826888423eb/docs/integrations/stripe.md#L152-L163), [postman.md#L157-L168](https://github.com/google/adk-docs/blob/a1b155152974cc4576b8231e37b8a826888423eb/docs/integrations/postman.md#L157-L168). This PR brings the last outlier into line.

## 🟢 Low risk items

- **Comment is long for a code sample.** [mcp-tools.md#L504-L508](https://github.com/google/adk-docs/blob/a1b155152974cc4576b8231e37b8a826888423eb/docs/tools-custom/mcp-tools.md#L504-L508) spends five lines explaining transport internals inside a snippet the reader is meant to copy. The "why Accept/Content-Type are absent" rationale is genuinely useful but reads more like a commit message than sample code. None of the four sibling samples cited above carry any comment at all. Optional tightening:

  ```suggestion
              // Grounding Lite is a remote MCP server reached over streamable
              // HTTP. Custom headers go in transportOptions.requestInit; the
              // transport sets Accept and Content-Type itself.
  ```

- **Now diverges from the adjacent Python and Java samples.** The Python sample at [mcp-tools.md#L349-L356](https://github.com/google/adk-docs/blob/a1b155152974cc4576b8231e37b8a826888423eb/docs/tools-custom/mcp-tools.md#L349-L356) and the Java one at [#L434-L437](https://github.com/google/adk-docs/blob/a1b155152974cc4576b8231e37b8a826888423eb/docs/tools-custom/mcp-tools.md#L434-L437) still pass `"Content-Type"` and `"Accept"` explicitly. Neither is *wrong* (both SDKs also set their own), but a reader working through the three tabs now sees a TypeScript comment asserting those headers are unnecessary sitting directly below two samples that supply them. Not a blocker for this PR; worth a follow-up to make the three consistent.

joefernandez (approved):
Thanks for the update!

PR #2080, opened 2026-08-05
chore(examples): bump @google/adk to 1.5.0 in five TypeScript examples that no longer type-check

AmaadMartin:
Five TypeScript examples were pinned at `@google/adk` versions old enough that **three of them no longer type-check**. This bumps those five to `1.5.0`.

| Example | Before | After |
|---|:-:|:-:|
| `snippets/agents/custom-agent` | pass | **pass** |
| `snippets/agents/llm-agent` | **FAIL** — `TS2322` (zod 3 vs 4) | **pass** |
| `snippets/get-started/multi_tool_agent` | **FAIL** — `TS2322` (zod 3 vs 4) | **pass** |
| `snippets/sessions` | pass | **pass** |
| `snippets/skills` | **FAIL** — `TS2305`, missing exports | **pass** |

The underlying cause of most failures is an **undeclared `zod` dependency**: the examples `import` from `zod` without declaring it, so a hoisted zod 4 met ADK 0.2.x's zod 3. `@google/adk@1.5.0` moved to zod ^4, so the bump strictly improves or leaves unchanged every one of them.

Split out of the TypeScript-naming PR on purpose — a reviewer approving a casing change should not have to evaluate a dependency upgrade.

## Four examples are deliberately left alone

`snippets/tools/overview`, `snippets/tools/function-tools`, `snippets/callbacks`, and `snippets/agents/workflow-agents` fail type-check at **both** their pinned version and 1.5.0. Bumping them would fix nothing and would put unverified entries in a branch that claims everything was type-checked. Their failures are tsconfig and sample-source bugs, not API drift.

Two of those are worth reporting separately:

- **`snippets/agents/workflow-agents` has `include: ["src/**/*"]` and no `src/` directory** — it has never actually been type-checked by anything.
- **`snippets/sessions` and `snippets/skills` ship no `tsconfig.json` at all**, so a bare `npx tsc --noEmit` in those directories checks nothing and exits 1 printing help. The results above were obtained with explicit compiler flags.

More broadly: `.github/workflows/` runs `go-snippets-pr-check` and `kotlin-snippets-pr-check` on every PR, so Go and Kotlin snippets are compiler-verified. **There is no equivalent for TypeScript**, which is why these drifted unnoticed. That is the durable fix; this PR is the immediate one.


joefernandez (approved):
Thanks for the update!

PR #2081, opened 2026-08-05
Revamping of MCP Tools page

zyantw:
Staged: https://deploy-preview-2081--adk-docs-preview.netlify.app/tools-custom/mcp-tools/

## PR Summary & Consolidation Mapping

This PR refactors and expands the Model Context Protocol (MCP) documentation in ADK. Previously, all MCP documentation was consolidated into a single monolithic 1,300-line page (`docs/tools-custom/mcp-tools.md`), mixing introductory tutorials, multi-language snippets, deployment guides, and advanced runtime configurations. 

This PR modularizes the content into a streamlined 5-page documentation suite organized by developer use case, adds comprehensive multi-language support (Python, TypeScript, Go, Java, and Kotlin), and incorporates technical patterns validated against all five language SDK repositories.

---

## Comparison: Original Page (`main`) vs. PR Changes (`HEAD`)

| Dimension | Original Page (`main: mcp-tools.md`) | PR Changes (`HEAD`: 5 Modular Pages) |
|---|---|---|
| **Structure & Length** | Single monolithic file (~1,300 lines) covering basics, deployment, server creation, and troubleshooting in one place. | Modular 5-page guide (~1,450 lines total), cleanly separated by developer journey and architecture. |
| **Core Guide (`mcp-tools.md`)** | Mixed introductory setup with container configuration, GKE manifests, and troubleshooting scripts. | Streamlined into a quickstart focused on client concepts, architecture sequence diagrams, and local/remote quickstarts across 5 languages. |
| **Advanced Production Features** | Limited mention of basic progress callbacks and headers. | **New Page (`mcp-tools-advanced.md`)**: Dedicated guide for dynamic per-user `header_provider`, `require_confirmation` human-in-the-loop approvals, session-aware progress factories, standalone `Runner` lifecycle teardown, `tool_name_prefix` namespacing, bi-directional sampling/elicitation, and interactive UI widget rendering. |
| **Serving ADK over MCP** | Basic custom script example buried in the middle of the client tutorial. | **New Page (`agent-as-mcp-server.md`)**: Complete guide covering one-line agent publishing via `to_mcp_server(agent)`, low-level tool conversion via `adk_to_mcp_tool_type`, stdio server lifecycle, and client testing. |
| **Hierarchical Multi-Agent Delegation** | Not covered. | **New Page (`agent-managed-mcp.md`)**: Architectural guide on encapsulating MCP toolsets inside sub-agents using `AgentTool`, applying cognitive `tool_filter`, eliminating parent context window bloat, and enabling model tiering. |
| **Production Deployment** | Scattered sections on Dockerfiles and raw GKE YAML snippets. | **New Page (`mcp-deployment.md`)**: Dedicated deployment reference emphasizing synchronous agent definition rules, Cloud Run Streamable HTTP ASGI microservices, GKE sidecars, and Agent Runtime commands. |
| **Multi-Language Parity** | Partial Python and Java snippets; missing Kotlin. | Complete, verified code samples across **Python**, **TypeScript**, **Go**, **Java**, and **Kotlin**. |

---
### 1. Go SDK Transport & Tool Filtering
- **Doc File:** `docs/tools-custom/mcp-tools.md`
- **Claim:** In ADK Go, MCP integration uses `github.com/modelcontextprotocol/go-sdk/mcp` with `mcptoolset.New` accepting `mcp.CommandTransport` and `tool.AllowedToolsPredicate`.
- **Repository Proof:**
  - **Dependency Declaration:** [`google/adk-go/go.mod:18`](https://github.com/google/adk-go/blob/a61147276167f6b1b636d863f23a030c0a6babdf/go.mod#L18)
    ```go
    require (
    	github.com/google/go-cmp v0.7.0
    	github.com/modelcontextprotocol/go-sdk v1.1.0
    	google.golang.org/genai v0.9.0
    )
    ```
  - **Constructor Signature:** [`google/adk-go/tool/mcptoolset/set.go:44-58`](https://github.com/google/adk-go/blob/a61147276167f6b1b636d863f23a030c0a6babdf/tool/mcptoolset/set.go#L44-L58)
    ```go
    func New(cfg Config) (tool.Toolset, error) {
    	transport, err := buildTransport(cfg)
    	if err != nil {
    		return nil, err
    	}
    	return &set{
    		mcpClient:                   newConnectionRefresher(cfg.Client, transport),
    		toolFilter:                  cfg.ToolFilter,
    		requireConfirmation:         cfg.RequireConfirmation,
    		requireConfirmationProvider: cfg.RequireConfirmationProvider,
    	}, nil
    }
    ```

---

### 2. TypeScript Remote Connection Headers Configuration
- **Doc File:** `docs/tools-custom/mcp-tools.md`
- **Claim:** In ADK TypeScript (`@google/adk`), custom headers for remote `StreamableHTTPConnectionParams` must be passed via `transportOptions.requestInit.headers`.
- **Repository Proof:**
  - **Interface Definition:** [`google/adk-js/core/src/tools/mcp/mcp_session_manager.ts:53-66`](https://github.com/google/adk-js/blob/8f792bbb98e8f360085b69a0dbafb3c1a071a4a4/core/src/tools/mcp/mcp_session_manager.ts#L53-L66)
    ```typescript
    export interface StreamableHTTPConnectionParams {
      type: 'StreamableHTTPConnectionParams';
      url: string;
      /**
       * @deprecated
       * Use transportOptions.requestInit.headers instead.
       */
      header?: Record<string, unknown>;
      timeout?: number;
      sseReadTimeout?: number;
      terminateOnClose?: boolean;
      transportOptions?: StreamableHTTPClientTransportOptions;
    }
    ```

---

### 3. Sub-Agent Wrapping & Description Delegation (`AgentTool`)
- **Doc File:** `docs/tools-custom/agent-managed-mcp.md`
- **Claim:** In ADK Python, `AgentTool` wraps an `LlmAgent` as a callable tool, automatically deriving its `name` and `description` from the child agent without accepting a direct `description` keyword argument.
- **Repository Proof:**
  - **Constructor Definition:** [`google/adk-python/src/google/adk/tools/agent_tool.py:137-152`](https://github.com/google/adk-python/blob/44e0b2a8b1215aa98f057c4a781ddc24bae220da/src/google/adk/tools/agent_tool.py#L137-L152)
    ```python
      def __init__(
          self,
          agent: BaseAgent,
          skip_summarization: bool = False,
          *,
          include_plugins: bool = True,
          propagate_grounding_metadata: bool = False,
      ):
        self.agent = agent
        self.skip_summarization: bool = skip_summarization
        self.include_plugins = include_plugins
        self.propagate_grounding_metadata = propagate_grounding_metadata

        super().__init__(name=agent.name, description=agent.description)
    ```

---

### 4. Human-In-The-Loop Confirmation Argument Introspection
- **Doc File:** `docs/tools-custom/mcp-tools-advanced.md`
- **Claim:** ADK Python dynamically introspects `require_confirmation` callable parameters against incoming tool arguments, requiring parameter names (such as `query`) or `**kwargs`.
- **Repository Proof:**
  - **Argument Preparation Logic:** [`google/adk-python/src/google/adk/tools/mcp_tool/mcp_tool.py:427-440`](https://github.com/google/adk-python/blob/44e0b2a8b1215aa98f057c4a781ddc24bae220da/src/google/adk/tools/mcp_tool/mcp_tool.py#L427-L440)
    ```python
      def _prepare_callable_args(
          self,
          target: Callable[..., Any],
          args: dict[str, Any],
          tool_context: ToolContext,
      ) -> dict[str, Any]:
        """Prepares arguments for invoking a user-provided callable."""
        args_to_call = args.copy()
        try:
          signature = inspect.signature(target)
        except (ValueError, TypeError):
          return args_to_call

        valid_params = set(signature.parameters.keys())
        has_kwargs = any(
            param.kind == inspect.Parameter.VAR_KEYWORD
            for param in signature.parameters.values()
        )
    ```

---

### 5. One-Line Agent Conversion to MCP Server (`to_mcp_server`)
- **Doc File:** `docs/tools-custom/agent-as-mcp-server.md`
- **Claim:** ADK Python provides `to_mcp_server(agent)` to publish an entire `LlmAgent` as a standard FastMCP server over stdio.
- **Repository Proof:**
  - **Function Declaration:** [`google/adk-python/src/google/adk/tools/mcp_tool/_agent_to_mcp.py:34-45`](https://github.com/google/adk-python/blob/44e0b2a8b1215aa98f057c4a781ddc24bae220da/src/google/adk/tools/mcp_tool/_agent_to_mcp.py#L34-L45)
    ```python
    def to_mcp_server(
        agent: BaseAgent,
        runner: Runner | None = None,
        name: str | None = None,
        **fastmcp_kwargs: Any,
    ) -> FastMCP:
      """Converts an ADK agent into a FastMCP server."""
    ```

---

### 6. Java Streamable HTTP Remote Parameters
- **Doc File:** `docs/tools-custom/mcp-deployment.md`
- **Claim:** ADK Java uses `StreamableHttpServerParameters.builder().url(...).headers(...).build()` to configure remote Streamable HTTP MCP connections.
- **Repository Proof:**
  - **Class Definition:** [`google/adk-java/core/src/main/java/com/google/adk/tools/mcp/StreamableHttpServerParameters.java:33-40`](https://github.com/google/adk-java/blob/4b058cd17346944ac047828eda6424e495204bf9/core/src/main/java/com/google/adk/tools/mcp/StreamableHttpServerParameters.java#L33-L40)
    ```java
    public class StreamableHttpServerParameters {
      private final String url;
      private final Map<String, String> headers;
      private final Duration timeout;
    ```

---

### 7. Kotlin Config-Driven Toolset & Header Provider
- **Doc File:** `docs/tools-custom/mcp-deployment.md`
- **Claim:** ADK Kotlin creates `McpToolset` via `McpToolset.McpToolsetConfig(streamableHttpConnectionParams = ...).toToolset(headerProvider = ...)`.
- **Repository Proof:**
  - **Factory Method:** [`google/adk-kotlin/core/src/jvmMain/kotlin/com/google/adk/kt/tools/mcp/McpToolset.kt:406-425`](https://github.com/google/adk-kotlin/blob/0e79ca73873a270ccfae90b7a82aecc52ce3cecc/core/src/jvmMain/kotlin/com/google/adk/kt/tools/mcp/McpToolset.kt#L406-L425)
    ```kotlin
        @JvmOverloads
        fun toToolset(
          headerProvider: (suspend (ReadonlyContext) -> Map<String, String>)? = null,
          progressConsumers: List<(McpSchema.ProgressNotification) -> Unit> = emptyList(),
        ): McpToolset {
          val params =
            listOfNotNull(stdioConnectionParams, sseConnectionParams, streamableHttpConnectionParams)

          require(params.size == 1) {
            "Exactly one of stdioConnectionParams, sseConnectionParams or streamableHttpConnectionParams must be set"
          }
    ```

---

## Conclusion

All critical claims and code implementations across the 5 files in PR #2081 directly match the API definitions and conventions of their respective language repositories. The documentation PR is **PASS**.


docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+    ]
+  return [
+      mcp_types.TextContent(type="text", text=f"Unknown tool: '{name}'")
+  ]
+
+```
+
+## Further resources
+
+Once you understand the basics, explore [Advanced use cases](/advanced-mcp-tools) for complex implementations and custom integrations.
```
zyantw, line 659: I can update this section once the page for PR #2122 is approved. @joefernandez

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
+# Model Context Protocol Tools
+
+<div class="language-support-tag">
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v0.3.10</span><span class="lst-typescript">Typescript v0.2.0</span><span class="lst-go">Go v0.1.0</span><span class="lst-java">Java v0.1.0</span>
+</div>
+
+## What is Model Context Protocol (MCP)?
```
joefernandez: remove. don't stack headers. The paragraph below is the intro

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+
+The **Model Context Protocol (MCP)** is an open standard for connecting Large Language Models (LLMs) to external data sources, tools, and systems. Think of it as a universal connection mechanism that simplifies how LLMs obtain context, execute actions, and interact with various systems.
```
joefernandez: LLMs --> generative AI models

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+
+---
+
+## Core architecture and concepts
+
+MCP follows a client-server architecture, defining how data or resources, interactive templates or prompts, and actionable functions or tools are exposed by an MCP server and consumed by an MCP client, which could be an LLM host application or an AI agent.
```
joefernandez: Add a sentence explaining that you use the McpToolset class as an interface between MCP Servers and ADK agents. Also mention that it's possible to configure an ADK server as an MCP server for use by other client systems.

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+    Toolset->>Server: Execute Tool (call_tool via Stdio/HTTP)
+    Server-->>Toolset: Execution Result (Text/JSON)
+    Toolset-->>Agent: Result returned to LLM
+```
+
+## Prerequisites and setup rules
+
+Before you begin, ensure you have the following set up:
+
+### Checklist
```
joefernandez: remove. you don't need this header

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+
+    Agents deployed to production **must define `McpToolset` synchronously** in `agent.py`. Dynamic asynchronous agent initialization is only supported for local debugging or custom standalone runners.
+
+---
+
+## Key considerations
+
+ When you start building with the Model Context Protocol (MCP) and ADK, these key architectural differences will help you design more stable and efficient agents.
+
+### Compare tools
```
joefernandez: remove; use the prior header to introduce this topic; you don't need a second one

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+    Get-Command npx
+    ```
+    
+!!! warning "Deployment rule"
+
+    Agents deployed to production **must define `McpToolset` synchronously** in `agent.py`. Dynamic asynchronous agent initialization is only supported for local debugging or custom standalone runners.
+
+---
+
+## Key considerations
```
joefernandez: there is a choice to be made, here, so be more clear about that, heading:

```
## MCP Implementation options
```

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+
+ When you start building with the Model Context Protocol (MCP) and ADK, these key architectural differences will help you design more stable and efficient agents.
```
joefernandez: Add a sentence to introduce the table that follows

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+| **Architecture** | External server process or remote service providing deterministic endpoints adapted into the primary `LlmAgent` tool list. | In-process, hierarchical agent encapsulation where a parent agent invokes a child `LlmAgent` as a callable tool. | An autonomous ADK agent compiled into an MCP server, callable by external clients (Claude Code, IDEs, external hosts). |
+| **Context Window Impact** | **High Context Bloat**: Every tool definition and raw output, for example: database rows or file blobs, enters the primary agent's history. | **Zero Context Bloat**: Intermediate exploratory reasoning, failed tool calls, and large raw outputs remain isolated in the sub-agent loop. | **Isolated**: The external caller only receives the final aggregated response text/blocks. |
+| **Cognitive Load & Model Tiering** | Single model must understand all tool schemas, validation constraints, and workflow state simultaneously. | Enables **model tiering**, for example: `gemini-2.5-pro` for orchestrator, and `gemini-2.5-flash` for sub-agent tool execution with dedicated system instructions. | Independent model reasoning dedicated solely to the wrapped task. |
```
joefernandez: don't call model operations "cognitive" (humanizing AI), also avoid using "&" in prose

Cognitive Load & Model Tiering. --> AI Model Load and Tiering

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+| **Latency & Token Cost** | **Lower Cost & Predictable Latency**: 1 LLM turn + 1 deterministic tool invocation + 1 response generation turn. | **Higher Cost & Variable Latency**: Multiplies LLM calls, sub-agent reasoning turns before returning to parent. | Client-driven; latency depends on internal agent execution depth. |
```
joefernandez: Multiplies --> Multiple

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+| **Ideal Use Cases** | <ul><li>Deterministic API integrations: Postgres, BigQuery, GitHub, Google Maps.</li><li>File system operations & static resource reading.</li><li>Reusing standard pre-built community MCP servers.</li></ul> | <ul><li>Multi-step autonomous workflows requiring trial-and-error. For example: code debugging or research synthesis.</li><li>Tasks needing isolated personas or specialized instructions.</li><li>Scenarios with >20 tools where schema overload harms accuracy.</li></ul> | <ul><li>Exposing complex ADK multi-agent capabilities to external MCP-compliant ecosystems.</li><li>Integrating ADK agents into IDEs, editors, or A2A pipelines.</li></ul> |
+
+!!! note "State restoration"
+
+    While your agent preserves its session state during lifecycle events, it **does not** automatically re-establish active MCP connections upon restoration. It will re-initialize the connection as needed.
```
joefernandez: 1) rephrase:
While your agent preserves its session state --> While ADK agents preserve session state during lifecycle events, they **do not** ...

2) clarify subject and avoid future tense:
It will re-initialize the connection as needed. --> Agents re-initialize connections as needed.

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+
+## Universal Setup Rules
+
+  1. **Absolute Paths**: File system MCP servers require absolute path arguments: `os.path.abspath(...)`. Relative paths cause runtime resolution errors in subprocesses.
+  2. **Package Discovery**: When running `adk web`, ensure an `__init__.py` file exists inside your agent folder so ADK can import the package.
+  3. **Windows Async Bug**: If you encounter `_make_subprocess_transport NotImplementedError` on Windows, start `adk web` with the `--no-reload` flag.
```
joefernandez: #3 is not a universal setup rule. It's a tip for Windows users only. find a place to list this separately

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+
+## Understand uses and integrations
+
+There are two main integration patterns:
+
+  1. **Use existing MCP Servers within ADK**: When an ADK agent acts as an MCP client.
+  1. **Expose ADK Tools via an MCP Server**: When you build an MCP server that wraps ADK Tools to make them accessible to any MCP Client.
+
+### Use existing MCP Servers within ADK
```
joefernandez: This should be a level 2 heading: you are introducing general implementation approach that lines up with the implementation options table above.

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+
+    While your agent preserves its session state during lifecycle events, it **does not** automatically re-establish active MCP connections upon restoration. It will re-initialize the connection as needed.
+
+## Universal Setup Rules
+
+  1. **Absolute Paths**: File system MCP servers require absolute path arguments: `os.path.abspath(...)`. Relative paths cause runtime resolution errors in subprocesses.
+  2. **Package Discovery**: When running `adk web`, ensure an `__init__.py` file exists inside your agent folder so ADK can import the package.
+  3. **Windows Async Bug**: If you encounter `_make_subprocess_transport NotImplementedError` on Windows, start `adk web` with the `--no-reload` flag.
+
+## Understand uses and integrations
```
joefernandez: This section seems like it belongs with the previous section that lists the implementation options and considerations. It doesn't belong in a level 2 heading on it's own

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+                timeout: 5,
+                sseReadTimeout: 300
+            }),
+        ],
+    });
+    ```
+    
+---
+
+## Expose ADK Tools via a Custom MCP Server
```
joefernandez: remove "Custom":

Expose ADK Tools with MCP Server

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+
+---
+
+## Key considerations
+
+ When you start building with the Model Context Protocol (MCP) and ADK, these key architectural differences will help you design more stable and efficient agents.
+
+### Compare tools
+
+| Dimension | **Direct MCP Tool Integration** (`McpToolset`) | **Specialized Sub-Agent Delegation** (`AgentTool`) | **Agent-Exposed MCP Server** (`to_mcp_server`) |
```
joefernandez, line 68: 1. The names of these implementation options need to be carried through to the headings of later sections. Right now, it's hard to figure our what implementation guides match to these options. 
2. Recommend linking these implementation options to sections of the guide.

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+
+!!! note "State restoration"
+
+    While your agent preserves its session state during lifecycle events, it **does not** automatically re-establish active MCP connections upon restoration. It will re-initialize the connection as needed.
+
+## Universal Setup Rules
+
+  1. **Absolute Paths**: File system MCP servers require absolute path arguments: `os.path.abspath(...)`. Relative paths cause runtime resolution errors in subprocesses.
+  2. **Package Discovery**: When running `adk web`, ensure an `__init__.py` file exists inside your agent folder so ADK can import the package.
```
joefernandez, line 83: not clear what this has to do with MCP implementations

docs/tools-custom/mcp-tools-2.md
```diff
@@ -0,0 +1,666 @@
...
+| **Ideal Use Cases** | <ul><li>Deterministic API integrations: Postgres, BigQuery, GitHub, Google Maps.</li><li>File system operations & static resource reading.</li><li>Reusing standard pre-built community MCP servers.</li></ul> | <ul><li>Multi-step autonomous workflows requiring trial-and-error. For example: code debugging or research synthesis.</li><li>Tasks needing isolated personas or specialized instructions.</li><li>Scenarios with >20 tools where schema overload harms accuracy.</li></ul> | <ul><li>Exposing complex ADK multi-agent capabilities to external MCP-compliant ecosystems.</li><li>Integrating ADK agents into IDEs, editors, or A2A pipelines.</li></ul> |
+
+!!! note "State restoration"
+
+    While your agent preserves its session state during lifecycle events, it **does not** automatically re-establish active MCP connections upon restoration. It will re-initialize the connection as needed.
+
+## Universal Setup Rules
```
joefernandez, line 80: this section seems out of place. Is this actually applicable to all MCP implementations?

docs/tools-custom/mcp-tools-2.md
joefernandez: @zyantw - to publish this properly, you should combine the changes of this PR and PR 2122 so that we don't have a temporary loss of content when publishing this cut-down version

docs/tools-custom/mcp-tools-2.md
joefernandez, line 1: after addressing the comments provided, and getting the right structure, let's overwrite the existing mcp-tools.md page with this updated content instead of having a separate page

docs/tools-custom/mcp-tools-2.md
zyantw: Thanks for this

docs/tools-custom/mcp-tools-2.md
zyantw: @joefernandez changes combined

docs/tools-custom/mcp-tools/mcp-tools-advanced.md
joefernandez, line 1: remember to modify mkdocs.yaml to add this to the navigation

docs/tools-custom/mcp-tools.md
```diff
@@ -1,691 +1,442 @@
 # Model Context Protocol Tools
 
 <div class="language-support-tag">
-  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v0.1.0</span><span class="lst-typescript">Typescript v0.2.0</span><span class="lst-go">Go v0.1.0</span><span class="lst-java">Java v0.1.0</span>
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v0.3.10</span><span class="lst-typescript">Typescript v0.2.0</span><span class="lst-go">Go v0.1.0</span><span class="lst-java">Java v0.1.0</span>
```
joefernandez: revert: 
python v0.3.10 -- > v0.1.0

mcp_tool.py added in v0.1.0 :
https://github.com/google/adk-python/commit/98278201436cc7f642ede1b22407783e8c155b5e#diff-405e85c475912c1b8df589d426b6d7ffa38bea22ec696101fe92bda2c6c64c61

docs/tools-custom/mcp-tools.md
```diff
@@ -694,585 +445,238 @@ root_agent = LlmAgent(
...
-  async for event in events_async:
-    print(f"Event received: {event}")
+=== "Python"
 
-  # Cleanup is handled automatically by the agent framework
-  # But you can also manually close if needed:
-  print("Closing MCP server connection...")
-  await toolset.close()
-  print("Cleanup complete.")
+  ```python
```
joefernandez: tabbed interface isn't rendering properly here, because this line isn't indented 4 spaces

docs/tools-custom/mcp-tools.md
```diff
@@ -694,585 +445,238 @@ root_agent = LlmAgent(
...
+            print(f"Content of {resource_name}:", content_blocks)
+  ```
 
-For long-running tools, `McpToolset` supports a `progress_callback`. This approach allows you to receive real-time updates from the MCP server. You can provide a simple callback function or a factory that creates callbacks with access to the runtime context, such as updating session state.
+=== "TypeScript"
 
-```python
-async def my_progress_callback(progress: float, total: float, message: str):
-    print(f"Progress: {progress}/{total} - {message}")
+  ```typescript
```
joefernandez: tabbed interface isn't rendering properly here, because this line isn't indented 4 spaces

docs/tools-custom/mcp-tools.md
```diff
@@ -1,691 +1,442 @@
...
+    While ADK agents preserve session state during lifecycle events, they do not automatically re-establish active MCP connections upon restoration. Agents re-initialize connections as needed.
 
-            System.out.println("Agent created: " + agent.name());
+## Understand uses and integrations
 
-            InMemoryRunner runner = new InMemoryRunner(agent);
-            String userId = "user123";
-            String sessionId = "1234";
-            String promptText = "Which files are in this directory - " + yourFolderPath + "?";
+There are three main integration patterns:
```
joefernandez: reuse this content by making it a second paragraph of the introduction. Explain that the direct integration is covered in this page and other implementations are covered in the advanced guide

docs/tools-custom/mcp-tools.md
```diff
@@ -1,691 +1,442 @@
...
+1. **Direct MCP Tool Integration**: When an ADK agent acts as an MCP client using `McpToolset`.
+2. **Agent-Exposed MCP Server**: When you build an MCP server that wraps ADK Tools using `to_mcp_server`.
+3. **Specialized Sub-Agent Delegation**: When an agent delegates to a sub-agent using `AgentTool`.
+   
 
-            // Explicitly create the session first
-            SessionKey sessionKey = runner.sessionService().createSession(runner.appName(), userId, null, sessionId).blockingGet().sessionKey();
-            System.out.println("Session created: " + sessionId + " for user: " + userId);
+### Direct MCP Tool Integration (McpToolset)
```
joefernandez: bump this section up to H2 and call it:

`## Get started`

docs/tools-custom/mcp-tools.md
```diff
@@ -1,691 +1,442 @@
...
 
-3.  **Interact in the UI**:
-    *   Select the `travel_planner_agent`.
-    *   Try prompts like:
-        *   "I will be in San Francisco tomorrow. What’s the weather like?"
-        *   "Find coffee shops near Golden Gate Park."
-        *   "Get directions from GooglePlex to SFO."
-
-You should see the agent use the Google Maps Grounding Lite MCP tools to provide directions or location-based information.
+### Agent-Exposed MCP Server (to_mcp_server)
```
joefernandez: move this section (and all it's sub-sections) to the advanced guide

docs/tools-custom/mcp-tools.md
```diff
@@ -1,691 +1,442 @@
...
+     ```
 
-1.  Instantiate the ADK tool(s) you want to expose (e.g., `FunctionTool(load_web_page)`).
-2.  Implement the MCP server's `@app.list_tools()` handler to advertise the ADK tool(s). This involves converting the ADK tool definition to the MCP schema using the `adk_to_mcp_tool_type` utility from `google.adk.tools.mcp_tool.conversion_utils`.
-3.  Implement the MCP server's `@app.call_tool()` handler. This handler will:
-    *   Receive tool call requests from MCP clients.
-    *   Identify if the request targets one of your wrapped ADK tools.
-    *   Execute the ADK tool's `.run_async()` method.
-    *   Format the ADK tool's result into an MCP-compliant response (e.g., `mcp.types.TextContent`).
+### Specialized Sub-Agent Delegation (AgentTool)
```
joefernandez: move this section (and all it's sub-sections) to the advanced guide

docs/tools-custom/mcp-tools.md
```diff
@@ -1,691 +1,442 @@
...
+#### Implementation Overview
+To implement this pattern, you combine `McpToolset` and `AgentTool`:
+1. Initialize your MCP connection (e.g., HTTP or Stdio).
+2. Create a specific `LlmAgent` (the sub-agent) and attach the `McpToolset` to it.
+3. Wrap that sub-agent in an `AgentTool`.
+4. Provide the `AgentTool` to your root agent.
+     
 
-Add the following code to `my_adk_mcp_server.py`. This script sets up an MCP server that exposes the ADK `load_web_page` tool.
+### Build the MCP server with ADK tools
```
joefernandez: this looks like it's at the wrong heading level (should be a H4)

docs/tools-custom/mcp-tools.md
```diff
@@ -1,691 +1,442 @@
...
         print("\nMCP Server (stdio) stopped by user.")
-    except Exception as e:
-        print(f"MCP Server (stdio) encountered an error: {e}")
-    finally:
-        print("MCP Server (stdio) process exiting.")
-# --- End MCP Server ---
 ```
 
-### Step 3: Test your Custom MCP Server with an ADK Agent
+### Test your custom MCP server with an ADK Agent
```
joefernandez: this looks like it's at the wrong heading level (should be H4)

docs/tools-custom/mcp-tools-advanced.md
```diff
@@ -0,0 +1,394 @@
...
+
+<div class="language-support-tag">
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v0.1.0</span><span class="lst-typescript">Typescript v0.2.0</span><span class="lst-go">Go v0.1.0</span><span class="lst-java">Java v0.1.0</span><span class="lst-kotlin">Kotlin v0.1.0</span>
+</div>
+
+This guide covers advanced integration patterns for the Model Context Protocol (MCP) in ADK. It provides production patterns for dynamic per-user authentication, human-in-the-loop approvals, long-running progress tracking, custom runtime execution, and enterprise cloud deployments.
+
+---
+
+## Choose your mechanism
```
joefernandez: What am I chosing, here? rework the header to indicate the choice and/or development goal

docs/tools-custom/mcp-tools/mcp-tools-advanced.md
```diff
@@ -0,0 +1,394 @@
...
+| **Stream real-time progress for long operations** | Progress Callback & Factory | `progress_callback=...` | Heavy SQL queries, web scraping, data indexing |
+| **Run agents in FastAPI / backend services without `adk web`** | Programmatic Runner Lifecycle | `Runner` + `await toolset.close()` | Custom microservices, CLI tools, worker queues |
+| **Deploy containerized MCP agents to Cloud Run or GKE** | Stateless Streamable HTTP / Sidecar | `StreamableHTTPConnectionParams` | Horizontally scalable serverless or cluster pods |
+| **Resolve tool naming collisions across servers** | Tool Namespacing & Filtering | `tool_name_prefix`, `tool_filter` | Aggregating multiple MCP servers (DB + GitHub) |
+| **Handle server-requested sampling or auth challenges** | Bi-directional Protocol Callbacks | `sampling_callback`, `elicitation_callback` | Server-initiated LLM generation & auth prompts |
+| **Inspect raw STDERR diagnostic streams** | Diagnostic Stream Logging | `errlog=sys.stderr` | Troubleshooting MCP subprocess crashes |
+
+---
+
+## Dynamic authentication and per-user headers (`header_provider`)
```
joefernandez, line 26: rework this heading to indicate the development problem being solved by this implementation

docs/tools-custom/mcp-tools.md
```diff
@@ -694,585 +445,238 @@ root_agent = LlmAgent(
...
-#### Pattern 1: Self-Contained Stdio MCP Servers
-
-For MCP servers that can be packaged as npm packages or Python modules (like `@modelcontextprotocol/server-filesystem`), you can include them directly in your agent container:
+## UI Rendering 
 
-**Container Requirements:**
-```dockerfile
-# Example for npm-based MCP servers
-FROM python:3.13-slim
+Standard MCP tools return plain text or JSON output. **Experimental UI Rendering** enables MCP tools to return rich, interactive visual widgets, such as maps, charts, or forms, directly inside the chat interface.
```
joefernandez: is this still experimental? we should check the code and/or with engineering

docs/tools-custom/mcp-tools.md
```diff
@@ -694,585 +445,238 @@ root_agent = LlmAgent(
...
-  --service_name=<your-service-name> \
-  ./path/to/your/agent_directory
 ```
 
-### Deployment Patterns
-
-#### Pattern 1: Self-Contained Stdio MCP Servers
-
-For MCP servers that can be packaged as npm packages or Python modules (like `@modelcontextprotocol/server-filesystem`), you can include them directly in your agent container:
+## UI Rendering
```
joefernandez: this seems like an advanced topic (move to advanced page?)

sesaymohsesay11-design (approved):

sesaymohsesay11-design (approved):

sesaymohsesay11-design (approved):

docs/tools-custom/mcp-tools/mcp-tools-advanced.md
zyantw: done!

mkdocs.yml
joefernandez: Remove this "MCP" listing from the navigation

joefernandez: remove the MCP page file (delete it) and add a redirect from this location to the main mcp-tools page

docs/tools-custom/mcp-tools-index.md
joefernandez, line 1: We don't need a completely separate page to explain MCP, couple of things:

1. Delete this page
2. Also delete the /docs/mcp/index.md file 
2. Redirect the (/mcp/index.md) location to the main page for mcp-tools

docs/tools-custom/mcp-tools/index.md
joefernandez, line 1: The file location for this page should be
`docs/tools-custom/mcp-tools/index.md`

use `git mv` to move the file while preserving the change history

docs/tools-custom/mcp-tools-advanced.md
```diff
@@ -0,0 +1,361 @@
+# Advanced MCP configurations and production guide
```
joefernandez:
```suggestion
# Advanced MCP configuration
```

docs/tools-custom/mcp-tools-advanced.md
```diff
@@ -0,0 +1,361 @@
...
+
+<div class="language-support-tag">
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v0.1.0</span>
+</div>
+
+This guide covers advanced integration patterns for the Model Context Protocol (MCP) in ADK. It provides production patterns for dynamic per-user authentication, human-in-the-loop approvals, long-running progress tracking, custom runtime execution, and enterprise cloud deployments.
+
+---
```
joefernandez:
```suggestion
```

docs/tools-custom/mcp-tools-advanced.md
```diff
@@ -0,0 +1,361 @@
...
+
+## Select a configuration pattern for your use case
```
joefernandez:
```suggestion
## Advanced configuration patterns
```

docs/tools-custom/agent-as-mcp-server.md
```diff
@@ -0,0 +1,188 @@
+# Agent-Exposed MCP Server (`to_mcp_server`)
```
joefernandez:
```suggestion
# Configure ADK agents as MCP servers
```

docs/tools-custom/agent-as-mcp-server.md
```diff
@@ -0,0 +1,188 @@
...
+ADK constructs. There are two primary ways to achieve this:
+
+* **Expose an entire Agent:** Wraps the full multi-turn agent reasoning and 
+  internal tool execution into a server using a simple one-line conversion.
+
+* **Expose individual Tools:** Involves manually building a lightweight MCP 
+  server to wrap specific, standalone ADK tools (like a `FunctionTool`) without 
+  the agent's reasoning loop.
+
+## Expose an Entire ADK Agent
```
joefernandez: This section is nice quick implementation with code, so follow the "Get started" pattern:

```suggestion
## Get started
```

docs/tools-custom/agent-as-mcp-server.md
```diff
@@ -0,0 +1,188 @@
+# Agent-Exposed MCP Server (`to_mcp_server`)
+
+You can make your ADK capabilities accessible to external MCP clients (such as
```
joefernandez:
```suggestion
You can make your ADK agent's capabilities accessible to external MCP clients (such as 
```

docs/tools-custom/agent-as-mcp-server.md
```diff
@@ -0,0 +1,188 @@
+# Agent-Exposed MCP Server (`to_mcp_server`)
+
+You can make your ADK capabilities accessible to external MCP clients (such as 
+Claude Desktop, IDEs, or custom hosts) by building an MCP server that exposes
```
joefernandez:
```suggestion
Antigravity, Claude Code, or custom agents) by hosting your ADK agent's 
```

docs/tools-custom/agent-as-mcp-server.md
```diff
@@ -0,0 +1,188 @@
+# Agent-Exposed MCP Server (`to_mcp_server`)
+
+You can make your ADK capabilities accessible to external MCP clients (such as 
+Claude Desktop, IDEs, or custom hosts) by building an MCP server that exposes 
+ADK constructs. There are two primary ways to achieve this:
```
joefernandez:
```suggestion
capabilities within an MCP server. There are two primary ways to achieve this:
```

docs/tools-custom/agent-as-mcp-server.md
zyantw: Thank you for this!

docs/tools-custom/mcp-tools/agent-managed-mcp.md
```diff
@@ -0,0 +1,84 @@
...
+which connects to the underlying MCP server (via `list_tools`), converts the 
+external MCP tool definitions into ADK-compatible `BaseTool` instances, and 
+proxies all execution calls (`call_tool`) asynchronously.
+
+## Tool scoping and cognitive filtering (`tool_filter`)
+When assigning an `McpToolset` to a specialized sub-agent, you can restrict the 
+exact tools made available to that agent using the `tool_filter` parameter:
+* **Cognitive Focus:** Exposes only the specific actions relevant to the sub-agent's 
+  mandate, such as, 'read_file', 'list_directory'.
+* **Security and sandboxing:** Prevents unintended access to dangerous tools
```
joefernandez: `tool_filter` is not a sandbox. It only controls which tool schemas are advertised to the model ([`mcp_toolset.py:511`](https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/mcp_toolset.py#L511)) — the MCP server still runs with its own privileges, nothing here isolates execution, and it does nothing about untrusted content coming back *from* the server.

Suggest replacing lines 59-61 with:

```markdown
* **Reduced attack surface:** Limits which server-side tools the model can call
  at all, so an overly permissive MCP server cannot expose destructive
  operations to this sub-agent. This is not a sandbox: the MCP server still
  runs with its own privileges, and its output is still untrusted input.
```

docs/tools-custom/mcp-tools/deployment.md
```diff
@@ -0,0 +1,334 @@
...
+*   Monitor memory usage closely for stdio MCP servers, as each connection spawns a new process.
+*   Configure appropriate CPU and memory limits for MCP server processes.
+*   Implement connection pooling for remote MCP servers to optimize resource consumption.
+
+### Security
+!!! important "Security Best Practices"
+    *   Use strict authentication headers for all remote MCP connections.
+    *   Restrict network access tightly between ADK agents and MCP servers.
+    *   Filter MCP tools using `tool_filter` to strictly limit exposed functionality.
+    *   Validate all MCP tool inputs to prevent prompt or command injection attacks.
```
joefernandez: Input validation does not prevent prompt injection — prompt injection is an open problem. This is also stronger than the wording it replaces on the deleted page ("Validate MCP tool inputs to prevent injection attacks").

```suggestion
    *   Validate MCP tool inputs and treat all MCP tool output as untrusted content; input validation reduces, but does not eliminate, command- and prompt-injection risk.
```

docs/tools-custom/mcp-tools/mcp-tools-advanced.md
```diff
@@ -0,0 +1,357 @@
...
+MCP servers can expose high-impact capabilities, for example: database schema modifications or record deletions. You can enforce confirmation globally across all tools in the toolset or conditionally via a predicate function that inspects tool arguments.
+
+```python
+from typing import Any
+from google.adk.agents import LlmAgent
+from google.adk.tools.mcp_tool import McpToolset
+from google.adk.tools.mcp_tool.mcp_session_manager import StdioConnectionParams
+from mcp import StdioServerParameters
+
+def should_require_approval(query: str = "", **kwargs) -> bool:
```
joefernandez: This approval gate fails open. ADK calls the predicate with the **tool's own arguments** ([`mcp_tool.py:463`](https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/mcp_tool.py#L463), via [`_prepare_callable_args`](https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/mcp_tool.py#L428)), so for any MCP tool whose argument is not literally named `query` — `sql`, `statement`, `path`, `command` — the default `query=""` is used, the predicate returns `False`, and the destructive call runs with no confirmation. The table at line 16 promises this covers "Database mutations, destructive shell/file ops".

Suggest replacing lines 67-70 with a version that inspects every argument (declaring `tool_context` explicitly keeps ADK's injected context out of the scanned values):

```python
def should_require_approval(tool_context=None, **kwargs) -> bool:
    """Heuristic gate: inspects every argument the tool was called with."""
    destructive_keywords = ["drop", "delete", "truncate", "alter", "update", "insert"]
    call_text = " ".join(str(value) for value in kwargs.values()).lower()
    return any(keyword in call_text for keyword in destructive_keywords)
```

Please also add a sentence to this section saying that keyword matching is a heuristic, not a guarantee (`DELETE` spelled via a view, a parameterised statement, or `exec` all slip past it), and that `require_confirmation=True` is the safe default for a server that can destroy data.

docs/tools-custom/mcp-tools/agent-as-mcp-server.md
```diff
@@ -0,0 +1,188 @@
...
+
+* **Expose an entire Agent:** Wraps the full multi-turn agent reasoning and 
+  internal tool execution into a server using a simple one-line conversion.
+
+* **Expose individual Tools:** Involves manually building a lightweight MCP 
+  server to wrap specific, standalone ADK tools (like a `FunctionTool`) without 
+  the agent's reasoning loop.
+
+## Get started
+The most powerful approach is to expose an entire `LlmAgent`. By using the
```
joefernandez: `to_mcp_server()` carries `@experimental(FeatureName.MCP_AGENT_SERVER)` in ADK Python ([`_agent_to_mcp.py`](https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/_agent_to_mcp.py#L169)) and emits an experimental warning at runtime, but this page presents it as GA. Follow the convention already used on [`confirmation.md:22`](https://github.com/google/adk-docs/blob/7d2607aba1f7bd17d48f229d3cae6358bd9412b5/docs/tools-custom/confirmation.md#L22):

```suggestion

!!! example "Experimental"
    `to_mcp_server()` is experimental. Its API may change, and ADK emits an
    experimental warning at runtime when you use it.

The most complete approach is to expose an entire `LlmAgent`. By using the
```

Please also add `<span class="lst-preview">Experimental</span>` to a `language-support-tag` block at the top of this page, as [`confirmation.md:4`](https://github.com/google/adk-docs/blob/7d2607aba1f7bd17d48f229d3cae6358bd9412b5/docs/tools-custom/confirmation.md#L4) does.

docs/tools-custom/mcp-tools/mcp-tools-advanced.md
```diff
@@ -0,0 +1,357 @@
...
+error_file = open("mcp_server_errors.log", "a")
+try:
+    toolset = McpToolset(connection_params=..., errlog=error_file)
+    # Run agent...
+finally:
+    await toolset.close()
+    error_file.close()
+```
+
+## Render interactive UI widgets
```
joefernandez, line 288: The table at line 22 calls this "Experimental UI Rendering", but the section itself reads as a shipped feature. Label it here too:

```suggestion
## Render interactive UI widgets

!!! example "Experimental"
    MCP UI widget rendering is experimental. The `meta.ui.resourceUri`
    contract and the client rendering behavior may change.
```

docs/tools-custom/mcp-tools/index.md
```diff
@@ -0,0 +1,489 @@
...
+This section shows you how to connect to remote MCP servers using authentication and how to read data **Resources** exposed by an MCP server. When an MCP server requires authentication, such as over Server-Sent Events `SseConnectionParams` or Streamable HTTP, `McpToolset` handles credential injection and token management automatically.
+
+### Key authentication parameters
+
+| Parameter | Type | Description |
+| :--- | :--- | :--- |
+| `auth_scheme` | `AuthScheme` | The authentication strategy (e.g., `Bearer`, `Basic`, `APIKey`, `OAuth2`). |
+| `auth_credential` | `AuthCredential` | The secret credential payload, for example, API token, OAuth access token, username/password. |
+
+ADK automatically constructs the required `Authorization` HTTP headers and manages OAuth 2.0 token refreshes during client requests.
```
joefernandez, line 353: The refresh claim is not supported by anything on this page: both samples below inject a static `Authorization` header, and `McpToolset` builds headers from the credential it is given ([`mcp_toolset.py:280`](https://github.com/google/adk-python/blob/main/src/google/adk/tools/mcp_tool/mcp_toolset.py#L280)). A reader who copies those samples gets no refresh at all and will be surprised when the token expires.

```suggestion
ADK constructs the required `Authorization` HTTP header from the credential you provide. Token lifetime is not managed for you in the header-based configuration shown below: supply a valid credential, and refresh or re-issue it before it expires.
```

docs/tools-custom/mcp-tools/index.md
```diff
@@ -0,0 +1,489 @@
...
+
+### Configure authentication
+
+When an MCP server requires authentication, `McpToolset` handles credential injection and token management automatically. Use the native `auth_scheme` and `auth_credential` parameters rather than manually injecting HTTP headers.
```
joefernandez, line 357: This tells readers not to inject HTTP headers manually, and then both samples immediately below do exactly that and never use `auth_scheme`/`auth_credential`. Either add an `auth_scheme`/`auth_credential` sample or reword:

```suggestion
When an MCP server requires authentication, `McpToolset` injects credentials into outgoing requests. A static bearer token can be passed as a header, as shown below; for OAuth 2.0 and API-key schemes, prefer the native `auth_scheme` and `auth_credential` parameters so ADK handles the credential exchange.
```

docs/tools-custom/mcp-tools/index.md
```diff
@@ -0,0 +1,489 @@
...
+
+| Dimension | [**Direct MCP Tool Integration** (`McpToolset`)](#direct-mcp-tool-integration-mcptoolset) | [**Agent-Exposed MCP Server** (`to_mcp_server`)](agent-as-mcp-server.md) | [**Specialized Sub-Agent Delegation** (`AgentTool`)](agent-managed-mcp.md) |
+| :--- | :--- | :--- | :--- |
+| **Architecture** | External server process or remote service providing deterministic endpoints adapted into the primary `LlmAgent` tool list. | An autonomous ADK agent compiled into an MCP server, callable by external clients (Claude Code, IDEs, external hosts). | In-process, hierarchical agent encapsulation where a parent agent invokes a child `LlmAgent` as a callable tool. |
+| **Context Window Impact** | **High Context Bloat**: Every tool definition and raw output, for example: database rows or file blobs, enters the primary agent's history. | **Isolated**: The external caller only receives the final aggregated response text/blocks. | **Zero Context Bloat**: Intermediate exploratory reasoning, failed tool calls, and large raw outputs remain isolated in the sub-agent loop. |
```
joefernandez, line 80: "Zero Context Bloat" is not true as an absolute — the sub-agent's final response still enters the parent's history, and that is often the largest part.

```suggestion
| **Context Window Impact** | **High Context Bloat**: Every tool definition and raw output, for example: database rows or file blobs, enters the primary agent's history. | **Isolated**: The external caller only receives the final aggregated response text/blocks. | **Minimal Context Growth**: Intermediate exploratory reasoning, failed tool calls, and large raw outputs stay inside the sub-agent loop; only the sub-agent's final response enters the parent's history. |
```

docs/tools-custom/mcp-tools/index.md
```diff
@@ -0,0 +1,489 @@
...
+| **AI Model Load and Tiering** | Single model must understand all tool schemas, validation constraints, and workflow state simultaneously. | Independent model reasoning dedicated solely to the wrapped task. | Enables **model tiering**, for example: `gemini-2.5-pro` for orchestrator, and `gemini-2.5-flash` for sub-agent tool execution with dedicated system instructions. |
+| **Latency & Token Cost** | **Lower Cost & Predictable Latency**: 1 LLM turn + 1 deterministic tool invocation + 1 response generation turn. | Client-driven; latency depends on internal agent execution depth. | **Higher Cost & Variable Latency**: Multiple LLM calls, sub-agent reasoning turns before returning to parent. |
```
joefernandez, line 82: Cost and latency depend on the model, the server, and the workload; these are stated as absolutes. Suggest hedging:

```suggestion
| **Latency & Token Cost** | **Typically Lower Cost & More Predictable Latency**: 1 LLM turn + 1 deterministic tool invocation + 1 response generation turn. | Client-driven; latency depends on internal agent execution depth. | **Typically Higher Cost & More Variable Latency**: Multiple LLM calls, sub-agent reasoning turns before returning to parent. |
```

docs/tools-custom/mcp-tools/index.md
```diff
@@ -0,0 +1,489 @@
...
+| **Ideal Use Cases** | <ul><li>Deterministic API integrations: Postgres, BigQuery, GitHub, Google Maps.</li><li>File system operations & static resource reading.</li><li>Reusing standard pre-built community MCP servers.</li></ul> | <ul><li>Exposing complex ADK multi-agent capabilities to external MCP-compliant ecosystems.</li><li>Integrating ADK agents into IDEs, editors, or A2A pipelines.</li></ul> | <ul><li>Multi-step autonomous workflows requiring trial-and-error. For example: code debugging or research synthesis.</li><li>Tasks needing isolated personas or specialized instructions.</li><li>Scenarios with >20 tools where schema overload harms accuracy.</li></ul> |
```
joefernandez, line 83: ">20 tools" is a specific quantitative accuracy claim with no source. Either cite a benchmark or drop the number:

```suggestion
| **Ideal Use Cases** | <ul><li>Deterministic API integrations: Postgres, BigQuery, GitHub, Google Maps.</li><li>File system operations & static resource reading.</li><li>Reusing standard pre-built community MCP servers.</li></ul> | <ul><li>Exposing complex ADK multi-agent capabilities to external MCP-compliant ecosystems.</li><li>Integrating ADK agents into IDEs, editors, or A2A pipelines.</li></ul> | <ul><li>Multi-step autonomous workflows requiring trial-and-error. For example: code debugging or research synthesis.</li><li>Tasks needing isolated personas or specialized instructions.</li><li>Large tool inventories, where schema overload harms tool-selection accuracy.</li></ul> |
```

mkdocs.yml
```diff
@@ -148,7 +148,8 @@ plugins:
         'tools/function-tools.md': 'tools-custom/function-tools.md'
         'tools/confirmation.md': 'tools-custom/confirmation.md'
         'tools/performance.md': 'tools-custom/performance.md'
-        'tools/mcp-tools.md': 'tools-custom/mcp-tools.md'
+        'tools/mcp-tools.md': 'tools-custom/mcp-tools/index.md'
+        'mcp/index.md': 'tools-custom/mcp-tools/index.md'
```
joefernandez, line 152: This redirect makes existing content unreachable rather than migrating it. `docs/mcp/index.md` is **not** deleted in this PR, but mkdocs-redirects writes its stub over the built page, so `/mcp/` becomes a redirect and the ["MCP Servers for Google Cloud Genmedia" section](https://github.com/google/adk-docs/blob/7d2607aba1f7bd17d48f229d3cae6358bd9412b5/docs/mcp/index.md#L50-L63) drops off the site. `grep` finds "Genmedia" nowhere else under `docs/`.

Please either port that section into the new `tools-custom/mcp-tools/` tree or delete `docs/mcp/index.md` in this PR so the loss is visible in review. Note [`docs/integrations/temporal.md:192`](https://github.com/google/adk-docs/blob/7d2607aba1f7bd17d48f229d3cae6358bd9412b5/docs/integrations/temporal.md#L192) still links to `/mcp/`.

docs/tools-custom/mcp-tools/agent-managed-mcp.md
```diff
@@ -0,0 +1,84 @@
...
+from google.adk.tools.agent_tool import AgentTool
+from google.adk.tools.mcp_tool import McpToolset
+from google.adk.tools.mcp_tool.mcp_session_manager import StdioConnectionParams
+from mcp import StdioServerParameters
+
+# Step 1: Initialize your MCP connection 
+mcp_connection = StdioConnectionParams(
+    server_params=StdioServerParameters(
+        command='npx',
+        args=['-y', '@modelcontextprotocol/server-postgres', 'postgres://user:pass@localhost:5432/db'],
```
joefernandez: Hardcoded database credentials in a sample agent file — readers copy these. `os` is already imported on line 12:

```suggestion
        args=['-y', '@modelcontextprotocol/server-postgres', os.environ['DATABASE_URL']],
```

Worth a one-line note under the snippet that connection strings should come from an environment variable or Secret Manager, never from source.

docs/tools-custom/mcp-tools/mcp-tools-advanced.md
```diff
@@ -0,0 +1,357 @@
...
+In multi-tenant or user-facing systems, hardcoding credentials into connection parameters is insecure. `McpToolset` supports `header_provider`, an asynchronous or synchronous callable that receives the active `ReadonlyContext` to dynamically construct authentication headers on every tool invocation.
+
+```python
+from google.adk.agents import LlmAgent
+from google.adk.agents.readonly_context import ReadonlyContext
+from google.adk.tools.mcp_tool import McpToolset, StreamableHTTPConnectionParams
+
+async def extract_per_user_headers(context: ReadonlyContext) -> dict[str, str]:
+    """Dynamically extracts session state or per-user token on every turn."""
+    user_token = context.state.get("user_access_token", "ANONYMOUS_TOKEN")
```
joefernandez: In a section specifically about per-user credentials, this fails open: when state has no token, the agent sends a literal `Bearer ANONYMOUS_TOKEN` upstream instead of stopping. Fail closed:

```suggestion
    user_token = context.state.get("user_access_token")
    if not user_token:
        raise ValueError("No per-user access token in session state; refusing to call the MCP server.")
```

docs/tools-custom/mcp-tools/index.md
```diff
@@ -0,0 +1,489 @@
+# Model Context Protocol tools
+
+<div class="language-support-tag">
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v0.1.0</span><span class="lst-typescript">Typescript v0.2.0</span><span class="lst-go">Go v0.1.0</span><span class="lst-java">Java v0.1.0</span><span class="lst-kotlin">Kotlin v0.1.0</span>
```
joefernandez, line 4: Two regressions against the page this replaces: `Kotlin v0.7.0` became `v0.1.0` (that asserts MCP tools worked six minor versions earlier than previously published — please confirm before shipping), and `TypeScript` lost its capital S, which the rest of the site uses.

```suggestion
  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v0.1.0</span><span class="lst-typescript">TypeScript v0.2.0</span><span class="lst-go">Go v0.1.0</span><span class="lst-java">Java v0.1.0</span><span class="lst-kotlin">Kotlin v0.7.0</span>
```

docs/tools-custom/mcp-tools/index.md
```diff
@@ -0,0 +1,489 @@
...
+      - Launch the Web Interface:
+
+        ```bash
+        cd ./adk_agent_samples 
+        adk web
+        ```
+    
+    - Interact with the Agent: select `filesystem_assistant` from the drop-down menu and prompt the Agent with commands: *List files in the current directory* or *What is the content of another_file.md?*
+
+    ```
```
joefernandez: Unmatched code fence at the end of the Python tab — it renders as a stray empty code block. Delete the line:

```suggestion
```

docs/tools-custom/mcp-tools/deployment.md
```diff
@@ -0,0 +1,334 @@
+# Deploy Agents with MCP Tools
```
joefernandez, line 1: This page has no `language-support-tag` block, unlike every other page in `docs/tools-custom/`, so readers get no supported-language signal on a page whose samples are Python-only. Same applies to `agent-as-mcp-server.md` and `agent-managed-mcp.md`.

```suggestion
# Deploy Agents with MCP Tools

<div class="language-support-tag">
  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v0.1.0</span>
</div>
```

docs/tools-custom/mcp-tools/mcp-tools-advanced.md
```diff
@@ -0,0 +1,357 @@
...
+      mcp_types.TextContent(type="text", text=f"Unknown tool: '{name}'")
+  ]
+
+``` 
+
+---
+
+## More information
+
+* Return to the [Model Context Protocol Overview](./index.md) for basic setup.
```
joefernandez: The deleted page's "Further Resources" links to the MCP spec and SDK are gone from all five new pages; this section only links back to the new index. Suggest restoring them here:

```suggestion
* Return to the [Model Context Protocol Overview](./index.md) for basic setup.
* [Model Context Protocol documentation](https://modelcontextprotocol.io/)
* [MCP specification](https://modelcontextprotocol.io/specification/)
* [MCP Python SDK and examples](https://github.com/modelcontextprotocol/)
```

docs/tools-custom/mcp-tools/index.md
```diff
@@ -0,0 +1,489 @@
...
+| **Ideal Use Cases** | <ul><li>Deterministic API integrations: Postgres, BigQuery, GitHub, Google Maps.</li><li>File system operations & static resource reading.</li><li>Reusing standard pre-built community MCP servers.</li></ul> | <ul><li>Exposing complex ADK multi-agent capabilities to external MCP-compliant ecosystems.</li><li>Integrating ADK agents into IDEs, editors, or A2A pipelines.</li></ul> | <ul><li>Multi-step autonomous workflows requiring trial-and-error. For example: code debugging or research synthesis.</li><li>Tasks needing isolated personas or specialized instructions.</li><li>Scenarios with >20 tools where schema overload harms accuracy.</li></ul> |
+
+!!! note "State restoration"
+
+    While ADK agents preserve session state during lifecycle events, they do not automatically re-establish active MCP connections upon restoration. Agents re-initialize connections as needed.
+   
+### Direct MCP Tool integration (McpToolset)
```
joefernandez, line 89: This rename breaks an inbound deep link: [`docs/integrations/agent-registry.md:390`](https://github.com/google/adk-docs/blob/7d2607aba1f7bd17d48f229d3cae6358bd9412b5/docs/integrations/agent-registry.md#L390) points at `/tools-custom/mcp-tools/#mcptoolset-class`, and the anchor is now `#direct-mcp-tool-integration-mcptoolset`. The page URL survives, only the anchor dies.

Fix it in the other file (don't override the id here with `attr_list` — the comparison table on line 77 links to the generated anchor):

```markdown
requests made by the [McpToolset](/tools-custom/mcp-tools/#direct-mcp-tool-integration-mcptoolset)
```

docs/tools-custom/mcp-tools/deployment.md
```diff
@@ -0,0 +1,334 @@
...
+            ],
+        )
+    ],
+)
+```
+
+```python
+# WRONG: Asynchronous patterns don't work in deployment
+async def get_agent():  # This won't work for deployment
+    toolset = await create_mcp_toolset_async()
```
joefernandez: Consider adding just this line to the warning to help developers know what to look out for. Otherwise, remove this code sample; showing what not to do in detail often causes innattentive users (and agents) to do exactly what they shouldn't.

docs/tools-custom/mcp-tools/deployment.md
```diff
@@ -0,0 +1,334 @@
+# Deploy Agents with MCP Tools
+
+When deploying ADK agents that use MCP tools to production environments like Cloud Run, GKE, or Agent Runtime, you need to consider how MCP connections will work in containerized and distributed environments.
+
+## Critical Deployment Requirement: Synchronous Agent Definition
+
+!!! warning
```
joefernandez: 1. move this warning until after the code example
2. add an introductory paragraph about what developers *should* do

docs/tools-custom/mcp-tools/deployment.md
```diff
@@ -0,0 +1,334 @@
...
+
+```python
+# WRONG: Asynchronous patterns don't work in deployment
+async def get_agent():  # This won't work for deployment
+    toolset = await create_mcp_toolset_async()
+    return LlmAgent(tools=[toolset])
+```
+
+## Quick Deployment Commands
+
```
joefernandez, line 58: don't stack headings. Add a introductory sentence here.

docs/tools-custom/mcp-tools/deployment.md
```diff
@@ -0,0 +1,334 @@
...
+```bash
+uv run adk deploy cloud_run \
+  --project=<your-gcp-project-id> \
+  --region=<your-gcp-region> \
+  --service_name=<your-service-name> \
+  ./path/to/your/agent_directory
+```
+
+## Deployment Patterns
+
```
joefernandez: don't stack headings: intro sentence

joefernandez (approved):
Approved with comments. Let me know if you want me to do a final review.

PR #2082, opened 2026-08-06
 docs: fix App import in Freeplay integration from google.adk.runners to google.adk.apps

Syed-Umer-Ali:
## What
The example imports `App` from `google.adk.runners`, but `App` does not exist there.

## Why
Running `from google.adk.runners import App` raises `ImportError`. The `App` class lives in `google.adk.apps` (verified: `google.adk.apps.App` exists with fields `name`, `root_agent`, `plugins`, `events_compaction_config`, `context_cache_config`, `resumability_config`; `google.adk.runners` only exports `Runner`/`InMemoryRunner`). The example uses `App` with `root_agent=` and `plugins=`, which matches the real API.

## Fix
Changed the import on line 77 to `from google.adk.apps import App`.

## Verification
`python -c "from google.adk.apps import App"` succeeds; the `App(name=..., root_agent=..., plugins=[...])` call matches the class's model fields.


joefernandez:
Thanks for the contribution. These updates are now out of date or does not follow our information strategy. Closing.

PR #2085, opened 2026-08-06
feat(search): replace the lunr search index with Pagefind

haranrk:
## Summary

Replaces MkDocs' built-in lunr.js search with [Pagefind](https://pagefind.app/).

The built-in `search` plugin builds one lunr index and the browser downloads all of it before the first query. On this site that's **3.5 MB (843 KB gzipped)**, fetched on every page load whether or not the reader opens search. Pagefind shards the index at build time and ships a WASM query engine, so the browser fetches only the shards a query touches.

## Measured impact

Built the previous commit as a baseline, served both locally, and read the **server access log** — Pagefind fetches its index from a Web Worker, and those requests don't appear in page-level devtools events.

| | before (lunr) | after (Pagefind) |
|---|---|---|
| on page load | **855 KB gzip** | **46 KB gzip** |
| on first search | 0 | 378 KB gzip |
| total, if the reader searches | 855 KB | 425 KB |

Readers who never search pay **95% less**. Readers who do pay about half.

One honest caveat: unlike lunr, Pagefind's cost isn't one-time — it fetches more fragments as you scroll results, so a session with many queries eventually crosses over.

## How it works

- `hooks/pagefind_index.py` runs the indexer in `on_post_build`: **+2.2s** on a ~31s build.
- Pagefind ships as a **PyPI wheel with a prebuilt binary**, so it installs from `requirements.txt` with **no Node toolchain** in CI. A `manylinux` wheel exists for `ubuntu-latest`.
- Index scope is set by `data-pagefind-body` on the article, via a `container` block override mirroring the theme's own. Once that attribute exists anywhere, Pagefind indexes only pages carrying it — which keeps the **3,130 generated Dokka/Javadoc/TypeDoc files** out. Those were never in the lunr index either, so **search scope is unchanged**: 232 pages in, 232 indexed.
- UI is Pagefind's **Component UI** (`<pagefind-modal>`), which supersedes the older Default UI as of Pagefind 1.5.0. Keeps `Cmd/Ctrl+K` and adds WAI-ARIA compliance.
- Theming maps `--pf-*` onto Material's `--md-*` **on `body`, not `:root`** — Material declares its light palette at `:root` but its dark palette on `[data-md-color-scheme]` on `<body>`, so mapping at `:root` would stay light after the palette toggle. Custom properties are exempt from Pagefind's `all: initial` host reset, so **dark mode needs no JavaScript**.
- The components live in the header, outside every region Material's instant navigation swaps, so they stay mounted across client-side navigation.

`PAGEFIND_SKIP=1` skips indexing, e.g. for link-checking builds.

## Index quality fixes

Pagefind concatenates adjacent inline elements without a separator, so siblings authored without whitespace fuse into one token. It does **not** fix this for free — three constructs here hit it, removed via `--exclude-selectors`:

| selector | produced |
|---|---|
| `.headerlink` | `¶` permalink anchors fused into headings on **230 of 232** pages |
| `.tabbed-labels` | `<label>Python</label><label>Java</label>` → `PythonJava` |
| `.language-support-tag` | `ADKPythonTypeScriptGoJava` |

Pages with fused language tokens: **109/232 → 5/232**, and all five remaining are genuine identifiers (`RxJava`, `pdfArtifactJava` is a real variable in a Java sample). Tab *contents* stay indexed, so per-language content remains searchable — verified `Kotlin agent` → 25 results, `Java streaming` → 34.

`--include-characters ".-@#+"` keeps dotted identifiers intact; without it `google.adk.agents` degrades to three bare words (133 results with it).

## Verification

Driven in headless Chrome against the built site:

- trigger renders, `Ctrl+K` opens, queries return results, and `¶` permalink anchors no longer appear in indexed headings
- **zero** requests for `search_index.json` or the lunr worker
- no JS errors, no same-origin HTTP failures
- dark mode: `dialogBg` resolves to `rgb(30, 33, 41)` with light text after the palette toggle
- **instant navigation**: same component node persists, no duplicates, search still works. This needed care — instant nav is inert on `localhost` because Material gates interception on `sitemap.xml`, whose URLs are absolute to `adk.dev`. I rebuilt with a local `site_url` to reproduce production conditions, and confirmed production `adk.dev` does use it.
- `mkdocs build --strict` passes; `mkdocs serve` works (search functions in dev too)
- both failure paths tested: `PAGEFIND_SKIP=1`, and a clear error if the dependency is missing

## Note for reviewers

Screenshots of the modal in light and dark mode can be attached if useful. Licensing is unchanged: Pagefind is MIT, same as mkdocs-material and lunr.

### DIFF
--- a/docs/stylesheets/pagefind.css
+++ b/docs/stylesheets/pagefind.css
@@ -1,0 +1,95 @@
+/**
+ * Copyright 2026 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Themes the Pagefind search components to match Material for MkDocs.
+ *
+ * Pagefind applies `all: initial` to its host elements, so no site styles
+ * cascade in. Custom properties are the documented theming hook, and they are
+ * exempt from `all`, so mapping Material's palette onto the `--pf-*` variables
+ * is enough to theme the whole search UI - including dark mode - without any
+ * JavaScript.
+ *
+ * These are declared on `body` rather than `:root` on purpose. Material only
+ * declares its default palette at `:root`; the dark palette is declared on
+ * `[data-md-color-scheme="slate"]`, which the theme sets on `<body>`. Mapping
+ * at `:root` would resolve against the light palette and stay light after the
+ * palette toggle.
+ */
+body {
+  --pf-font: var(--md-text-font-family, system-ui, sans-serif);
+
+  --pf-text: var(--md-default-fg-color);
+  --pf-text-secondary: var(--md-default-fg-color--light);
+  --pf-text-muted: var(--md-default-fg-color--light);
+  --pf-background: var(--md-default-bg-color);
+
+  --pf-border: var(--md-default-fg-color--lightest);
+  --pf-border-focus: var(--md-default-fg-color--lighter);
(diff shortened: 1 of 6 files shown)


haranrk:
Superseded by https://github.com/google/adk-docs/pull/2096

PR #2088, opened 2026-08-07
Upgrade Kotlin examples to adk-kotlin 0.7.0

happyhuman:
## Summary

The Kotlin examples project is pinned to adk-kotlin **0.5.0**, which predates
the context caching, Vertex AI memory, and RAG retrieval APIs. This bumps the
pin to **0.7.0** so snippets for those features can be written.

This PR is deliberately just the upgrade and its fallout — no new snippets. It
is the base for three follow-up docs PRs (context caching, Memory Bank / RAG
memory, and Knowledge Engine retrieval).

## What's in it

**The bump itself**

- `google-adk-kotlin-core`, `-webserver`, and `-processor` 0.5.0 → 0.7.0.

**Three consequences that have to land with it**

- `ExperimentalResumabilityFeature` was **removed** in 0.7.0 — the annotation
  class is gone, not merely deprecated. `RunConfigExample.kt` no longer opts
  into it. This is a hard compile break at 0.7.0, not a warning that could be
  deferred to a follow-up.
- Added `ktor-client-core` / `ktor-client-java`. The Vertex AI session and
  memory services expose Ktor's `HttpClient` as a *defaulted constructor
  parameter*, so any snippet naming them needs Ktor on the **compile**
  classpath, not just at runtime.
- Removed the `resolutionStrategy` block forcing kotlin-stdlib 2.1.20. It
  worked around 0.5.0 publishing a stdlib newer than this project's compiler;
  0.6.0 fixed that upstream. Verified the highest stdlib on the compile
  classpath is still 2.1.20 without it, so the block was dead code carrying a
  reason that no longer applies.

**Two unrelated snags found while validating the above**

- `check_kotlin_snippets.sh` walked `examples/kotlin` with a bare `find` and no
  build-output pruning, so after any local build it reported every generated
  KSP file as an unregistered snippet — hundreds of false failures. CI only
  ever runs it against a clean checkout, so the bug was invisible there but
  made the script unusable locally. Now prunes `build/`, `bin/`, `out/`,
  `.gradle/`.
- A transclusion path in `docs/observability/logging.md` was split across two
  lines, so the include never resolved and the code block rendered empty.
- The `plugin/` → `plugins/` link fixes in `docs/integrations/reflect-and-retry.md`
  and `docs/plugins/index.md`. adk-python renamed that directory; the three
  links 404 and were failing the repo-wide `link-check` job. **These are
  byte-identical to a commit in #2091**, because both PRs target `main` and
  each needs them to go green independently. Whichever merges second will see
  the change already applied. Happy to drop them from whichever you'd rather
  carry it.

## Verification

- `./gradlew compileKotlin` passes at 0.7.0.
- `check_kotlin_snippets.sh` passes, and now also passes against a working tree
  that has been built locally.
- ktlint clean on every changed file.


joefernandez (approved):
Thanks for the updates!

PR #2091, opened 2026-08-07
Add Kotlin snippet for context caching configuration

happyhuman:
## Summary

`docs/context/caching.md` showed Python and Java only. `ContextCacheConfig` and
the `App`-level wiring are available in adk-kotlin as of 0.7.0, so this adds a
Kotlin tab and advertises Kotlin support in the page badge.

> **Independent of #2088** despite documenting a 0.7.0 API. The snippet is
> inline, so it is never compiled by `kotlin-snippets-pr-check.yaml`, which
> builds only changed `.kt` files under `examples/kotlin/snippets/`. This can
> merge on its own.

## What's in it

- Kotlin tab on the configuration example, plus the `lst-kotlin` badge (the
  page had no prior Kotlin content, so `v0.7.0` is the version support was
  introduced).
- The `plugin/` → `plugins/` link fixes for `docs/integrations/reflect-and-retry.md`
  and `docs/plugins/index.md`. **These are byte-identical to a commit in #2088**,
  because both PRs target `main` and each needs them to get `link-check` green
  independently. Whichever merges second will see the change already applied.
  Happy to drop them from whichever you'd rather carry it.

## Notes for reviewers

- **Inline rather than transcluded**, to match the Python and Java tabs on this
  page. The trade-off is real: it gets no compile coverage in CI, and it was
  checked against the adk-kotlin 0.7.0 sources by hand. If you'd rather have
  the regression coverage, say so and I'll move it to a `.kt` file plus a
  `files_to_test.txt` entry at the cost of diverging from the page's format.
- **`minTokens = 8192`, not Python's 2048.** The 0.7.0 KDoc states Gemini
  enforces a hard 4096-token floor that always applies, "so values below 4096
  have no additional effect" — copying 2048 across would have documented an
  inert setting.
- Opts in at file level via `ExperimentalContextCachingFeature`. The
  constructor is annotated `@RequiresOptIn(level = ERROR)`, so this is
  mandatory, not advisory. Note that in a real file `@file:OptIn(...)` must
  precede the `package` declaration.
- Uses `kotlin.time` durations so the TTL reads as `10.minutes` rather than
  mirroring Java's `Duration.ofMinutes(10)`, and names the parameters instead
  of carrying over the Java tab's positional args with trailing comments.


docs/context/caching.md
```diff
@@ -64,6 +64,41 @@ these settings, as shown in the following code sample:
                  .build();
     ```
 
+=== "Kotlin"
+
+    ```kotlin
+    @file:OptIn(ExperimentalContextCachingFeature::class)
```
wikaaaaa, line 70: The PR description mentions "Note that in a real file @file:OptIn(...) must
precede the package declaration." -- This compiles as rendered only because there's no package line; pasted into a real file the ordering is wrong. 

Can we make the opt-in statement-level? So instead `@file:OptIn` -> `@OptIn(ExperimentalContextCachingFeature::class)` directly above ```val app =```.

That would match the way we used the experimental opt in for resumability (removed in https://github.com/google/adk-docs/pull/2088).

wikaaaaa (approved):
Approved with one comment on the OptIn placement.

joefernandez:
## Technical review report

Verdict: the Kotlin code is technically correct

Compiled the snippet verbatim against the published `com.google.adk:google-adk-kotlin-core-jvm:0.7.0` artifact:

| Test | Version | Result |
|---|---|---|
| `compileKotlin`, snippet as written | 0.7.0 | **SUCCESS**, no errors/warnings |
| Construct `app`/`rootAgent` at runtime | 0.7.0 | **`app = my-caching-agent-app / agent = my_caching_agent`** |
| Same snippet, `@file:OptIn` removed | 0.7.0 | **Compile error** (proves opt-in is mandatory) |
| Construct `app` at runtime | 0.6.0 | **Throws** `IllegalArgumentException: Invalid app name 'my-caching-agent-app'` |

Every API element verified against the `v0.7.0` tag (I confirmed all cited files except `Gemini.kt` are byte-identical to `main`):

- [`ContextCacheConfig.kt#L50-L57`](https://github.com/google/adk-kotlin/blob/v0.7.0/core/src/commonMain/kotlin/com/google/adk/kt/agents/ContextCacheConfig.kt#L50-L57) — `minTokens`/`ttl`/`cacheIntervals` all exist; `ttl` is a `kotlin.time.Duration`, so `10.minutes` is right
- [`ExperimentalContextCachingFeature.kt#L27-L33`](https://github.com/google/adk-kotlin/blob/v0.7.0/core/src/commonMain/kotlin/com/google/adk/kt/annotations/ExperimentalContextCachingFeature.kt#L27-L33) — `@RequiresOptIn(level = ERROR)`, confirming "mandatory, not advisory"
- [`App.kt#L49-L56`](https://github.com/google/adk-kotlin/blob/v0.7.0/core/src/commonMain/kotlin/com/google/adk/kt/apps/App.kt#L49-L56) and [`LlmAgent.kt#L129-L131`](https://github.com/google/adk-kotlin/blob/v0.7.0/core/src/commonMain/kotlin/com/google/adk/kt/agents/LlmAgent.kt#L129-L131)
- [`Gemini.kt#L89-L93`](https://github.com/google/adk-kotlin/blob/v0.7.0/core/src/commonMain/kotlin/com/google/adk/kt/models/Gemini.kt#L89-L93) — `Gemini(name = …)` resolves to the `@JvmOverloads` API-key constructor, not the primary one (which needs a `Client`)

## Two issues found

**1. The `v0.7.0` badge is wrong — caching shipped in 0.6.0.** The PR's stated rationale ("the page had no prior Kotlin content, so v0.7.0 is when support was introduced") is a non-sequitur, and the fact is wrong: config type, request processor, Gemini manager, and App→Runner propagation are all in the [0.6.0 changelog](https://github.com/google/adk-kotlin/blob/v0.7.0/CHANGELOG.md?plain=1#L57-L59), and [`App.contextCacheConfig` exists at v0.6.0](https://github.com/google/adk-kotlin/blob/v0.6.0/core/src/commonMain/kotlin/com/google/adk/kt/apps/App.kt#L55). But there *is* a hidden 0.7.0 dependency: the hyphenated `appName`, since [v0.6.0's regex rejected hyphens](https://github.com/google/adk-kotlin/blob/v0.6.0/core/src/commonMain/kotlin/com/google/adk/kt/apps/App.kt#L68). So either badge `v0.6.0` + use `my_caching_agent_app`, or keep `v0.7.0` and fix the description.

**2. The "hard 4096-token floor" comment overstates a model-specific minimum.** It quotes the [adk-kotlin KDoc](https://github.com/google/adk-kotlin/blob/v0.7.0/core/src/commonMain/kotlin/com/google/adk/kt/agents/ContextCacheConfig.kt#L40-L41) faithfully, but that KDoc is the outlier — [adk-python](https://github.com/google/adk-python/blob/0477e5743bdf5e0cce5a698951c7b70a07baa80a/src/google/adk/agents/context_cache_config.py#L38-L41) and [adk-kotlin's own demo](https://github.com/google/adk-kotlin/blob/v0.7.0/examples/src/main/kotlin/com/google/adk/kt/examples/caching/ContextCachingDemoAgent.kt#L47-L50) both say 2048 for Gemini 2.5, 4096 for Gemini 3. `minTokens = 8192` is still the right value; only the reasoning needs rewording.

Also noted: the PR body's CI claim checks out (the workflow only builds changed `.kt` files under `examples/kotlin/snippets/`), and the `plugin/`→`plugins/` link fixes already landed via #2088 so they're no longer in the diff.

docs/context/caching.md
```diff
@@ -64,6 +64,41 @@ these settings, as shown in the following code sample:
...
+        )
+
+    // Create the app with context caching configuration
+    val app =
+        App(
+            appName = "my-caching-agent-app",
+            rootAgent = rootAgent,
+            contextCacheConfig =
+                ContextCacheConfig(
+                    // Gemini enforces a hard 4096-token floor of its own, so only a
```
joefernandez, line 93: See Tech Report Issue #2 about "hard token floor" - This is not the same ("4095") value for all model versions.

https://github.com/google/adk-docs/pull/2091#issuecomment-5243184877

remove the "4096" value and let the user look up the minimum value per model, if needed.

joefernandez (approved):
Approved with a comment.

happyhuman:
Your comments landed just before this merged and I missed them — sorry. Branch was already deleted and a merged PR can't be reopened, so I applied all three in #2104:

- badge `v0.7.0` → `v0.6.0`, with `appName` switched to `my_caching_agent_app` so the hyphen stops being the version floor
- the "hard 4096-token floor" wording removed; no number cited, reader is pointed at their model's docs
- @wikaaaaa's `@file:OptIn` → statement-level `@OptIn` on the `app` declaration

Ran it in a throwaway project against both pins to check the badge rather than assert it, since the appName rule is a runtime `require`:

| pin | appName | result |
|---|---|---|
| 0.6.0 | `my_caching_agent_app` | `OK app=my_caching_agent_app minTokens=8192` |
| 0.6.0 | `my-caching-agent-app` | throws `IllegalArgumentException: Invalid app name` |
| 0.7.0 | `my_caching_agent_app` | `OK` |

Thanks for the depth on both of these — the 0.6.0 changelog trace and the adk-python/demo cross-check were the parts I should have done myself.

PR #2093, opened 2026-08-07
feat(search): replace lunr with Pagefind and a grouped-results search modal

haranrk:
## Summary

Replaces the built-in `search` plugin (lunr.js) with [Pagefind](https://pagefind.app), plus a search modal whose markup and behaviour live in this repo.

Two problems, solved together:

**Relevance.** The lunr index is tokenized with the bare default separator (`[\s\-]+` — whitespace and hyphens only), while **60% of the index sits inside `<code>`**. Searching `append_event`, `LlmAgent`, `google.adk.agents` or `before_agent_callback` currently returns **nothing**. Separately, Pagefind and lunr both concatenate adjacent inline elements, so the language badge and tab-label strip fuse into single tokens — `ADKPython` × 154, `PythonTypeScriptGoJava` × 59, across **193 of 226 pages**. A reader searching `Python` misses all of them.

**Interface.** Material's dropdown returns a flat list of section rows. This returns results grouped by page with the matching headings nested beneath, like [genkit.dev](https://genkit.dev) and [pydantic.dev/docs](https://pydantic.dev/docs) (both of which also use Pagefind).

## The indexed page set is unchanged

The main risk in swapping search engines is quietly changing what is searchable. It did not happen:

| | pages | URL set |
| --- | --- | --- |
| lunr on `main` | 226 | — |
| Pagefind here | 226 | **identical** |

Zero pages added, zero dropped. Verified by extracting both sets from CI-equivalent builds (`git archive`).

## Measured effects

| | before | after |
| --- | --- | --- |
| search bytes per page load | 841 KB gz index + 39 KB worker | **~9 KB gz** (`search.js` + `search.css`) |
| index / WASM / engine | fetched eagerly on every page | deferred to first modal open |
| `append_event`, `LlmAgent`, `google.adk.agents`, `before_agent_callback` | 0 results | all return results |
| fused language tokens | 219 across 193 pages | **0** |
| build time | — | +2.4s on a ~30s build |
| CI install | — | +5.4 MB |

The index is sharded, so a query fetches only the shards it touches (~162 KB to initialise, ~460 KB for a full cold search). A reader who never searches pays nothing beyond the ~9 KB.

## How it works

A `on_post_build` hook runs the Pagefind binary over the rendered HTML in `site_dir`. `data-pagefind-body` on `<article>` scopes the index — and because Pagefind indexes *only* pages carrying that attribute, it also keeps the 3,140 generated API-reference files out, matching the current lunr scope exactly. The attribute is conditional on `{% if page %}`, which keeps it off `404.html` (MkDocs renders theme templates without a page in context).

`--exclude-selectors ".headerlink, .tabbed-labels, .language-support-tag"` fixes the token fusion; tab *contents* stay indexed, so per-language content remains searchable. `--include-characters ".-@#+"` keeps `google.adk.agents` from splitting into three unrelated words.

Pagefind is used as a query API rather than a widget, so result rendering, keyboard handling and styling are ours — which is what makes the page-grouped layout and the deferred loading possible.

## Notable details

- **The trigger and the dialog are deliberately split across the `</header>` boundary.** The dialog must sit outside the seven elements instant navigation replaces (so it survives client-side navigation) *and* outside `.md-header`, because `custom.css`'s `[data-md-color-scheme="slate"] .md-header *` at specificity 0-2-0 beats the modal's own rules and flattens every colour in dark mode. Only the split satisfies both.
- **`target="_self"` on result links is load-bearing.** Material intercepts internal links by stripping query and hash and testing the bare URL against its sitemap, so `?pagefind-highlight=` links would be intercepted — leaving the dialog open and preventing the arrival highlighter from running. Material's handler bails on a truthy `target`.
- **Material's `__search` toggle input stays.** `toggle/index.ts` resolves it eagerly at module scope and throws if missing, which would take down all Material JS.
- **Keyboard**: `Ctrl`/`Cmd`+`K`, plus `/`, `s` and `f` rebound — those were bound inside Material's `mountSearch`, which no longer runs, so they would otherwise have silently stopped working. Suppressed while typing in a field.
- **Accessibility**: ARIA combobox/listbox with `aria-activedescendant`; native `<dialog>` for focus trapping and Escape.
- Search-term highlighting on arrival is preserved (`pagefind-highlight.js`, loaded only when the URL carries the param, and palette-aware via `--md-typeset-mark-color`).
- No Node toolchain — Pagefind ships as a Python wheel. `pagefind-bin` is pinned explicitly because `pagefind[bin]` only constrains the binary to `<2`.

## Guardrails

`hooks/pagefind_index.py` fails the build on: a missing or failing indexer, zero indexable pages, a page-count mismatch, and any `--exclude-selectors` entry that has stopped matching (a stale-but-valid selector would otherwise silently let fused tokens back in).

`scripts/verify_search_index.py` runs after the build in both `build-docs.yaml` and `publish-docs.yaml`, asserting index scope, that nothing under the generated API-reference subtrees or `404.html` is indexed, and that no fused tokens survive. Counts are always measured, never hardcoded.

`PAGEFIND_SKIP=1` skips indexing for `mkdocs serve` authoring loops; it logs a warning, so `--strict` builds fail while it is set.

## Testing

- `mkdocs build --strict` clean; all 10 commits build individually (bisect-safe).
- Verified interactively in Chrome: lazy loading (zero pagefind requests on page load), `Ctrl+K`, `/`/`s`/`f`, Escape, grouped results, arrow-key navigation, `aria-activedescendant`, full-document navigation on Enter, 167 terms marked on arrival, dark mode, and header layout from 1000px to 1920px.
- Fused-token fix proven with a control build: identical site indexed without `--exclude-selectors` → 219 tokens across 193 pages; with → 0.

### DIFF
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -39,3 +39,4 @@ jobs:
             mkdocs-material-
       - run: pip install -r requirements.txt
       - run: mkdocs build --strict
+      - run: python scripts/verify_search_index.py site
--- a/.github/workflows/publish-docs.yaml
+++ b/.github/workflows/publish-docs.yaml
@@ -42,4 +42,6 @@ jobs:
           restore-keys: |
             mkdocs-material-
       - run: pip install -r requirements.txt
+      - run: mkdocs build --strict
+      - run: python scripts/verify_search_index.py site
       - run: mkdocs gh-deploy --force
--- a/.gitignore
+++ b/.gitignore
@@ -177,6 +177,10 @@ cython_debug/
 # mkdocs documentation
 docs/site
 
+# Local design/plan scratch space. Never committed; see the LOCAL_ONLY_PREFIX
+# exemption in scripts/verify_search_index.py.
+docs/superpowers/
+
 # Utilities for local development
 .gemini/
 .scripts/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -102,6 +102,10 @@ when using compatible AI coding tools in this repo.
     This command starts a local docs server, typically at
     `http://127.0.0.1:8000/`.
 
+    Set `PAGEFIND_SKIP=1` to skip building the search index on each rebuild
+    while authoring; site search will not work in that build, and because
+    skipping logs a warning, `mkdocs build --strict` fails while it is set.
+
 ## Create Your Contribution
 
 | Type | Description |
--- a/docs/javascript/search.js
+++ b/docs/javascript/search.js
@@ -365,11 +365,27 @@
    * Windows sends AltGr as Ctrl+Alt: on layouts where AltGr+K is a printable
    * character, matching on Ctrl alone would swallow it.
    */
+  /*
+   * `/`, `s` and `f` were bound by Material's own search, which this replaces.
+   * Rebinding them keeps existing muscle memory working instead of silently
+   * dropping three shortcuts. They are bare keys, so they must not fire while
+   * the reader is typing into a field or editing content.
+   */
+  const TYPING_TAGS = new Set(["INPUT", "TEXTAREA", "SELECT"]);
+  const isTyping = (target) =>
+    TYPING_TAGS.has(target.tagName) || target.isContentEditable;
+
   document.addEventListener("keydown", (event) => {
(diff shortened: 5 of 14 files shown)


haranrk:
Superseded by https://github.com/google/adk-docs/pull/2096

PR #2102, opened 2026-08-10
Add Kotlin snippet for VertexAiSessionService

happyhuman:
## Summary

The `VertexAiSessionService` section of `docs/sessions/session/index.md` showed
Python, Go and Java. Kotlin gained the service in adk-kotlin 0.7.0, so this adds
a Kotlin tab and adds Kotlin to that section's language-support badge.

## Why this snippet carries so much comment

Kotlin addresses the reasoning engine differently from **every** sibling tab on
the page, in two independent ways. Both are enforced at runtime, so getting it
wrong fails at the first call rather than at compile time.

| | how the engine is supplied | format |
|---|---|---|
| Python | `app_name` on each call | full `projects/…/reasoningEngines/…` |
| Java | `app_name` on each call | bare numeric `"123456789"` |
| Go | `VertexAIService(ctx, modelName)` | n/a — different shape entirely |
| **Kotlin** | **`reasoningEngineId` at construction** | **bare numeric only** |

From the 0.7.0 KDoc:

> Unlike the Python and Java ADK, the session key's `SessionKey.appName` is
> never parsed to derive the engine — it is only a label — so the engine must
> be supplied explicitly here.

and the constructor:

```kotlin
require(reasoningEngineId.all { it.isDigit() }) {
  "reasoningEngineId must be the numeric reasoning engine id ..., not a resource name"
}
```

A reader copying the Python idiom directly above would fail twice: passing a
resource name where a numeric id is required, and expecting `appName` to select
the engine when it is inert. The snippet calls both out inline rather than
burying them here.

## Notes for reviewers

- **JVM-only.** `VertexAiSessionService` is in `core/src/jvmMain`, so it does
  not exist on Android. Kotlin is the only language on this page where that
  distinction applies, so the tab says so above the code.
- **Scoped like the Java tab**, the closest analogue: construct the service,
  then create one session. No runner or agent — no sibling tab on this page
  builds one, and this section's prose is a characteristics list rather than a
  wiring guide. An earlier revision added an `LlmAgent` + `Gemini` +
  `InMemoryRunner` and ran to 30 code lines against Python's 5 and Java's 12;
  that was misapplying a note that was correct on `sessions/memory`, where the
  prose does promise a Runner and the Python tab shows one.
- **`runBlocking` is the counterpart of the Java tab's `.blockingGet()`** —
  `createSession` is a suspend function, so some bridge is unavoidable.
- **The primary constructor is `internal`.** The snippet uses the public
  secondary one, `(project, location, reasoningEngineId, credentials,
  httpClient)`. `kotlin_api.py sig` shows only the primary, so checking the
  signature alone suggests a constructor callers cannot invoke.
- **`id = null` in `SessionKey` is deliberate.** `toAdk(appName, userId,
  fallbackId)` prefers the server-assigned id from the response `name` and
  falls back to `key.id` only when absent, so null is the correct way to let
  the service assign one. Upstream does the same.
- **No initial state.** `createSession(key, state = null)` defaults it and the
  Java tab notes none is needed, so passing one would add a concept the
  siblings avoid.
- **Section badge, not page badge.** The page-level badge already reads
  `Kotlin v0.1.0` and stays there — it marks when Kotlin support for the page
  was introduced. This `###` section has its own badge, which had no Kotlin at
  all; it gets `v0.7.0`, when the service landed.
- **Written inline** to match the two existing Kotlin tabs on this page.

## Verification

- `verify_snippets.py`: L0/L2/L3/L5/L6 pass.
- Inline snippets never reach Gradle, so this one was **additionally compiled**
  against the 0.7.0 pin in a scratch file, which is not part of the commit.
  That caught a missing required `model` argument on `LlmAgent` that L0 alone
  did not.
- API surface read from the adk-kotlin v0.7.0 sources.

## Pre-existing issue found while reviewing, not fixed here

The page's **first** Kotlin tab (the `InMemorySessionService` example, around
L142) does not compile: it calls the suspend `createSession` outside a
coroutine and has top-level statements. Out of scope for this PR, but it should
be fixed — happy to file or follow up.


wikaaaaa (approved):

joefernandez (approved):
thanks for the update!

PR #2105, opened 2026-08-10
Add Kotlin snippet for DebugLoggingPlugin

happyhuman:
## Summary

The Kotlin section of `docs/observability/logging.md` covered `LoggingPlugin`'s
console output but not `DebugLoggingPlugin`, available since adk-kotlin
**0.6.0**, which records the same activity in full to a YAML file instead of
truncated console summaries.

11 added lines in each file, matching the shape of the `LoggingPlugin`
subsection directly above it.

## Notes for reviewers

- **Transcluded, not inline.** `LoggingExamples.kt` is already registered in
  `files_to_test.txt`, so unlike an inline snippet this one is genuinely
  compiled and linted by `kotlin-snippets-pr-check.yaml` — both steps ran on
  this PR rather than being skipped.
- **The sample sets `includeSystemInstruction = false`.** The plugin's KDoc
  cautions that it writes raw prompts, tool arguments and session state to
  disk. I checked the upstream test
  (`DebugLoggingPluginTest.includeSystemInstructionFalse_marksPresenceWithoutLeakingText`)
  to confirm the flag substitutes `has_system_instruction` for the text rather
  than dropping the field. Defaulting the sample to the safer value seemed
  better than showing the permissive default.
- **`outputPath` left at its default** rather than passed explicitly, so the
  snippet doesn't imply the parameter is required. The default filename
  (`adk_debug.yaml`) is given in the prose instead.
- **Known gap, not fixed here:** adk-python ships both `logging_plugin.py` and
  `debug_logging_plugin.py`, but this page documents plugins only in the Kotlin
  section. The asymmetry predates this PR (`LoggingPlugin` was already
  Kotlin-only here) but this widens it — filed as #2106.
- **Page badge left at `Kotlin v0.1.0`.** The page's other Kotlin snippets work
  from 0.1.0, so bumping it would imply a floor that doesn't apply to them; the
  0.6.0 requirement is stated in the prose instead.
- `DebugLoggingPlugin` is in `commonJvmAndroidMain`, so it works on Android too
  — no platform caveat needed.

## Verification

- CI compiled and linted the changed `.kt`.
- `verify_snippets.py`: L0/L1/L2/L3/L5/L6 pass.
- API surface read from the adk-kotlin `v0.7.0` sources and its test.

Staged:
- https://deploy-preview-2105--adk-docs-preview.netlify.app/observability/logging/


wikaaaaa (approved):

docs/observability/logging.md
```diff
@@ -217,6 +217,17 @@ To get detailed logs of agent activity (user messages, model requests/responses,
 --8<-- "examples/kotlin/snippets/observability/LoggingExamples.kt:logging_plugin"
 ```
 
+#### Full debug capture to a file
+
```
joefernandez, line 221: use a tag under the header for language version support instead of the parenthetical comment "(adk-kotlin 0.6.0 and later)":

```
  <span class="lst-supported">Supported in ADK</span><span class="lst-kotlin">Kotlin v0.6.0</span>
</div>
```

docs/observability/logging.md
```diff
@@ -217,6 +217,17 @@ To get detailed logs of agent activity (user messages, model requests/responses,
...
+To record the same activity in full, as YAML appended to `adk_debug.yaml` rather than truncated console output, use the `DebugLoggingPlugin` (adk-kotlin 0.6.0 and later):
```
joefernandez: remove "(adk-kotlin 0.6.0 and later)" and add language tag instead

joefernandez (approved):
Approved with comment:

PR #2115, opened 2026-08-11
Document createHttpOptions on the Kotlin context cache config

happyhuman:
## Summary

`ContextCacheConfig` gained a fourth parameter, **`createHttpOptions`**, in
adk-kotlin 0.7.0. The Kotlin tab on `docs/context/caching.md` didn't show it.

4 added lines: one argument, one comment, two imports.

## What it does, and why the comment is there

It bounds the `CachedContent.create()` call specifically — and it is **fail-open**.
From the KDoc:

> When the cache creation call exceeds the timeout, it fails and the request
> proceeds without caching.

So it's a latency guard, not a correctness switch. That's the part the signature
doesn't convey, so it's the only thing the comment says.

## Notes for reviewers

- **Kotlin-first, not a backfill.** No other language documents this — Python and
  Java have no equivalent parameter at all (the Java record has exactly three
  components). So there's no sibling tab to mirror.
- **`HttpOptions` is ADK's own type**, `com.google.adk.kt.types.HttpOptions`,
  deliberately not the backend SDK's, so that cache config stays independent of
  any particular backend. Two other types share the name (Kotlin GenAI, Java
  GenAI), which makes the import line worth a glance.
- **Genuinely 0.7.0.** `HttpOptions.kt` does not exist at v0.6.0 — the class
  `ContextCacheConfig` existed, the parameter did not.
- **Not changed:** the `## Configuration settings` prose list still documents only
  three options, under their Python names (`min_tokens`, `ttl_seconds`,
  `cache_intervals`). Adding a fourth would mean a Kotlin-only bullet in a
  Python-named list, so I left it. Happy to add it if you'd rather the list were
  complete.
- **Inline**, matching the page's Python and Java tabs, so CI won't compile it.

## Verification

Extracted the snippet, compiled and ran it against the 0.7.0 pin:

```
OK timeout=10s
```

## Staged
- https://deploy-preview-2115--adk-docs-preview.netlify.app/context/caching/


wikaaaaa (approved):
It is not a Kotlin-first feature, Python supported it before Kotlin did (https://github.com/google/adk-python/blob/main/src/google/adk/agents/context_cache_config.py#L84).

So maybe the `## Configuration settings` prose should mention it as well? LGTM either way

happyhuman:
> It is not a Kotlin-first feature, Python supported it before Kotlin did (https://github.com/google/adk-python/blob/main/src/google/adk/agents/context_cache_config.py#L84).
> 
> So maybe the `## Configuration settings` prose should mention it as well? LGTM either way

I added a sentence to address this. PTAL.

wikaaaaa (approved):

joefernandez (approved):

PR #2116, opened 2026-08-11
Add Kotlin snippet for Vertex AI model routing

happyhuman:
## Summary

`routingConfig` and `GenerateContentConfig` labels landed in adk-kotlin **0.6.0**
and were undocumented — the docs had **no mention of routing config in any
language**. This adds a `### Route requests across models` subsection to
`docs/agents/llm-agents.md`, transcluded from the existing `CapitalAgent.kt`.

## Why a new subsection, not the existing Kotlin tab

The `Fine-tune AI model operation` group is about generation *parameters* —
`temperature`, `maxOutputTokens`. Routing picks a *model* rather than tuning one,
so it reads as a separate concern. Extending that tab would also mean editing a
group that already has Kotlin.

## Notes for reviewers

- **Both features are Vertex-only, and the snippet has to show that, not just say
  it.** `routingConfig` is documented as *"Not supported in the Gemini API"*, and
  `labels` are silently stripped by `sanitizeForGeminiApi()` when calling with an
  AI Studio API key — a caller would see them vanish with no error. So the model
  is configured with `VertexCredentials`.
- **Badged `Kotlin v0.6.0`, and I tested that rather than assuming it.** The docs
  pin is 0.7.0, so the easy move was to badge 0.7.0. Instead I extracted the
  region and ran it against a temporary **0.6.0** pin as well:

  | pin | result |
  |---|---|
  | 0.6.0 | `OK pref=PRIORITIZE_COST labels={team=search}` |
  | 0.7.0 | `OK pref=PRIORITIZE_COST labels={team=search}` |

  A reader on 0.6.0 can use this exactly as written, so 0.6.0 is the honest floor.
- **`manualMode` is named in the prose but not shown.** It's a one-field
  alternative (`modelName`); showing both modes would double the snippet for
  little gain.
- **The subsection is Kotlin-only.** Precedented on this very page — the next
  section, `Configure a default model`, carries a Python-only badge.
- **Transcluded** from `CapitalAgent.kt`, already registered, so CI compiles and
  lints it.

## Verification

- CI compiles and lints the changed `.kt`.
- Ran the region standalone: constructing a Vertex-backed `Gemini` needs no
  credentials at build time, so the snippet loads without crashing.
- `verify_snippets.py` passes all six levels.

### DIFF
--- a/docs/agents/llm-agents.md
+++ b/docs/agents/llm-agents.md
@@ -421,6 +421,20 @@ You can adjust how the underlying AI model generates responses using
     --8<-- "examples/kotlin/snippets/agents/llm-agent/CapitalAgent.kt:gen_config"
     ```
 
+### Route requests across models
+
+<div class="language-support-tag" title="">
+   <span class="lst-supported">Supported in ADK</span><span class="lst-kotlin">Kotlin v0.6.0</span>
+</div>
+
+`routingConfig` asks Vertex AI to choose the model for each request, rather than
+pinning one. Use `autoMode` with a `ModelRoutingPreference` to let Vertex balance
+quality against cost, or `manualMode` to name a model explicitly.
+
+```kotlin
+--8<-- "examples/kotlin/snippets/agents/llm-agent/CapitalAgent.kt:routing_config"
+```
+
 ### Configure a default model
 
 <div class="language-support-tag" title="">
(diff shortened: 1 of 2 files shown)


wikaaaaa:
I'd suggest we skip this snippet, for a two reasons:

1. the current snippet woudn't work, as in order for the routing to happen the specified model endpoint needs to be a specific model routing endpoint, we dont have any examples for this in our codebase

2. **It's a genai passthrough, not an ADK feature, and it's uniformly undocumented.** `routingConfig` rides on `GenerateContentConfig`, which is supported identically in Python, Java, and Kotlin (via `generate_content_config`) — and documented in none of them.

PR #2118, opened 2026-08-11
Add A2A consuming quickstart for Kotlin

happyhuman:
## Summary

adk-kotlin has been able to **consume** remote A2A agents since 0.6.0, and the
docs had no Kotlin page for it. This adds `docs/a2a/quickstart-consuming-kotlin.md`
alongside the Python, Go and Java quickstarts, with its nav entry, a transcluded
snippet, and the two dependencies it needs.

## A new page, not a tab

`docs/a2a/` has **zero** tab groups — it is one page per language
(`quickstart-consuming{,-go,-java}.md`). So a Kotlin page follows the section's
own shape rather than breaking a pattern.

## The dependency set: two, not one — established by compiling

The a2a artifact alone is **not enough**. It publishes the A2A SDK as
runtime-only (`jvmApiElements-published` exposes only `kotlin-stdlib`), while
`A2AAgent`'s `httpClient` parameter defaults to `JdkA2AHttpClient()` — so the
type must resolve at compile time:

```
e: Cannot access class 'A2AHttpClient'. Check your module classpath ...
```

Adding `a2a-java-sdk-client` fixes it, and nothing further is required — `spec`
and the JSON-RPC transport arrive transitively. I verified that by building, not
by reading module metadata, and it contradicts an earlier estimate of four
artifacts.

## Notes for reviewers

- **Consuming only, because that's all that exists.** No `webserver/**` source at
  v0.7.0 mentions a2a, so there is no Kotlin equivalent of the *exposing*
  quickstarts. The page states this and links to the Python and Java ones rather
  than leaving a dead end.
- **`A2AAgent` is a suspending factory**, and `A2AAgentImpl` is `internal`, so the
  factory is the only way to build one. There's also a non-suspending overload
  taking an `AgentCard` directly, which the page mentions.
- **Badge is unversioned** (`Kotlin`, not `Kotlin v0.6.0`), matching every other
  page in this section — and avoiding a contradiction with the `0.7.0` dependency
  the page tells you to add.
- **Transcluded and registered**, so CI compiles and lints the snippet.

## Verification

Not just compiled — **run end to end**. I served a real agent card from a local
server at the well-known path and executed the snippet:

```
GET /.well-known/agent-card.json
Root agent root_agent delegates to prime_agent
```

So the factory fetched the card, parsed it, and the remote agent was accepted as
a sub-agent. `verify_snippets.py` passes all six levels.


docs/a2a/quickstart-consuming-kotlin.md
```diff
@@ -0,0 +1,73 @@
...
+will do, in any language. The A2A protocol requires each agent to publish an
+**agent card** describing what it does, served at the well-known path:
+
+```text
+http://localhost:9090/.well-known/agent-card.json
+```
+
+## Connect to the remote agent
+
+`A2AAgent` fetches that card and reads the remote agent's name, description and
```
wikaaaaa: Technically we dont fetch the name, the name is given by the user as an argument to A2AAgent, so maybe just  `A2AAgent fetches that card from the URL and reads the remote agent's description from it.` or sth similar. To be changed here and in the code comment.

examples/kotlin/snippets/a2a/A2AConsumer.kt
```diff
@@ -0,0 +1,55 @@
...
+        // --8<-- [start:root_agent]
+        // The remote agent is a BaseAgent, so it goes in subAgents like any local one.
+        // ADK handles the A2A wire protocol from here.
+        val rootAgent =
+            LlmAgent(
+                name = "root_agent",
+                model = Gemini(name = "gemini-flash-latest"),
+                instruction =
+                    Instruction(
+                        "Roll dice yourself, and delegate prime checks to prime_agent.",
```
wikaaaaa: Nit: root_agent has no dice-rolling tool, so "Roll dice yourself" points at nothing. prime_agent is the only sub-agent here, so the instruction should cover just the prime delegation, matching the Java snippet's "You are a helpful assistant that can check prime numbers by delegating to prime_agent." ([A2aConsumerSnippet.java:49](https://github.com/google/adk-docs/blob/main/examples/java/snippets/src/main/java/a2a/A2aConsumerSnippet.java#L49C27-L49C113))

wikaaaaa (approved):

happyhuman:
Both fixed, thanks. On the card: you're right the name is ours, and the transport claim was wrong too — it's hardcoded to JSON-RPC. The card only supplies the description and streaming capability. Updated the page and the code comment. Instruction now matches the Java snippet's prime-delegation wording.

wikaaaaa (approved):

joefernandez:
### Technical review report

**Verdict: technically correct, one factual error to fix suggestion.**

## What checks out

Every API claim matches adk-kotlin v0.7.0 — the version the page pins:

- Suspending factory `A2AAgent(name, agentCardUrl)` — [A2AAgent.kt#L89-L107](https://github.com/google/adk-kotlin/blob/v0.7.0/a2a/src/jvmMain/kotlin/com/google/adk/kt/a2a/jvm/A2AAgent.kt#L89-L107)
- `httpClient` defaults to `JdkA2AHttpClient()` — [#L92](https://github.com/google/adk-kotlin/blob/v0.7.0/a2a/src/jvmMain/kotlin/com/google/adk/kt/a2a/jvm/A2AAgent.kt#L92)
- Non-suspending `agentCard` overload — [#L50-L58](https://github.com/google/adk-kotlin/blob/v0.7.0/a2a/src/jvmMain/kotlin/com/google/adk/kt/a2a/jvm/A2AAgent.kt#L50-L58)
- `A2AAgentImpl` is `internal` — [A2AAgentImpl.kt#L59](https://github.com/google/adk-kotlin/blob/v0.7.0/a2a/src/commonJvmAndroidMain/kotlin/com/google/adk/kt/a2a/agent/A2AAgentImpl.kt#L59)
- Description + streaming read from the card — [A2AAgentImpl.kt#L83-L85](https://github.com/google/adk-kotlin/blob/v0.7.0/a2a/src/commonJvmAndroidMain/kotlin/com/google/adk/kt/a2a/agent/A2AAgentImpl.kt#L83-L85)
- Well-known path + `Failed to parse agent card` — [AgentCardResolver.kt#L37-L64](https://github.com/google/adk-kotlin/blob/v0.7.0/a2a/src/commonJvmAndroidMain/kotlin/com/google/adk/kt/a2a/agent/AgentCardResolver.kt#L37-L64)

The two-dependency claim — the page's least obvious assertion — is right, and I confirmed the negative rather than trusting it. The A2A SDK is declared `implementation`, never `api` ([a2a/build.gradle.kts#L42-L44](https://github.com/google/adk-kotlin/blob/v0.7.0/a2a/build.gradle.kts#L42-L44)), and the published `.module` exposes only `kotlin-stdlib` on `jvmApiElements`. Deleting the client dep and rebuilding reproduces the quoted error verbatim at `A2AConsumer.kt:34:13`. I also ran the snippet end to end against a served card and got the PR's claimed output, and confirmed the adk-java sample card fails exactly as the page warns.

## Fix recommendations

The page says the `remote_a2a` sample in **adk-python** doesn't work as a server "as they stand today." It does. adk-python doesn't serve its `agent.json` verbatim — `fast_api.py` routes it through `_compat.parse_agent_card`, which under a2a-sdk 1.x up-converts legacy `url`/`preferredTransport` into `supportedInterfaces`. Since the pin is `a2a-sdk>=0.3.4,<2`, a fresh install gets that. I applied the transform to the actual sample card and ran the PR's own snippet against it: `Root agent root_agent delegates to prime_agent`.

The adk-java half of that sentence is correct (pinned to A2A SDK 0.3.x). Fixing this also improves the page — adk-python becomes a real server to point at instead of a hand-written JSON file, via `agentCardUrl = "http://localhost:8001/a2a/check_prime_agent"`.

Two non-blocking notes: `a2a-java-sdk-http-client` is the strictly minimal artifact (but `a2a-java-sdk-client` is still the better recommendation), and v0.8.0 was tagged today though its artifacts aren't on Maven Central yet — so `0.7.0` is currently the correct pin.

docs/a2a/quickstart-consuming-kotlin.md
```diff
@@ -0,0 +1,110 @@
...
+```text
+http://localhost:9090/.well-known/agent-card.json
+```
+
+The Kotlin client reads **A2A 1.0** cards, so the card must carry a
+`supportedInterfaces` array whose entries each have a `protocolBinding`.
+Cards written for A2A 0.3 omit it, and `A2AAgent` rejects them with
+`AgentCardResolutionError: Failed to parse agent card`. That includes the
+`a2a_server` sample in adk-java and the `remote_a2a` sample in adk-python as
+they stand today, so neither works as the server for this page yet.
```
joefernandez: see technical review report comment for possible correction, here.

joefernandez (approved):
Approved with comment.

happyhuman:
Fixed — adk-python is the server now, adk-java note kept.

On the pin: a2a 0.8.0 hit Maven Central since your note; #2143 bumps it.

PR #2120, opened 2026-08-11
Document how to check whether the context cache was used

happyhuman:
## Summary

`docs/context/caching.md` explains how to turn caching **on** and never how to
tell whether it's **working**. This adds a `## Check whether the cache is being
used` section built on `CacheMetadata`, available since adk-kotlin 0.6.0.

## Why this is a real gap

adk-python has **14** code references to `cache_metadata`. adk-docs mentions it
**twice**, both incidental — a BigQuery schema table
(`integrations/bigquery-agent-analytics.md:829`) and a bullet in the Live dev
guide (`live/dev-guide/part3.md:186`). Nobody documents how to read it.

## Notes for reviewers

- **The two states aren't stylistic, they're enforced.** `CacheMetadata`'s `init`
  block requires that `cacheName`, `expireTime` and `invocationsUsed` are either
  *all* set (an active cache) or *all* null (the fingerprint-only state used for
  prefix matching before a cache exists). The snippet branches on `isActive`
  rather than null-checking one field, which is why.
- **Token counts are deliberately absent** from `CacheMetadata` — the KDoc says
  to read them from `LlmResponse.usageMetadata` to avoid duplication. A reader
  would reasonably look for them here, so the page says where they are.
- **Badged `Kotlin v0.6.0`, verified not assumed** — the snippet compiles against
  a temporary 0.6.0 pin as well as the current 0.7.0 one.
- **Kotlin-only section**, like the routing one in #2116. Python and Java expose
  the concept but no ADK docs cover reading it in any language.
- **Transcluded and registered**, so CI compiles and lints it.

## Verification

All four paths exercised with synthetic events:

```
-- no metadata at all:            (silent, early return)
-- fingerprint-only:              Not cached yet; fingerprinted 3 contents.
-- active cache, far from expiry: Cache .../cachedContents/9 reused 4 time(s).
-- active cache, about to expire: Cache .../cachedContents/9 reused 9 time(s).
                                  Cache expires shortly; the next turn will create a new one.
```

## Note on overlap

#2115 also touches `caching.md`, but edits the existing Kotlin tab (~L69-100);
this adds a new section near the end. No textual overlap.


docs/context/caching.md
```diff
@@ -117,6 +117,23 @@ all agents within your app.
     control how frequently the cache is updated, even if the TTL has not
     expired. Defaults to `10`.
 
+## Check whether the cache is being used
+
+<div class="language-support-tag">
+   <span class="lst-supported">Supported in ADK</span><span class="lst-kotlin">Kotlin v0.6.0</span>
+</div>
+
+Every event backed by an LLM response carries a `CacheMetadata`, which reports
```
wikaaaaa: nit: CacheMetadata will be null when context cashing is disabled, maybe

"When caching is enabled, events backed by an LLM response carry..."

but then this is a section about context cache so maybe its obvious that it is enabled

wikaaaaa (approved):

happyhuman:
Good catch — you're right, and it turned out to be a bit more than a nit.

I went to check exactly when `cacheMetadata` is null, and the KDoc on `LlmResponse.cacheMetadata` says it's "populated when context caching is enabled. `null` when caching is disabled or no cache information is available." So it can be null even with caching switched on, which means your suggested wording would still have been slightly optimistic. I've gone with "can carry" and named both null cases, so the `?: return` in the snippet has a visible reason to exist now.

While I was in there I found something worse a few lines down. I had this:

```kotlin
if (cache.expireSoon) {
    println("Cache expires shortly; the next turn will create a new one.")
}
```

The second half is just wrong. `expireSoon` is a 2-minute buffer, but nothing in the framework actually reads it — grep it in 0.7.0 and all you get is its own definition and its unit tests. The real reuse check in `isCacheValid` is a hard `now >= expireTime`, so a cache that `expireSoon` flags gets happily reused on the next turn anyway. And `expireSoon` stays true *after* expiry too (there's a test literally called `expireSoon_pastExpiry_isTrue`), so "expires shortly" was wrong in that case as well.

It now says "Cache is at or near expiry", and the page has a line explaining it's advisory only, along with what does actually cause a new cache: past `expireTime`, past `cacheIntervals`, or a changed prefix.

I also tightened the fingerprint-only comment. I'd listed two causes as though that were the full set, but `handleContextCaching` has three, and cache creation bails for more reasons than just the prefix being too small.

Embarrassingly, my "verification" in the description had exercised that `expireSoon` branch and printed the sentence — which proved the code ran, not that what it said was true. So this time I actually ran `logCacheUse` against events in all five states (no metadata, fingerprint-only, active far from expiry, active inside the buffer, active past expiry) instead of just compiling it.

Badge stays at v0.6.0 — I checked the tag and `CacheMetadata`, `isActive`, `expireSoon` and `Event.cacheMetadata` are all present there.

wikaaaaa (approved):

docs/context/caching.md
```diff
@@ -117,6 +117,30 @@ all agents within your app.
...
+caching is disabled, and also when the call produced no cache information, so
+check for it before reading it. When present it has two states: an **active
+cache**, where `cacheName`, `expireTime` and `invocationsUsed` are all set, and
+a **fingerprint-only** state, where all three are null.
+
+```kotlin
+--8<-- "examples/kotlin/snippets/context/CacheMetadataExample.kt:cache_metadata"
+```
+
+`expireSoon` means the cache expires within about two minutes, or has already
```
joefernandez, line 143: Don't start sentences with code syntax, and be more explicit about the value meaning:

An `expireSoon` value of `True` means the

docs/context/caching.md
```diff
@@ -117,6 +117,30 @@ all agents within your app.
...
+expired. It is a signal for your own code, not something ADK acts on: ADK keeps
+reusing a cache until it is actually past `expireTime`, has run past
+`cacheIntervals`, or its cached prefix changes.
+
+Token counts are not on `CacheMetadata`; read them from `LlmResponse.usageMetadata`.
```
joefernandez, line 148: Remove this sentence. While this might be useful information, adding it here makes it unlikely to be discovered.

joefernandez (approved):
Approved with comments

PR #2121, opened 2026-08-11
Add Kotlin snippet for the CallbackContext memory-write helpers

happyhuman:
## Summary

adk-kotlin 0.7.0 added `addEventsToMemory` and `addMemory` to `CallbackContext`,
alongside the `addSessionToMemory` that `docs/sessions/memory.md` already shows
in Kotlin. This adds a subsection covering the two cases the existing snippet
can't: a chosen **subset of events**, and **facts you construct yourself**.

## Compiling changed the snippet's shape

The obvious version read the current turn's events off the context:

```kotlin
val thisTurn = context.invocationContext.session.events.filter { ... }   // ✗
```

That doesn't compile — `CallbackContext.invocationContext` is **`internal`**, and
the class exposes no other route to the session. So `addEventsToMemory` takes
events the caller already holds, and the snippet's KDoc says so plainly rather
than leaving a reader to discover it. Worth knowing when reviewing: the ergonomic
gap is real, not a documentation choice.

## Notes for reviewers

- **All three helpers throw `IllegalStateException`** when the runner has no
  memory service — a runtime failure with nothing at compile time to warn you, so
  the page states it. Verified both messages by running them:
  `Cannot add events to memory: memory service is not available.` and
  `Cannot add memory: memory service is not available.`
- **Badged `Kotlin v0.7.0`, verified not assumed** — neither helper exists on
  `CallbackContext` at v0.6.0 (0 of 2 present).
- **Kotlin-only subsection.** The service-level equivalents are covered for Python
  under *Extend memory capabilities*; these are the callback-side convenience
  wrappers, which only Kotlin has.
- **Appended to the existing `MemoryExample.kt`**, already registered, so CI
  compiles and lints it.

## Verification

```
OK addEventsToMemory succeeded with a memory service
OK addMemory succeeded with a memory service
without a memory service -> IllegalStateException: Cannot add memory: ...
addEventsToMemory w/o service -> IllegalStateException: Cannot add events to memory: ...
```


wikaaaaa (approved):

docs/sessions/memory.md
```diff
@@ -597,6 +597,25 @@ For example, you can automate this step with a callback:
     --8<-- "examples/kotlin/snippets/sessions/MemoryExample.kt:auto_save_callback"
     ```
 
+### Write specific events or facts from a callback
+
+   <span class="lst-supported">Supported in ADK</span><span class="lst-kotlin">Kotlin v0.7.0</span>
+</div>
+
+`addSessionToMemory` saves the whole session. When you want finer control,
```
joefernandez: don't start sentences with code syntax, and provide more context when you lead into a topic:

The `CallbackContext.addSessionToMemory` method is the default behavior for memory and saves the whole session of your agent.

joefernandez (approved):
Approved with comment

PR #2128, opened 2026-08-13
docs(integrations): add tracegauge Cost Evaluator for ADK agents

gaurav-gandhi-2411:
## What & why

Adds [`adk-tracegauge`](https://pypi.org/project/adk-tracegauge/) to the integrations catalog. It registers a `google.adk.evaluation` metric (`adk_tracegauge_cost_usd`) that reports real per-invocation dollar cost, built on [tracegauge](https://github.com/gaurav-gandhi-2411/token-efficiency-scorer)'s cost engine. ADK's built-in evaluation metrics cover quality (trajectory match, response similarity, safety, hallucination) but none report cost or token usage — this fills that gap.

## Changes

- `docs/integrations/adk-tracegauge.md`: new integration page, following the standard template (frontmatter + Use cases / Prerequisites / Installation / Use with agent / Available metrics / Resources). Closest in shape to the existing [MLflow Scorers](https://adk.dev/integrations/mlflow-scorers/) entry — the only other catalog integration that also registers against ADK's `MetricEvaluatorRegistry`. `catalog_tags: ["evaluation"]` matches that precedent.
- `docs/integrations/assets/adk-tracegauge.png`: 512×512 square icon.

## Testing

- Frontmatter parses as valid YAML; code fences balanced; icon confirmed square PNG.
- The "Use with agent" code example was executed against the actual published `adk-tracegauge==0.1.0` package (not copy-pasted untested).
- Package itself: 47 tests, ruff clean, mypy strict clean, published via PyPI Trusted Publishing (OIDC); post-publish verified from a clean environment against the real PyPI index.

## Risk & rollback

New page, additive only — doesn't touch any existing catalog entry or navigation config (individual integration pages aren't listed in `mkdocs.yml`; the catalog index auto-renders from `docs/integrations/*.md`). Revert is a two-file delete.


gaurav-gandhi-2411:
Correction to this PR's original "Use with agent" instructions, found and fixed before this could reach anyone following the docs.

The original page instructed wiring `TraceGaugeUsagePlugin` into an `App` and said `AgentEvaluator` "picks up this `app`... automatically." That claim doesn't hold against `google-adk==2.6.3`: `AgentEvaluator._get_agent_for_eval`, `EvaluationGenerator._process_query`/`_generate_inferences_from_root_agent`, and the `adk eval` CLI's own `get_root_agent` all resolve only `root_agent`/`get_agent_async` from the agent module and build their own internal `Runner` from a bare `root_agent` — none of them ever look at an `App` or its `plugins`. Live-verified: `AgentEvaluator.evaluate()` raises `AssertionError` unconditionally when this metric is registered, and `adk eval` runs without crashing but discards the metric's per-invocation score and rationale (`score: null` in both the printed table and the persisted `eval_history/*.evalset_result.json`).

Digging into why surfaced a second, independent issue in `LocalEvalService` itself: it discards per-invocation results for any metric whose `eval_status` is `NOT_EVALUATED` (`local_eval_service.py:428-436`) — which is this metric's permanent, by-design status, since cost is lower-is-better and has no honest fit in ADK's `score >= threshold -> PASSED` convention. Filed as a design question upstream, with a minimal self-contained repro: [google/adk-python#6725](https://github.com/google/adk-python/issues/6725).

I've pushed a revision to this PR that:
- Removes the "Use with agent" instructions that crash, and states the limitation up front with a link to the filed issue.
- Documents the hand-rolled `Runner` harness that's actually the working integration path today, with a pointer to the package README's full worked example.
- Presents an `after_model_callback`-based partial workaround with its real limits stated (still doesn't fix `AgentEvaluator.evaluate()`; still no per-invocation detail).

The package itself (`adk-tracegauge`) also shipped a `0.2.0` with the same reframing — README leads with this limitation now, diagnostics moved to `warnings.warn` since rationale text can't survive `LocalEvalService`'s discarding either way.

Happy to adjust tone/length/placement to fit the catalog's normal style — flag anything that reads oddly for the format.

joefernandez:
Thanks for this integration contribution. We would like to see this integration grow more adoption and maturity before we list it as an ADK integration.

PR #2145, opened 2026-08-19
Fix tool name in TypeScript documentation

luizgribeiro:
Fixes tool name in agent instructions

### DIFF
--- a/docs/get-started/typescript.md
+++ b/docs/get-started/typescript.md
@@ -75,7 +75,7 @@ export const rootAgent = new LlmAgent({
   model: 'gemini-flash-latest',
   description: 'Tells the current time in a specified city.',
   instruction: `You are a helpful assistant that tells the current time in a city.
-                Use the 'getCurrentTime' tool for this purpose.`,
+                Use the 'get-current-time' tool for this purpose.`,
   tools: [getCurrentTime],
 });
 ```


docs/get-started/typescript.md
```diff
@@ -75,7 +75,7 @@ export const rootAgent = new LlmAgent({
   model: 'gemini-flash-latest',
   description: 'Tells the current time in a specified city.',
   instruction: `You are a helpful assistant that tells the current time in a city.
-                Use the 'getCurrentTime' tool for this purpose.`,
+                Use the 'get-current-time' tool for this purpose.`,
```
joefernandez: L0 RISK REVIEW (Medium): The tool is registered as `name: 'get_current_time'` at [line 63](https://github.com/google/adk-docs/blob/3daea8d2ffbfd7ddbd15078b08f8110a2cfc80a4/docs/get-started/typescript.md#L63), so this instruction still names a tool the model does not have — hyphens instead of underscores. The original `'getCurrentTime'` was indeed wrong (that is the const identifier); this is the spelling that matches the registration, and the Python quickstart at [python.md#L84](https://github.com/google/adk-docs/blob/3daea8d2ffbfd7ddbd15078b08f8110a2cfc80a4/docs/get-started/python.md#L84).

```suggestion
                Use the 'get_current_time' tool for this purpose.`,
```

joefernandez (commented):
# Content review report

- **Review Score**: 97
- **Overall assessment**: 🟡 Medium Risk
- **Summary reasoning**: A one-line docs fix with no legal, privacy, security, pricing, performance, or third-party claims. The only risk is accuracy: the replacement tool name is a third spelling that matches neither the registered tool name nor the variable name, so the quickstart still gives readers a broken instruction.
- **Recommended action**: Approve with these specific changes. Change `'get-current-time'` to `'get_current_time'` on line 78 so the instruction matches the `name:` the tool is registered with; optionally fold in the prose on line 54 while you are here.
- **Number of lines changed**: XS, 1
- **Number of files changed**: 1

## 🟡 Medium risk items

- **The corrected tool name is still wrong — hyphens instead of underscores.** The tool is registered as `name: 'get_current_time'` at [docs/get-started/typescript.md#L63](https://github.com/google/adk-docs/blob/3daea8d2ffbfd7ddbd15078b08f8110a2cfc80a4/docs/get-started/typescript.md#L63), but [line 78](https://github.com/google/adk-docs/blob/3daea8d2ffbfd7ddbd15078b08f8110a2cfc80a4/docs/get-started/typescript.md#L78) now says `Use the 'get-current-time' tool`. You were right that the original `'getCurrentTime'` was the variable name rather than the tool name, but the replacement names a tool the model was never given — the same class of bug the PR set out to fix. Suggested fix inline below.

## 🟢 Low risk items

- **Pre-existing inconsistency worth folding in.** The prose at [line 54](https://github.com/google/adk-docs/blob/3daea8d2ffbfd7ddbd15078b08f8110a2cfc80a4/docs/get-started/typescript.md#L54) still calls the tool `getCurrentTime`, which is the const identifier, not the registered tool name. That line is outside this diff so it cannot carry a suggestion, but since the PR is about tool-name accuracy it would be cheap to change it to ``called `get_current_time` `` in the same commit, leaving one name in the section instead of two.

Both changes line the TypeScript quickstart up with the Python one, which already uses `'get_current_time'` in its instruction at [docs/get-started/python.md#L84](https://github.com/google/adk-docs/blob/3daea8d2ffbfd7ddbd15078b08f8110a2cfc80a4/docs/get-started/python.md#L84).

joefernandez:
Replaced with: https://github.com/google/adk-docs/pull/2256

PR #2147, opened 2026-08-19
Document the BigQuery agent analytics plugin for Kotlin

happyhuman:
## Summary

`docs/integrations/bigquery-agent-analytics.md` had no Kotlin content.
adk-kotlin **0.8.0** ships `BigQueryAgentAnalyticsPlugin`, so the Quickstart's
setup group now carries a Kotlin tab beside Python and Java, backed by a
transcluded snippet that CI compiles and lints.

Only the setup group gets Kotlin. The page's six other tab groups document event
payloads and configuration surface the Kotlin plugin does not have.

## The interesting part: saying what Kotlin *doesn't* do

This page describes the Python and Java plugins, and a Kotlin badge silently
extends every one of its promises to Kotlin. It shouldn't. At the pin,
`BigQueryAgentAnalyticsPlugin` overrides exactly two `Plugin` callbacks, so the
page's opening — Auto Schema Upgrade, tool provenance, HITL tracing, automatic
view creation, ADK 2.0 workflow events, drop stats — describes nothing Kotlin
implements.

Rather than bury that inside the Kotlin tab, where a reader on the Python tab
never renders it, there is a page-level **"Kotlin support"** note next to the
pricing warning, following the `!!! note "Java support"` idiom this page already
uses further down. It records four things a reader would otherwise get wrong:

- **Two event types.** `INVOCATION_STARTING` and `INVOCATION_COMPLETED` only;
  identity columns plus `content` populated, with `trace_id`, `span_id`,
  `latency_ms` and `attributes` left null.
- **No views.** `BigQuerySchema` has only `getEventsSchema()` and
  `getDefaultClusteringFields()`. The `v_*` names in the captured-events table
  do not exist for Kotlin, so `SELECT … FROM v_invocation_completed` would have
  returned a not-found.
- **A different ingestion path, and a different bill.** Kotlin calls
  `tabledata.insertAll`, one row at a time, synchronously on the invocation
  path — not the gRPC Storage Write API the page describes and links pricing
  for. That is a separate SKU: charged per inserted row with a 1 KB minimum and
  no monthly free tier, so cost tracks invocation count rather than bytes.
- **Failures are swallowed.** A table that cannot be created, or a row that
  cannot be inserted, is logged and the turn continues — a misconfigured agent
  looks perfectly healthy while writing nothing.

The Kotlin tab adds the artifact coordinate, the full
`BigQueryLoggerConfig` surface (the [Configuration
options](#configuration-options) section is Python and Java only), and the class
name to raise the log level on when rows go missing.

`location` deserves a specific mention: it defaults to `"US"` and is passed
straight to the BigQuery client, and nothing in the page would have told a
Kotlin reader with an `EU` dataset that the parameter exists. The snippet takes
it as an argument instead of pinning a `tableName` that already matches its own
default.

## Build change

One line: `com.google.adk:google-adk-kotlin-integrations:0.8.0`. Unlike the a2a
artifact, this one publishes `google-cloud-bigquery` and `google-auth` on
`jvmApiElements`, so the BigQuery types named by the plugin constructor's
defaults land on the compile classpath without a second dependency. Confirmed by
a real compile, not just by reading the module metadata.

## Verification

Everything below was actually run:

- `./tools/kotlin-snippets/runner.sh build …/BigQueryAnalyticsExample.kt` →
  **PASS** (JDK 17). Same command CI runs.
- `./tools/kotlin-snippets/runner.sh lint …` → **PASS**.
- `check_kotlin_snippets.sh` → all Kotlin files registered.
- `verify_snippets.py` → **L0 symbols, L1 compile, L2 ktlint, L3 transclusions,
  L5 registration, L6 badge all PASS.** L4 skips: this project registers no
  `runSnippets` task.
- Rendered the page through `python-markdown` with the extension list parsed out
  of `mkdocs.yml`: the Quickstart set's labels come back
  `['Python', 'Java', 'Kotlin']` and `Kotlin support` renders as an admonition,
  so neither the tab nor the note leaks raw markdown.

Every API claim is grounded in the `v0.8.0` tag of `google/adk-kotlin`, never the
working tree, which runs ahead of the pin.

## Not fixed here

The [IAM permissions](#iam-permissions) step grants `roles/bigquery.dataEditor`
at **Table Level**, which cannot create a table that does not exist yet — while
Prerequisites says the plugin creates the events table for you. That is wrong for
all three languages and predates this change, so I left it alone. Happy to fix it
in a follow-up.


wikaaaaa (approved):

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -7,8 +7,8 @@ catalog_tags: ["observability", "google"]
 
 # BigQuery Agent Analytics plugin for ADK
 
-<div class="language-support-tag">
-  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v1.21.0</span><span class="lst-java">Java v1.5.0</span>
+<div class="language-support-tag" title="Kotlin support covers invocation lifecycle logging only; see the Kotlin support note below.">
```
joefernandez: this title popup is too hidden / non-obvious for real notification. Remove.

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -58,6 +58,22 @@ The plugin includes three reliability and observability fixes:
     For information on costs, see the [BigQuery
     documentation](https://cloud.google.com/bigquery/pricing?e=48754805&hl=en#data-ingestion-pricing).
 
+!!! note "Kotlin support"
```
joefernandez: make this expandable

??? note

joefernandez (approved):
(see below)

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -58,6 +58,22 @@ The plugin includes three reliability and observability fixes:
     For information on costs, see the [BigQuery
     documentation](https://cloud.google.com/bigquery/pricing?e=48754805&hl=en#data-ingestion-pricing).
 
+??? note "Kotlin support"
+
+    The **Kotlin** plugin supports subset of available features. It logs
```
joefernandez: Grammar: "supports subset of available features" → "supports a subset of the available features".

While here — "supports a subset" is doing a lot of vague work for what is actually a precise list. Consider leading with the concrete statement ("logs `INVOCATION_STARTING` and `INVOCATION_COMPLETED` only") and dropping the abstract preamble.

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -58,6 +58,22 @@ The plugin includes three reliability and observability fixes:
...
+    `INVOCATION_STARTING` and `INVOCATION_COMPLETED` only; it fills the identity
+    columns and `content`, leaving `trace_id`, `span_id`, `latency_ms`,
+    `attributes` and the rest null; and it creates **no views**, so the `v_*`
+    views in the table below do not exist for Kotlin. Auto Schema Upgrade, tool
+    provenance, HITL tracing, drop stats and the ADK 2.0 workflow events are not
+    implemented.
```
joefernandez: **This contradicts the table below without editing it.**

"it creates **no views**, so the `v_*` views in the table below do not exist for Kotlin" — the table still documents those views with no Kotlin qualification. The page now says two different things in two places, and the more prominent one is the wrong one.

Same pattern applies to the null-column claim: `trace_id`, `span_id`, `latency_ms` and `attributes` are documented above as though they are populated.

Please qualify the table rows directly rather than describing them as wrong from a separate block. "the table below" is also a fragile reference — it breaks if the page is ever reordered or if a reader arrives via anchor link.

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -58,6 +58,22 @@ The plugin includes three reliability and observability fixes:
...
+
+    It also ingests differently: rows go one at a time through
+    `tabledata.insertAll`, synchronously on the invocation path, not through the
+    gRPC Storage Write API described above. Those are separate billing lines:
+    inserted rows are charged with a 1 KB minimum each and no monthly free tier,
+    so cost scales with invocation count rather than bytes.
```
joefernandez: **Pricing specifics don't belong in ADK docs.**

Remove:
"Those are separate billing lines:
    inserted rows are charged with a 1 KB minimum each and no monthly free tier,
    so cost scales with invocation count rather than bytes."

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -193,6 +209,37 @@ shows the BigQuery view optionally created when
...
+    ```kotlin title="BigQueryAnalyticsExample.kt"
+    --8<-- "examples/kotlin/snippets/integrations/BigQueryAnalyticsExample.kt:quickstart"
+    ```
+
+    `BigQueryLoggerConfig` is the whole Kotlin configuration surface —
+    `projectId`, `datasetId`, `enabled` (default `true`), `location` (default
+    `"US"`, passed to the BigQuery client), `tableName` (default
+    `"agent_events"`) and `credentials` (default: application default
+    credentials). The options under [Configuration
+    options](#configuration-options) are Python and Java only.
```
joefernandez: **This reclassifies an existing shared section from inside a Kotlin tab.**

"The options under [Configuration options](#configuration-options) are Python and Java only" changes what that section means for every reader — but the section itself is untouched and still presents as language-neutral. Anyone who lands on `#configuration-options` directly, or reads top-to-bottom without opening the Kotlin tab, gets no signal.

If those options really are Python/Java-only, that scoping belongs in the Configuration options section itself. A Kotlin tab shouldn't be the only place the rest of the page's scope is defined.

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -193,6 +209,37 @@ shows the BigQuery view optionally created when
...
+
+    **Logging failures are swallowed.** If the table cannot be created or a row
+    cannot be inserted, the plugin logs and the invocation continues, so a
+    misconfigured agent looks healthy while writing nothing. When rows are
+    missing, raise the log level for
+    `com.google.adk.kt.plugins.agentanalytics.BigQueryAgentAnalyticsPlugin` —
+    logs are emitted under that class name, not under the plugin's ADK name.
+    Note also that Kotlin writes `content` as
+    `{"message": "Invocation started"}` rather than the `{}` shown for these two
+    event types below.
```
joefernandez: **Register: this reads as a known-issues list rather than documentation.**

Between this paragraph and the block at line 61, the Kotlin content is roughly: eight unimplemented features, a warning that failures are silent ("a misconfigured agent looks healthy while writing nothing"), a logger-name gotcha, and a payload that differs from what the page documents. A reader comparing tabs will conclude Kotlin is the option to avoid.

I'm not asking you to hide the limitations — the silent-failure behavior in particular is worth stating, and the logger-name tip is genuinely useful. But it should read as guidance: lead with what the plugin does and how to configure it, then a short, factual limitations list. Right now the caveats outweigh the instructions.

joefernandez (approved):
Approved with comments: please address.

happyhuman:
Thanks for the review — all addressed in 8ce83b3.

- **Contradicting the tables.** Dropped the "table below" reference. Each caveat now sits with what it qualifies: the captured-events table (two event types, no views), the schema reference (which columns are populated), and the lifecycle payload table (`{"message": ...}` rather than `{}`).
- **Configuration options.** Added a Kotlin tab documenting the six `BigQueryLoggerConfig` properties, and removed the "Python and Java only" claim from the Quickstart tab — the scoping now lives in the section it scopes.
- **Pricing.** Removed the 1 KB minimum / free tier / cost-scaling sentence. Only the mechanism remains (`tabledata.insertAll`, one row at a time, synchronous).
- **Tone.** The Quickstart tab leads with usage and configuration; limitations are one sentence in the collapsed note. "A misconfigured agent looks healthy while writing nothing" is now "logging never fails the turn". Kept the logger-name tip.
- **Opener.** Leads with what the plugin logs; "supports a subset" preamble is gone.

Your two commits covered the `title` attribute and `??? note`.

Verification: L0 symbols, L1 compile, L2 lint, L3 transclusions, L5 registration and L6 badge all pass (L4 skipped as before — no `runSnippets` task). Re-grounded against the `v0.8.0` tag. The new config sample is inline rather than transcluded, matching the sibling Python and Java tabs, so it is symbol-checked but not compiled.

docs/integrations/bigquery-agent-analytics.md
```diff
@@ -716,6 +800,11 @@ account) under which the agent is running needs these Google Cloud roles:
 The events table (`agent_events`) uses a flexible schema. The following table
 provides a comprehensive reference with example values.
 
+In **Kotlin**, the plugin creates the table with these same columns but
```
joefernandez: This placement interrupts the introduction of the table; moved this detail to after the table.

joefernandez (approved):
Approved with a couple edits

PR #2149, opened 2026-08-19
Add the Kotlin tab for answering a long-running tool call

happyhuman:
## Summary

The **Intermediate / final result updates** section of `tools-custom/function-tools`
had tabs for Python, TypeScript, Go and Java, but none for Kotlin. This adds one.

Long-running tools are documented in two halves on this page: defining the tool,
which Kotlin already covered, and answering it from the client, which it did not.
Nothing anywhere in the docs showed a Kotlin reader how a deferred result gets
back to the model — the only Kotlin mention of `longRunningToolIds` in the repo
is a commented-out field listing in `events/index.md`.

## What the snippet shows

It continues the reimbursement scenario the Kotlin tab further up the page
already sets up, rather than importing the nav-agent scenario the SDK's own demos
use, so the two Kotlin tabs read as one story. Three things it makes explicit,
each of which is easy to get wrong:

- **Finding the pending call.** It is the one whose id the event also lists in
  `longRunningToolIds`.
- **The id is load-bearing.** The framework matches the response to the original
  call by id, and an unknown or missing one **throws** —
  `HistoryRewriterProcessor.findMatchingFunctionCallEvent`, and for a null id too,
  since the id set is built with `mapNotNull` and an empty set matches no event.
  Not a soft mismatch the model shrugs off.
- **Turn 1 differs by app.** Because this tool returns a value rather than `Unit`,
  a non-resumable app emits the placeholder as a function response and calls the
  model a *second* time, so the user sees an interim reply before the decision
  exists. A resumable app pauses on the function call with no second model call.
  Anyone building a HITL flow needs to know this before they budget model calls.

## Two things the backlog row got wrong

The KT-21 row proposed a new file and told the author to subclass `BaseTool`.
Both would have hurt:

- **No new file.** This page already owns a registered
  `snippets/tools/function-tools/LongRunningTool.kt`, transcluded 90 lines
  earlier. A second file in another directory splits one page's Kotlin in two, so
  this appends a region instead.
- **`@Tool(isLongRunning = true)`, not a `BaseTool` subclass.** The page's
  existing Kotlin tab uses the annotation. Following the row would have left two
  Kotlin tabs on one page prescribing different APIs.

The row also missed that `Runner.runAsync` takes an `invocationId` at all — see
below, where it turns out not to be needed.

## Also fixed on the page

"Key aspects of this example" explains the group purely through
`LongRunningFunctionTool`, a class Kotlin does not have (`git grep` at `v0.8.0`
finds it only in a test comment). Two bullets now give the Kotlin form and the
turn-count behaviour.

## Review notes

A sceptical pass over the first commit caught three claims that were wrong, all
corrected in `1c48e14f`:

- The snippet passed `invocationId` on the resume call behind an `appIsResumable`
  flag, claiming a resumable app must. It must not: `resolveInvocationId`
  (`AbstractRunner.kt:468-483`) derives the invocation from the response's own id
  and **discards the caller's value**. The flag was inert. Both removed.
- "Returns a placeholder and the turn ends" was wrong for the snippet's own
  default — see the turn-count point above. Corrected from
  `LongRunningToolIntegrationTest`'s scenario table (`off | {status: pending} |
  2 calls | [FC, FR, text]`).
- A wrong id was described as the model failing to match; it throws.

## Verification

- `./tools/kotlin-snippets/runner.sh build …/LongRunningTool.kt` → **PASS**
  (JDK 17; the same command CI runs). `lint` → **PASS**.
- `verify_snippets.py` → **L0 symbols, L1 compile, L2 ktlint, L5 registration,
  L6 badge all PASS.** L4 skips — this project registers no `runSnippets` task.
- Rendered the page with the extension list from `mkdocs.yml`: the target group
  comes back as `['Python', 'TypeScript', 'Go', 'Java', 'Kotlin']`.

Every API claim is grounded in the `v0.8.0` tag of `google/adk-kotlin`, the
version `examples/kotlin/build.gradle.kts` pins — never the working tree.

## One pre-existing checker complaint, not from this PR

`check_transclusions.py` flags orphaned Kotlin tabs at lines 123 and 227. Both
reproduce on `main` with this branch stashed, and both are false positives: the
checker mis-flags a Kotlin tab whose preceding sibling tab ends in prose rather
than a code fence, and rendering shows those groups intact as four-tab sets. The
ladder's L3 gate is file-level, so while this page had uncommitted edits it
attributed them to this change; against the committed branch L3 passes and
reports them as pre-existing elsewhere. The checker heuristic wants fixing, in
its own change.


wikaaaaa (approved):

joefernandez (approved):
Approved with minor updates.

happyhuman:
Thanks for the review. There were no change requests here, so 3857932 is small:

- **Grammar** in the turn-count bullet — "non-resumable apps send that placeholder ... and **calls** it a second time" → "and call it", and dropped the doubled "so".
- **Qualified the Resume note.** Adding a Kotlin tab to this section quietly extended that note to Kotlin, where it does not hold. `resolveInvocationId` (`AbstractRunner.kt:467`) matches the function response's own call ID against the session and ignores the `invocationId` the caller passes, so a Kotlin reader would go looking for a parameter that changes nothing. A response whose ID matches no function call throws rather than starting a new invocation. One sentence, in the note itself — same principle as the scoping change on #2147.

Happy to drop the second one if you would rather keep this PR to the tab it was opened for.

Verification: L0, L1 compile, L2 lint, L3, L5 and L6 all pass; L4 skipped as before. The orphaned-tab false positive noted in the description does not reproduce against the committed branch — L3 is clean.

joefernandez (approved):
Thanks for the update!

PR #2151, opened 2026-08-19
Add the Kotlin tab for structured input and output schemas

happyhuman:
## Summary

The structured-data section of `agents/llm-agents` had Python, TypeScript, Go and
Java tabs but no Kotlin. This adds one, and adk-kotlin **0.8.0** is what makes it
worth writing: `com.google.adk.kt.types.Schema` gained the twelve JSON Schema
constraint fields — `pattern`, `minLength`, `maxLength`, `minimum`, `maximum`,
`minItems`, `maxItems`, `format`, `nullable`, `default`, `anyOf` and `title`. At
0.7.0 the class had six properties and none of these.

The snippet declares an output schema with `minLength`/`maxLength` on the capital
name and `^[A-Z]{2}$` on an ISO 3166-1 country code, then wires it into
`LlmAgent(outputSchema = …, outputKey = …)`.

## Four corrections to what this page implies for Kotlin

Most of the review effort went here rather than into the snippet. The page's
surrounding bullets and warning describe Python behaviour, and a bare Kotlin tab
would have inherited all of it. Each of these is checked against the pinned
sources, not inferred:

- **The constraints are the model's to honour, not ADK's.** `SchemaUtils` reads
  none of the twelve fields; its validation covers `type`, `required`,
  `nullable`, `anyOf` and `items` only. They are forwarded to the API. A reader
  would otherwise assume `minLength` guarantees a non-empty value in state.
- **`outputKey` receives a `Map`, not text.** With `outputSchema` set,
  `LlmAgent.kt:360-374` stores the parsed object; on a validation failure it
  logs and stores the *raw string* under the same key. So
  `state["found_capital"] as String` throws on the happy path, and the value's
  runtime type is the only signal of which outcome occurred — while the page
  bullet directly above the tab group says the text content is saved.
- **Only a top-level object schema is accepted**, as in the Java ADK. Listing
  `minItems`/`maxItems` without this would point readers at a top-level array
  that fails validation and, per the previous point, fails *silently*.
- **Tools are not excluded.** The warning above the group says `output_schema`
  with tools is only supported by specific models. Kotlin applies the schema
  directly on models that support both and falls back to a `set_model_response`
  tool on those that don't (`LlmAgent.kt:100-107`) — so a Kotlin reader should
  not restructure into sub-agents to dodge a limitation they don't have.

Two of the new fields carry conditions of their own, both from the upstream
KDoc: Gemini rejects a `format` other than `int32`/`int64` on `Type.INTEGER` or
`Type.NUMBER`, or `enum`/`date-time` on `Type.STRING`; and `default` must hold a
JSON-native value, which ADK's own `Json` serializes but a hand-rolled one
without a contextual `Any` serializer does not.

## Review notes

A sceptical pass over the first commit caught a bad example, fixed in `6c0242b4`:
the original demonstrated `pattern` with `^[A-Z][A-Za-z .'-]*` on a capital-city
name. Under full-match semantics that rejects Bogotá, Brasília, Reykjavík, San
José and "Washington, D.C." — correct answers the model would be penalised for.
Under JSON Schema's partial-match default it rejects nothing at all, since the
tail can match zero characters. Either way it was the wrong constraint to teach
for that data, so `pattern` moved to the country code, where it is exactly right,
and the instruction was updated to match the schema it is paired with.

The same pass removed a `// Cannot use tools effectively here.` comment copied
from the Python and Java tabs, which is false for adk-kotlin.

## Not inline, unlike the backlog row

KT-22 specifies inline Kotlin. Every other Kotlin tab on this page transcludes a
region from `CapitalAgent.kt`, even though its Python and Java siblings are
inline, so inline here would break the page's convention and give up CI compile
coverage. It is a `schema_example` region instead.

## Verification

- `./tools/kotlin-snippets/runner.sh build …/CapitalAgent.kt` → **PASS** (JDK 17;
  the command CI runs). `lint` → **PASS**.
- `verify_snippets.py` → **L0 symbols, L1 compile, L2 ktlint, L3 transclusions,
  L5 registration, L6 badge all PASS.** L4 skips — no `runSnippets` task here.
- Rendered the page with the extension list from `mkdocs.yml`: the target group
  comes back as `['Python', 'TypeScript', 'Go', 'Java', 'Kotlin']` and the new
  bullets render inside the Kotlin tab.

Every claim is grounded in the `v0.8.0` tag of `google/adk-kotlin` — the version
`examples/kotlin/build.gradle.kts` pins — never the working tree, which is
currently 13 commits ahead of it.


wikaaaaa (approved):

docs/agents/llm-agents.md
```diff
@@ -574,6 +574,39 @@ schema definitions.
             .build();
     ```
 
+=== "Kotlin"
```
joefernandez, line 598: **Structural: this tab is shaped differently from every other tab on the page.**

The Python, Java and TypeScript tabs above are a short lead-in plus a snippet. This one is a lead-in, a snippet, a four-item caveat list, and a two-condition footnote — roughly 25 lines of prose against their two or three.

Because tabs are read comparatively, that asymmetry carries a message the content doesn't intend: a reader flipping between tabs sees Kotlin as the one that needs warnings and concludes it's the flaky option. Most of what's here isn't Kotlin-specific at all — "constraints are the model's to honour", "only a top-level object schema is accepted", tool fallback behavior — that's ADK-wide behavior that happens to be getting written down for the first time in the Kotlin tab.

Suggestion: promote the language-neutral parts into the section's shared prose above the tabs, where they benefit every reader and don't skew the comparison. Keep in the Kotlin tab only what is genuinely Kotlin-specific — realistically just the `com.google.adk.kt.types.Schema` vs GenAI SDK disambiguation at line 579.

docs/agents/llm-agents.md
```diff
@@ -574,6 +574,39 @@ schema definitions.
...
+
+    The input and output schema is ADK's own `com.google.adk.kt.types.Schema`,
+    not the same-named type in the GenAI SDK. Since adk-kotlin 0.8.0 it carries
```
joefernandez: version language support is unclear, here, change to:

Since adk-kotlin 0.8.0 it carries --> Starting with ADK Kotlin v0.8.0, the JSON schema includes constraints for the following fields:

docs/agents/llm-agents.md
```diff
@@ -574,6 +574,39 @@ schema definitions.
...
+    the JSON Schema constraint fields: `pattern`, `minLength`, `maxLength`,
+    `minimum`, `maximum`, `minItems`, `maxItems`, `format`, `nullable`,
+    `default`, `anyOf` and `title`.
+
+    ```kotlin
+    --8<-- "examples/kotlin/snippets/agents/llm-agent/CapitalAgent.kt:schema_example"
+    ```
+
+    Four things are worth knowing before relying on this:
```
joefernandez: "Four things are worth knowing before relying on this" sets a hedging frame — it tells the reader the feature is something to be wary of before they've seen a reason to be. 

At the minimum, remove or restate this over-the-top language:

- "ADK forwards them to the API but never checks them"

If these four points are correct they're just how the feature works, and should read that way: "How schema validation behaves" or similar, stated as documentation rather than as warnings. The counted-list framing ("four things") also becomes a maintenance trap the moment someone adds a fifth.

docs/agents/llm-agents.md
```diff
@@ -574,6 +574,39 @@ schema definitions.
...
+
+    - **The constraints are the model's to honour.** ADK forwards them to the
+      API but never checks them; its own validation covers structure only —
+      `type`, `required`, `nullable`, `anyOf` and `items`.
+    - **`outputKey` receives a `Map`, not text.** With `outputSchema` set, the
+      parsed object is what lands in state. If the response fails validation,
+      ADK logs the error and stores the raw string under the same key, so the
+      value's type is the only signal of which happened.
```
joefernandez: **Two substantive behavioral claims that need sourcing, and one that describes a rough edge as if it were an API.**

"ADK forwards them to the API but never checks them" and the `outputKey` fallback are real, useful things to document — but they're stated flatly with nothing to verify against. Can you link the validation code, the way the compaction page links `LlmEventSummarizer`? Otherwise a reader has no way to confirm it and no way to notice when it changes.

The second bullet is the one I'd push back on hardest: "If the response fails validation, ADK logs the error and stores the raw string under the same key, so the value's type is the only signal of which happened." That's telling readers to type-sniff a state value to find out whether an error occurred. If that really is the only signal, it's a bug worth filing against adk-kotlin rather than a technique worth documenting, writing it down here makes it more likely someone will build on it.

Is this behavior Kotlin-only, or does Python/Java do the same? If it's shared, it belongs above the tabs (see my first comment).

docs/agents/llm-agents.md
```diff
@@ -574,6 +574,39 @@ schema definitions.
...
+      top-level array or primitive fails validation.
+    - **Tools are not excluded.** On models that support a response schema
+      alongside tools the schema is applied directly; on models that do not,
+      ADK falls back to a `set_model_response` tool.
+
+    Two of the new fields carry conditions: Gemini rejects a `format` other than
+    `int32`/`int64` on `Type.INTEGER` or `Type.NUMBER`, or `enum`/`date-time` on
+    `Type.STRING`; and `default` must hold a JSON-native value, which ADK's own
+    `Json` serializes but a hand-rolled one without a contextual `Any`
+    serializer does not.
```
joefernandez: **Third-party API behavior that will drift out from under this page.**

"Gemini rejects a `format` other than `int32`/`int64` on `Type.INTEGER` or `Type.NUMBER`, or `enum`/`date-time` on `Type.STRING`" is a claim about the Gemini API's current accepted values, not about ADK. That list changes on Gemini's release schedule, with nothing tying it to this repo — no test covers it, no CI catches it, and nobody updates adk-docs when Gemini adds a format. It'll be quietly wrong at some point and no one will know.

Prefer linking the Gemini structured-output / `Schema` reference and letting it be authoritative, rather than restating a value list here.

The `default` half is different — the contextual `Any` serializer requirement is genuinely ADK-Kotlin-specific and worth keeping. Might be clearer split into its own sentence rather than sharing one with the Gemini constraint.

joefernandez (commented):
Needs some work. See comments.

happyhuman:
Thanks — this was the right call, and checking the other SDKs made it clearer than I expected. Addressed in 2114b55.

**You were right that most of it wasn't Kotlin-specific.** I checked all four SDKs; only one of the four bullets survived as Kotlin's:

| Claim | Python | Java | Kotlin | Go |
|---|---|---|---|---|
| `set_model_response` fallback when tools + schema | yes | yes | yes | yes |
| `output_key` gets the parsed object, not text | `dict` | `Map` | `Map` | — |
| Constraint fields not re-checked by ADK | no (Pydantic enforces) | yes | yes | — |
| Top-level object schema only | no (also list/primitive) | yes | yes | — |
| Raw string stored on validation failure | no (raises) | yes | yes | — |

So:

- **Tools are not excluded** is ADK-wide. Python has `_OutputSchemaRequestProcessor`, Go has `outputschema_processor.go`, Java has `OutputSchema.java`. The warning's "consider using sub-agents" was pointing every reader around a limitation none of the SDKs has, so I named the automatic fallback in the warning and kept your reliability caveat.
- **`output_key` receives the parsed object** is also ADK-wide, which means the existing bullet ("the text content ... will be automatically saved") was already wrong for anyone setting `output_schema`. Corrected in the bullet itself.
- The remaining three are **Java + Kotlin**, not Kotlin — they're now one shared note titled "Schema validation in Java and Kotlin", which says what Python does differently instead of leaving Kotlin looking like the outlier.

**On the type-sniffing bullet** — you pushed hardest here and you were right to. But it isn't a Kotlin bug: `LlmAgent.java:628-642` does exactly the same catch-log-store-raw. I've dropped "the value's type is the only signal of which happened" so nothing suggests sniffing the type, and just stated that the error is logged and the raw string stored. If that behavior should change, it's one issue against both JVM SDKs rather than adk-kotlin alone — happy to file it, but not from this PR.

**Also done:**
- Version wording changed to yours verbatim.
- Dropped "Four things are worth knowing before relying on this" and "ADK forwards them to the API but never checks them".
- Claims now link source, pinned at `v0.8.0` so the line references don't rot: [`SchemaUtils`](https://github.com/google/adk-kotlin/blob/v0.8.0/core/src/commonMain/kotlin/com/google/adk/kt/SchemaUtils.kt), [`LlmAgent`](https://github.com/google/adk-kotlin/blob/v0.8.0/core/src/commonMain/kotlin/com/google/adk/kt/agents/LlmAgent.kt). (The compaction page turned out not to link source — the precedent I followed is the adk-python link already in this section's warning.)
- Gemini `format` value list replaced with a link to the [`Schema` reference](https://ai.google.dev/api/caching#Schema). Kept the `default` contextual-serializer note, split into its own sentence as you suggested.

The Kotlin tab is now a lead-in, the snippet, and two short sentences — same shape as its siblings.

Verification: L0, L1 compile, L2 lint, L3, L5, L6 pass; L4 skipped as before. Tab group still renders `['Python', 'TypeScript', 'Go', 'Java', 'Kotlin']`.

docs/agents/llm-agents.md
```diff
@@ -574,6 +595,25 @@ schema definitions.
...
+    not the same-named type in the GenAI SDK. Starting with ADK Kotlin v0.8.0,
+    the JSON schema includes constraints for the following fields: `pattern`,
+    `minLength`, `maxLength`, `minimum`, `maximum`, `minItems`, `maxItems`,
+    `format`, `nullable`, `default`, `anyOf` and `title`.
+
+    ```kotlin
+    --8<-- "examples/kotlin/snippets/agents/llm-agent/CapitalAgent.kt:schema_example"
+    ```
+
+    `format` accepts only the values the model allows for the field's type. For
```
joefernandez: Don't start sentences with code syntax: 
```suggestion
    The `format` field accepts only the values the model allows for the field's type. For
```

docs/agents/llm-agents.md
```diff
@@ -574,6 +595,25 @@ schema definitions.
...
+    the accepted values, see the Gemini [`Schema`
+    reference](https://ai.google.dev/api/caching#Schema).
+
+    `default` must hold a JSON-native value. ADK's own `Json` serializes one,
```
joefernandez:
```suggestion
    The `default` field must contain a JSON-native value. ADK's own `Json` serializes one,
```

joefernandez (approved):
Thanks for updates! Approved.

PR #1975, opened 2026-07-18
docs(bqaa): document ADK Java preview-readiness fixes (pause/resume tracing, drop stats, redaction)

caohy1988:
## Summary

Documents the ADK Java **BQAA "preview-readiness fixes"** — shipped in **adk-java `v1.7.0`** (released 2026-07-20) — in the BigQuery Agent Analytics page:

- google/adk-java@c685ece46b (July 9) — redaction, table bootstrap, drop stats
- google/adk-java@2027a4b53d (July 16) — tracing, lifecycle, redaction, HITL/`TOOL_PAUSED` pairing

All behavioral claims were verified against the plugin source (`BigQueryAgentAnalyticsPlugin`, `BigQueryLoggerConfig`, `BigQueryUtils`, `JsonFormatter`, `PluginState`, `BatchProcessor`, `TraceManager`); the reviewed commit `ba23601c` has no `agentanalytics` diffs against the `v1.7.0` tag.

## What changed

- **ADK 2.0 section**: `Java v1.7.0` language-support chip added (per the #1942 convention). Java emits `TOOL_PAUSED` and the pause/resume pairing keys, but stores `pause_kind` / `function_call_id` at the **top level of `attributes`** (no `attributes.adk` envelope, which is Python-only), and does **not** emit `AGENT_TRANSFER` / `AGENT_STATE_CHECKPOINT` / `EVENT_COMPACTION`. Includes the Java variant of the base-table pairing query, and notes that `HITL_*_REQUEST_COMPLETED` rows carry the same top-level pair keys (no dedicated HITL-completion view).
- **Views footnote**: Java (v1.7.0+) creates `v_tool_paused` and the pause columns on `v_tool_completed`; the three workflow views are Python-only.
- **Reliability fixes intro + Dropped-event observability**: Java `getDropStats()` with all **eight** drop reasons (`queue_full`, `append_error`, `serialization_error`, `after_close`, `shutdown_timeout`, `writer_permit_exhausted`, `writer_create_error`, `late_after_finalize`); `append_error` defined per the `BatchProcessor` contract (any batch-prep/append failure other than `AppendSerializationError`); tabbed code samples; closing guidance treats any non-zero counter as actual loss.
- **Built-in redaction**: Java (v1.7.0+) redacts the six sensitive keys plus `temp:`-prefixed keys — but **not** `secret:` (Python-only). Java `contentFormatter` contract documented (thread-safe, fast/non-blocking, return a new object) plus fail-closed behavior on exceptions.
[...]

PR #2154, opened 2026-08-20
docs(bqaa): align plugin docs with Python 2.9 and Java 1.9

caohy1988:
## Summary

Aligns the BigQuery Agent Analytics integration page with the released ADK Python 2.9.0 and Java 1.9.0 implementations, while preserving the Kotlin integration material now on docs `main`.

The feature availability gates remain accurate: the cumulative update carries forward Python 2.7 delivery, schema, error, and workflow-node behavior and Java 1.7 reliability, redaction, tracing, and lifecycle behavior. It now also covers the Java configuration change released in 1.8 and still present in 1.9.

## Key decisions

- Java v1.8.0 and later require `datasetId` and default `tableName` to `agent_events`; v1.7.0 and earlier used `agent_analytics` and `events`.
- The shared page explicitly distinguishes language-specific contracts: Python-only error/workflow events and `event_id`, Java view-column differences, Java opt-in view creation, and Python-only fork handling.
- Default Python delivery remains best-effort. `event_id` is the consumer deduplication key, while `exactly_once_delivery` prevents ambiguous-retry duplicates within a live processor rather than guaranteeing lossless delivery.
- Drop counters distinguish lost rows from formatter/parser incidents that still write sentinel content.

## Validation

- `uv run --isolated --with-requirements requirements.txt mkdocs build --strict`
- `uvx --from codespell==2.4.1 codespell docs/integrations/bigquery-agent-analytics.md`
- `git diff --check origin/main`
- Verified all 22 external links on the page return HTTP 200.
- Mechanically compared documented config fields, schema columns, event types, view definitions, drop reasons, and lifecycle methods with `google/adk-python@v2.9.0` (`16793842ccb8206824a80a206c2ab42d3064d9f4`) and `google/adk-java@v1.9.0` (`5696f43ebe6287e431e7a34421fddded60574381`).
- Merged docs `main` at `3b350ea25b155503b722deb2d1bf37c27f29f6ed`; the cumulative PR diff remains limited to `docs/integrations/bigquery-agent-analytics.md`.

## Related

Related: #1975


caohy1988:
## Full review — verified against `google/adk-python@4599a526` and the `v2.7.0` / `v2.7.1` tags

**Verdict:** The PR is in very good shape — accurate, and materially better than what it replaces. I verified the doc's factual claims line-by-line against the actual plugin source at the pinned commit and the release tags. One likely-broken SQL snippet and one broken sentence; both easy fixes. Everything else checked matched the source exactly.

### Issues to fix

**1. The `event_id` dedup query will likely fail in BigQuery — `QUALIFY` requires a `WHERE`, `GROUP BY`, or `HAVING` clause** (*Delivery and deduplication* section)

```sql
SELECT *
FROM `your-gcp-project-id.adk_agent_logs.agent_events`
QUALIFY
  event_id IS NULL
  OR ROW_NUMBER() OVER (PARTITION BY event_id ORDER BY timestamp) = 1;
```

BigQuery rejects a bare `SELECT * FROM t QUALIFY …` with "The QUALIFY clause requires at least one of these clauses: WHERE, GROUP BY, HAVING" — a long-standing, documented restriction. The conventional fix is to add `WHERE TRUE` before `QUALIFY`. Worth running the snippet once against a real table to confirm.

**2. Broken sentence in the "Java attribute location" admonition.** The em-dash cleanup rewrote the first half of the sentence but left the old continuation line behind:

> The Java plugin writes the pair keys at the top level of `attributes`:
> `"attributes": {"pause_kind": "tool", "function_call_id": "call-789"}`.
> with no `adk` wrapper. In the base-table query below, …

"with no `adk` wrapper." is now a stranded fragment after a full stop. Suggested rewrite: *"The Java plugin writes the pair keys at the top level of `attributes`, with no `adk` wrapper: `"attributes": {…}`. In the base-table query below, replace …"*

### Verified correct (against adk-python source)

- **Packaging**: the `bigquery-analytics` extra exists with exactly the four documented deps (`google-cloud-bigquery`, `-bigquery-storage`, `-storage`, `pyarrow`); `pyarrow` is absent from the `gcp` extra; the import error names the `bigquery-analytics` extra.
- **Version gating**: `event_id`, `exactly_once_delivery`, and `NODE_OUTPUT`/`NODE_ERROR` are all inside the `v2.7.0` tag, so "Python v2.7.0 and later" is right. (The mainline `event_id` commit `4aa1a811` isn't an ancestor of the tag, but the release branch carries an equivalent cherry-pick — checked the tag's file content directly.)
- **Config table**: `final_response_tool_names: frozenset[str] = frozenset()`, `flush_on_run_end: bool = True`, `exactly_once_delivery: bool = False` all match, including the `source_tool` attribute on tool-sourced `AGENT_RESPONSE`.
- **Drop reasons**: all 13 Python reason strings exist verbatim in source, and the split between lost-row reasons and the two written-with-sentinel incident counters (`formatter_failed` → `[FORMATTER_FAILED]`, `content_parse_failed` → `[CONTENT_PARSE_FAILED]`) matches the implementation. The 30-second rotation backoff is real (`self._rotation_retry_at = now + 30.0`).
- **Redaction**: the 20-key list matches `_SENSITIVE_KEYS` exactly, including hyphen/underscore normalization; only `temp:` is a redacted prefix (`secret:` genuinely has no special handling anymore); the text/URI sanitization claims (auth headers, bearer/basic, signed-URL params) match the regex battery in the source.
- **Schema & views**: `event_id` is the second schema field (matching the DDL placement), `_SCHEMA_VERSION = "2"`, `event_id` is in the views' common columns; `v_llm_response` columns match in content **and order** (through `cache_metadata`, `cache_type`, `finish_reason`); `v_agent_error` / `v_invocation_error` / `v_node_output` / `v_node_error` columns match; A2A's `content` = `a2a:response` payload with namespaced keys under `attributes.a2a_metadata` matches.
- **Semantics**: `NODE_OUTPUT` gating (`event.output` present and `message_as_output is not True`), `NODE_ERROR` gating (`error_code` not in the LLM finish/block set), envelope fields `route` / `render_ui_widgets` / `rewind_before_invocation_id`, `flush()` being current-loop-scoped, and `close()` delegating to `shutdown()` — all confirmed.

### Caveat

The Java-side claims (v1.7.0 gating for `getDropStats` / shutdown hook / redaction, the six-key Java redaction list, pre-1.7 plugin-owned OTel spans, top-level pair keys) weren't verifiable in this pass — no adk-java checkout at hand. They read consistent with the prior doc text, but if any weren't already reviewed in the closed Java-doc PR, a spot check against adk-java would close the loop.

One nice touch worth keeping: replacing the issue-#3845 credential warning with the honest "redaction is not DLP" framing is more accurate now that the sanitizer battery exists, without overpromising.

---
🤖 Review performed with [Claude Code](https://claude.com/claude-code)

caohy1988:
> The `event_id` dedup query will likely fail in BigQuery — `QUALIFY` requires a `WHERE`, `GROUP BY`, or `HAVING` clause.

Not changing this snippet. GoogleSQL's [`SELECT` grammar](https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#select_list) makes `WHERE`, `GROUP BY`, and `HAVING` optional before `QUALIFY`; the `QUALIFY` requirement is that a window function appear in `QUALIFY` or the select list. I also validated the same query shape with `bq query --use_legacy_sql=false --dry_run`; it succeeded and reported 0 bytes processed.

> Broken sentence in the "Java attribute location" admonition.

Addressed in `5f48f804`: "with no `adk` wrapper" is now part of the opening sentence, and the JSON example no longer leaves a stranded fragment.

The Java-side behavioral claims were carried forward from the earlier Java-focused review in #1975; this follow-up does not expand them.

PR #2156, opened 2026-08-20
Add Kotlin tabs to the Skills page

happyhuman:
## What

Adds Kotlin to all three tab groups on `docs/skills/index.md` (get started, skills defined in code, skills read from the filesystem), plus the `Kotlin v0.1.0` language badge and a Kotlin entry under **Next steps**.

New snippet file: `examples/kotlin/snippets/skills/SkillsExample.kt`, registered in `files_to_test.txt`.

## Why

`SkillToolset` has existed since adk-kotlin v0.1.0, but the page has never shown Kotlin. Earlier coverage audits diffed one release tag against the next, so anything that already existed at v0.1.0 was invisible to them.

## Kotlin differs from the sibling tabs in two ways

Both are deliberate, not translation shortcuts:

- **`SkillToolset` takes a single `SkillSource`**, not a list of pre-loaded skills plus `additional_tools`. `NewFileSystemSource` discovers every skill directory under a base directory, so there is no per-skill load call. Extra tools go on the agent's `tools`, not on the toolset.
- **There is no built-in source for skills defined in code.** Only `NewFileSystemSource` (JVM/Android) and `AssetSkillSource` (Android-only) implement `SkillSource` at v0.8.0. The inline-skills tab therefore implements `SkillSource` directly, mirroring how the ADK Go tab handles the same gap, rather than inventing a Python-style model class.

`AssetSkillSource` is `androidMain` and is intentionally left out of scope.

## Verification

Grounded against the `v0.8.0` git tag (not the adk-kotlin working tree). Full ladder green:

| Level | Check | Result |
| --- | --- | --- |
| L0 | symbols/named args exist at 0.8.0 | PASS |
| L1 | `gradlew build` (JDK 17, KSP) | PASS |
| L2 | ktlint (report-parsed) | PASS |
| L3 | transclusions + tab structure | PASS |
| L5 | `files_to_test.txt` registration | PASS |
| L6 | language badge | PASS |

L4 (`runSnippets`) reports SKIP — no such Gradle task exists in this repo, so it is skipped for every snippet, not just these.

The page is badged Experimental for the other languages; I checked and `SkillToolset` carries no opt-in annotation at v0.8.0, so the Kotlin badge is added plainly.

Tracked as KT-23.

---

### DIFF
--- a/docs/skills/index.md
+++ b/docs/skills/index.md
@@ -17,7 +17,8 @@ impact on the operating context window of the agent.
     respective ADK GitHub repositories:
     [ADK Python](https://github.com/google/adk-python/issues/new?template=feature_request.md&labels=skills),
     [ADK TypeScript](https://github.com/google/adk-js/issues/new?template=feature_request.md&labels=skills),
-    [ADK Go](https://github.com/google/adk-go/issues/new?template=feature_request.md&labels=skills).
+    [ADK Go](https://github.com/google/adk-go/issues/new?template=feature_request.md&labels=skills),
+    [ADK Kotlin](https://github.com/google/adk-kotlin/issues/new).
 
 ## Get started
 
--- a/docs/skills/index.md
+++ b/docs/skills/index.md
@@ -1,7 +1,7 @@
 # Skills for ADK agents
 
 <div class="language-support-tag">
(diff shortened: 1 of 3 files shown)


wikaaaaa (approved):

joefernandez (approved):
LGTM. Approved.

PR #2157, opened 2026-08-20
Add Kotlin tabs to the tool confirmation page

happyhuman:
## What

Adds Kotlin to all three tab groups on `docs/tools-custom/confirmation.md` (boolean, dynamic/threshold, advanced), plus the `Kotlin v0.1.0` badge and two prose updates that previously listed every language except Kotlin.

New snippets, both registered in `files_to_test.txt`:

- `examples/kotlin/snippets/tools/confirmation/ToolConfirmationExample.kt` — boolean + advanced
- `examples/kotlin/snippets/tools/confirmation/dynamic/ReimbursementTools.kt` — threshold

## One correction to the tracking notes

The backlog entry for this row said there is no confirmation support to mirror and warned against inventing one. The first half turned out to be wrong: `@Tool(requireConfirmation = true)` exists at v0.8.0 and the KSP processor forwards it to `FunctionTool(requiresConfirmation = ...)`. So the boolean tab is a direct equivalent of Python's `FunctionTool(require_confirmation=True)`, not a hand-rolled workaround. The warning itself still held — there is no separate confirmation *tool type*, and none was invented.

## Deliberate divergences from the sibling tabs

- **Dynamic gating happens inside the tool.** The annotation flag is a compile-time constant, and `@Tool` has no predicate form, so the threshold is evaluated with `ToolContext` exactly as the Java tab does. Called out in a note on the tab and in the section prose.
- **The advanced example reads the payload through `Number`.** `ToolConfirmation.payload` is `Any?` and arrives decoded from JSON, so its numeric type is not guaranteed. Casting straight to `Int` would be a latent `ClassCastException`; this is the same hazard the Go tab flags for `float64`.

## Why two files

`@Tool` generates a wrapper class named after the function, in the function's package. The boolean and threshold examples are both naturally called `reimburse`, which would collide. Rather than rename one in a way that would read as meaningful, the threshold example lives in its own package — the `package` line sits outside the transcluded region, so both tabs show a plainly named `reimburse`.

## Verification

Tracked as KT-24.

---

### DIFF
--- a/docs/tools-custom/confirmation.md
+++ b/docs/tools-custom/confirmation.md
@@ -207,14 +207,14 @@ You can modify the behavior of the confirmation requirement by using a function 
 
 === "Kotlin"
 
+    ```kotlin
+    --8<-- "examples/kotlin/snippets/tools/confirmation/dynamic/ReimbursementTools.kt:dynamic_confirmation"
+    ```
+
     !!! note
         The `@Tool` annotation's `requireConfirmation` flag is a compile-time
         constant, so a threshold is evaluated inside the tool using the
         `ToolContext`, as in ADK Java.
-
-    ```kotlin
-    --8<-- "examples/kotlin/snippets/tools/confirmation/dynamic/ReimbursementTools.kt:dynamic_confirmation"
-    ```
 
 ## Advanced confirmation {#advanced-confirmation}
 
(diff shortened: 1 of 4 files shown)


wikaaaaa (approved):

joefernandez (approved):
Thanks for the update

PR #2158, opened 2026-08-20
Add Kotlin safety settings and includeContents to the agent docs

happyhuman:
## What

Three related Kotlin parity gaps, all resting on the same three types (`SafetySetting`, `HarmCategory`, `HarmBlockThreshold`), so they are fixed together:

| Page | Gap |
| --- | --- |
| `docs/safety/index.md` | No Kotlin tab in the built-in Gemini safety section, and no Kotlin in the badge div |
| `docs/agents/llm-agents.md` (`gen_config`) | Kotlin tab existed but set only `temperature` and `maxOutputTokens`; the Python sibling also sets safety settings |
| `docs/agents/llm-agents.md` (`include_contents`) | No Kotlin tab |

None of this is new API — safety settings landed in adk-kotlin 0.5.0, `includeContents` has been on `LlmAgent` since 0.1.0.

## The gap class worth noting

The `gen_config` case is a Kotlin tab that *exists* but under-shows the feature. A "missing Kotlin tab" scan cannot see it, and neither can a symbol diff, since every symbol the tab names is present. Only comparing a tab's contents against its siblings surfaces it.

## Review notes

- **The safety page tab is inline, not a transclusion.** Its siblings are inline, and like them the snippet elides the required `name` and `model` arguments, so there is nothing there that could compile standalone. It is covered by L0 (imports and named arguments checked against real v0.8.0 signatures) but not by the compiler — the same level of assurance the Python and Go tabs on that page get.
- **`gen_config` now shows more than the TypeScript and Java tabs do.** Those two also omit safety settings. This PR brings Kotlin to parity with Python, the semantic source for the group, and deliberately does not touch sibling-language tabs. The TS/Java gap is left as-is.
- **The safety page badge div carries no version numbers** for any language, so the Kotlin span follows suit and reads just `Kotlin`.

## :warning: Conflicts with #2151

Both branches modify `examples/kotlin/snippets/agents/llm-agent/CapitalAgent.kt`, and #2151 inserts its `schema_example` region at the same anchor this one inserts `include_contents` — immediately after `[end:gen_config]`. The import blocks also overlap. Both sides are pure additions, so the resolution is mechanical, but **please land #2151 first and let me rebase**, rather than merging this one on top blind.

## Verification

Tracked as KT-25, KT-32 and KT-36.

### DIFF
--- a/docs/agents/llm-agents.md
+++ b/docs/agents/llm-agents.md
@@ -663,6 +663,12 @@ Control whether the agent receives the prior conversation history.
             .build();
     ```
 
+=== "Kotlin"
+
+    ```kotlin
+    --8<-- "examples/kotlin/snippets/agents/llm-agent/CapitalAgent.kt:include_contents"
+    ```
+
 !!! note "Go v2.0.0: agent execution modes"
 
     ADK Go v2.0.0 introduces an explicit `Mode` field on `llmagent.Config` that
--- a/docs/safety/index.md
+++ b/docs/safety/index.md
@@ -1,7 +1,7 @@
 # Safety and Security for AI Agents
 
 <div class="language-support-tag">
-  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python</span><span class="lst-typescript">TypeScript</span><span class="lst-go">Go</span><span class="lst-java">Java</span>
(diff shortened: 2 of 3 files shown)


happyhuman:
**Update:** pushed a follow-up commit restoring the Go tab's indentation on `docs/safety/index.md`. Adding the Kotlin tab had accidentally reindented one line inside the Go sample (that file mixes tabs and spaces). The Go tab is a sibling and should not appear in this diff at all — the page's only non-additive change is now the badge line.

wikaaaaa (approved):

joefernandez (approved):
thanks!

PR #2159, opened 2026-08-20
Add the Kotlin tab for driving a long-running tool to completion

happyhuman:
## What

Adds the missing Kotlin tab to the `call_reimbursement_tool` group on `docs/tools-custom/function-tools.md`, via a new `call_reimbursement_tool` region in the `LongRunningTool.kt` file that already backs this page.

## Why

The page already has a Kotlin tab for *defining* a long-running tool (`@Tool(isLongRunning = true)`), but none for *resuming* one. That leaves the Kotlin reader exactly at the point where the invocation pauses, with no shown way forward — while Python, TypeScript, Go and Java all continue.

The Kotlin flow mirrors Python's: find the call whose id the event lists in `Event.longRunningToolIds`, hold on to the matching `FunctionResponse`, then send a copy back with the final status to resume.

## Scope: half this backlog row is deliberately not here

The row also named `RequestInputTool` and `GetUserChoiceTool`. I did not document them, because they have **no host page in any language**:

- Neither appears in any narrative `docs/**.md` page — only in generated API reference, plus incidental mentions (`docs/agents/config.md` name list, BigQuery analytics event names).
- `docs/graphs/human-input.md` is the graphs subsystem, which Kotlin does not have, so it is the wrong home — as the backlog itself noted.
- There is no general built-in-tools page to extend.

Adding a Kotlin-only section would invent page structure rather than close a Kotlin gap, so this is Kotlin being *level* with the other SDKs, not behind. Recorded for a product-docs request instead.

## Verification

Tracked as KT-26.


wikaaaaa (approved):

joefernandez (approved):
thanks!

PR #2160, opened 2026-08-20
Add Kotlin tabs for grounding with Agent Search

happyhuman:
## What

Adds Kotlin to both tab groups on `docs/grounding/grounding_with_search.md` — creating the grounded agent, and the optional citation display — plus the `Kotlin v0.1.0` badge.

## Why

Both groups showed Python and Java only. `VertexAiSearchTool` has been in adk-kotlin since v0.1.0; this was missed because earlier coverage audits diffed release tags against each other and never checked the v0.1.0 baseline.

## Translation notes

- **No builder.** The Java tab uses `VertexAiSearchTool.builder().dataStoreId(...).build()`. Kotlin has named parameters, so the tab uses the constructor directly. The builder does exist at v0.8.0, but it is marked `@AdkJavaInteropApi` and exists for Java callers.
- **`collect`, not a for-loop.** Events arrive as a `Flow<Event>` in Kotlin, and `isFinalResponse` is a property rather than a method.
- **`groundingChunks` is nullable**, so the count is guarded with `isNullOrEmpty()` rather than Java's `isPresent()`.

## Verification

Both tabs are **inline**, matching their siblings — the snippets elide surrounding setup and carry placeholder datastore ids, so there is nothing that could compile standalone. They are covered by L0, which validates every import and named argument against the real v0.8.0 signatures, and not by the compiler. That is the same level of assurance the Python and Java tabs on this page have.

Ladder: L0, L3, L5, L6 PASS; L1/L2 PASS across the unchanged snippet tree; L4 `runSnippets` SKIP (no such Gradle task in this repo).

Tracked as KT-27.

---


docs/grounding/grounding_with_search.md
wikaaaaa, line 15: Lets maybe add Kotlin guidance here as well? 

I think its the same as for Java.

It needs the `GOOGLE_APPLICATION_CREDENTIALS` and the environment variables that Python sets (GOOGLE_GENAI_USE_ENTERPRISE, GOOGLE_CLOUD_PROJECT, GOOGLE_CLOUD_LOCATION) just not in .env file but in the application environment (same as  `GOOGLE_APPLICATION_CREDENTIALS`)

wikaaaaa (approved):

happyhuman:
Added — Kotlin now shares the Java auth bullet. Also fixed the support tag: VertexAiSearchTool landed in v0.2.0.

docs/grounding/grounding_with_search.md
joefernandez: don't make this a note. just explain the situation and link for more details

```suggestion
Agent Search requires your ADK agent to be connected to a Google Cloud project authentication. You can not use a Gemini API Key from Google AI Studio when using this tool. For more information on connecting your ADK agent to Google Cloud projects, see the [Connect to Google Cloud](/get-started/google-cloud/) guide.
```

joefernandez (commented):
The two Kotlin tabs are clean additions. One flag on the prerequisites bullet, which changes existing Java guidance.

docs/grounding/grounding_with_search.md
joefernandez: looks like this is resolved for this page

REQUEST: Please add this Java / Kotlin tip to the Connect to Google Cloud page: https://adk.dev/get-started/google-cloud/

joefernandez (approved):
Approved. 

Please see open comment for a follow-up PR for https://adk.dev/get-started/google-cloud/

PR #2161, opened 2026-08-20
Extend the compaction summarizer section to Kotlin

happyhuman:
## What

- Adds the Kotlin tab to the **Define a Summarizer** group on `docs/context/compaction.md`.
- Fixes two lines of prose that understated Kotlin.

## Why

The group showed Python, Java and TypeScript only, and the prose around it repeated the omission:

| Line | Before | Issue |
| --- | --- | --- |
| ~199 | "`LlmEventSummarizer` (Python/Java) or `LlmSummarizer` (TypeScript)" | Kotlin also has `LlmEventSummarizer` |
| ~281 | "In Python and Java, customize the `prompt_template`" | Kotlin exposes `promptTemplate` as a constructor parameter |

Kotlin has had `LlmEventSummarizer` since v0.3.0. This complements the earlier work that covered only the token-threshold *config*.

## Inline, not a new .kt file

The backlog row suggested creating `CompactionSummarizer.kt`. I did not, deliberately:

- All four sibling tabs in this group are inline.
- The page's **existing** Kotlin tab (token-threshold config) is inline too.

A reader switching between the two Kotlin tabs on one page should not see one transcluded and one inline. Both snippets also reference an undefined `rootAgent`, exactly as their siblings do, so neither could compile standalone anyway.

The cost is that this tab gets L0 rather than the compiler. L0 does validate every import and named argument against real v0.8.0 signatures — which is the failure mode that matters here — but it cannot see type mismatches. `LlmEventSummarizer(model: Model, promptTemplate: String)` and `EventsCompactionConfig(compactionInterval, overlapSize, summarizer, ...)` were both read directly from the v0.8.0 tag.

## Verification

Ladder: L0, L3, L5, L6 PASS; L1/L2 PASS across the snippet tree; L4 `runSnippets` SKIP (no such Gradle task in this repo).

Tracked as KT-28.


docs/context/compaction.md
```diff
@@ -278,8 +278,37 @@ The following code example demonstrates how to define and configure a custom sum
...
+                    compactionInterval = 3,
+                    overlapSize = 1,
+                    summarizer = mySummarizer,
+                ),
+        )
+    ```
+
+You can further refine the compactor by modifying its summarizer. In Python, Java
+and Kotlin, customize the prompt template on `LlmEventSummarizer` — the property is
+`prompt_template` in Python and Java, and `promptTemplate` in Kotlin. In TypeScript,
```
wikaaaaa: Its actually promptTemplate in Java as well - https://github.com/google/adk-java/blob/c1bda9c14dd16a46849dbd6e74773e5595c050b8/core/src/main/java/com/google/adk/summarizer/LlmEventSummarizer.java#L53

wikaaaaa (approved):

happyhuman:
Good catch — fixed. Java uses `promptTemplate` too; only Python is snake_case.

joefernandez (approved):
Thanks for the updates!

PR #2163, opened 2026-08-20
Show maxLlmCalls in the RunConfig Kotlin tabs

happyhuman:
## What

- Adds `maxLlmCalls` to the two Kotlin regions on `docs/runtime/runconfig.md` (`basic_usage` → 200, `streaming_config` → 150), matching the values their siblings already use.
- Corrects one line of the runtime-limits prose that described the overflow error in Python-only terms.

## Why this was invisible

This is a parity gap, not a missing tab. Both Kotlin tabs **exist**, and every symbol in them is valid at v0.8.0 — so a symbol diff sees nothing and a "missing Kotlin tab" scan sees nothing. Only comparing a tab's *contents* against its siblings surfaces it. `RunConfig.maxLlmCalls` has been there since v0.5.0.

## Notes

- **The `streaming_config` region gets a comment.** Its siblings set `supportCfc` alongside `maxLlmCalls`; Kotlin's `RunConfig` has no equivalent, so the comment says so rather than leaving the reader to wonder what was dropped.
- **The audio/speech group is deliberately left without a Kotlin tab.** Kotlin's `RunConfig` carries only `streamingMode`, `maxLlmCalls` and `customMetadata` — there is no `speechConfig` or `responseModalities` to show, so a tab there would have to invent API.
- **One backlog detail was wrong.** The row said the `init` block "rejects `Int.MAX_VALUE` and any value <= 0". At v0.8.0 only `Int.MAX_VALUE` throws; a value <= 0 logs a warning and disables the limit. The existing page prose ("Set to 0 or negative for unlimited calls") is therefore correct for Kotlin as written, and I left it alone.
- **`LlmCallsLimitExceededException` is not documented here**, because the page does not describe what happens on limit exhaustion in *any* language. That is a docs-wide gap rather than a Kotlin one; adding a Kotlin-only mention would create a new asymmetry.

## Verification

Grounded against the `v0.8.0` git tag. Full ladder green (L0, L1 compile with KSP, L2 ktlint, L3, L5, L6). L4 `runSnippets` SKIP — no such Gradle task in this repo.

Tracked as KT-31.


wikaaaaa (approved):
"RunConfig.maxLlmCalls has been there since v0.1.0." in the comment description is not true. It was added in 0.5 - https://github.com/google/adk-kotlin/releases/tag/v0.5.0.

happyhuman:
Confirmed and corrected in the description — v0.5.0, not v0.1.0. Docs content itself is unaffected.

examples/kotlin/snippets/runtime/RunConfigExample.kt
```diff
@@ -68,9 +70,11 @@ val metadataConfig =
 // --8<-- [end:custom_metadata]
 
 // --8<-- [start:streaming_config]
+// Kotlin has no supportCfc equivalent; the other tabs set it alongside these.
```
joefernandez:
```suggestion
// Note: Kotlin currently has no supportCfc equivalent
```

joefernandez (approved):
Approved with minor edit

PR #2164, opened 2026-08-20
Add Kotlin to the session rewind page

happyhuman:
## What

`docs/sessions/session/rewind.md` was single-language: one bare Python fenced block, no tab structure, badge div listing Python only. This converts the block into a Python/Kotlin tab group (Python content kept verbatim, only re-indented into the tab), adds the Kotlin badge, and adds a compiled `examples/kotlin/snippets/sessions/RewindExample.kt`.

## Semantics were verified, not assumed

The backlog row asked for this explicitly, and the page makes specific promises. Both check out against the v0.8.0 implementation:

| Page says | Implementation |
| --- | --- |
| "all requests, including rewound requests, are preserved in the log" | `AbstractRunner.rewindAsync` appends a synthetic user-authored event carrying reversing `stateDelta` / `artifactDelta` — nothing is deleted |
| "App-level and user-level state and artifacts are *not* restored" | `computeStateDeltaForRewind` skips keys prefixed `app:` / `user:` |

The KDoc also states it mirrors Python's `runners.py:rewind_async`.

## :warning: One caveat worth a reviewer's eye

`Runner.rewindAsync` is an **interface method whose default implementation throws `NotImplementedError`** — "override to support session rewind". `AbstractRunner` does override it and `InMemoryRunner` extends `AbstractRunner`, so the documented path genuinely works. But anyone implementing `Runner` directly gets the throwing default. I did not put this in the page, since the page only ever discusses `InMemoryRunner`; say the word if you'd like it called out.

## Also in the row, deliberately excluded

`Runner.close` was listed alongside `rewindAsync`. It is a lifecycle concern unrelated to rewinding, so it does not belong on this page.

## Verification

Grounded against the `v0.8.0` git tag. Full ladder green (L0, L1 compile with KSP, L2 ktlint, L3, L5, L6). L4 `runSnippets` SKIP — no such Gradle task in this repo.

Tracked as KT-33.


happyhuman:
**Update:** the badge originally read `Kotlin v0.8.0` (the version adk-docs compiles against). Corrected to `Kotlin v0.3.0`, the release `rewindAsync` actually shipped in — it is absent at v0.1.0 and v0.2.0 and first appears in `AbstractRunner` at v0.3.0. Note this also means the backlog row recording it as a v0.1.0 member was wrong.

wikaaaaa (approved):

joefernandez (approved):
Thanks for the updates!

PR #2172, opened 2026-08-24
Fix TypeScript samples that do not compile against the ADK API

kalenkevich:
## What this fixes

Seven TypeScript snippets across five pages used APIs that don't exist, so
copying them produced a compile error. None of these are cosmetic — each one
fails on `tsc`.

| Page | Problem |
|---|---|
| `agents/custom-agents.md`, `workflows/patterns.md` | ADK helpers imported from `@google/genai` |
| `agents/custom-agents.md` | `createEvent` / `createEventActions` imported with `import type`, then called |
| `context/index.md` | `SearchMemoryResponse.results` / `.text` — neither exists |
| `plugins/index.md` | `BasePlugin` callbacks declared with positional args |
| `tools-custom/mcp-tools.md` | `type: "SseConnectionParams"` is not a valid connection type |

### Details

**Wrong package.** `@google/genai` exports none of `createEvent`,
`createEventActions` or the ADK `Event` — only `Part` genuinely belongs to it.
Those imports never resolved.

**`import type` on runtime functions.** Both helpers are functions and the
samples call them, which fails with TS1361 ("cannot be used as a value because
it was imported using 'import type'").

**Memory API.** `SearchMemoryResponse` exposes `memories`, not `results`, and
each entry is a `MemoryEntry` whose text lives on `content.parts` rather than a
`.text` field. The TypeScript tab had mirrored the Python tab's
`search_results.results` literally.

**Plugin callbacks.** `BasePlugin` takes a single params object:
`beforeAgentCallback({agent, callbackContext})` and
`beforeModelCallback({callbackContext, llmRequest})`. As written the overrides
didn't match the base signature, so they wouldn't have received their
arguments even if they had compiled.

**MCP connection type.** Valid values are `StdioConnectionParams` and
`StreamableHTTPConnectionParams`; the sample's own comment said it meant the
latter. Correcting `type` let the compiler narrow the union, which surfaced a
second problem the wrong type had been masking: headers belong on
`transportOptions.requestInit.headers`, because the top-level `header` field
is deprecated.

## How these were found

Extracted every inline TypeScript block in `docs/` and type-checked it against
the real `@google/adk` types.

One methodology note worth recording, because it made the first pass useless:
`tsc` skips **all** semantic checking when any file in the program has a syntax
error, and ~48 of the blocks are fragments that don't parse standalone. The
initial run therefore reported zero errors. Quarantining the fragments, and
planting deliberately-broken canary code to confirm the checker actually
catches things, is what surfaced these.

Also worth noting for anyone reading the audit: `Event.actions`,
`getFunctionResponses` and `LlmResponse` all looked broken at first but are
fine — those blocks just omit imports, so bare `Event` resolves to the DOM
global.

## Verification

- 102 of 149 inline TypeScript blocks now compile clean against the real API,
  up from 95. The remaining 47 are fragments that don't stand alone.
- All 500 `@google/adk` imports across the docs resolve to real exports.
- Python, Go, Java and Kotlin tabs are untouched — only the TypeScript tabs
  changed, and only where the code was wrong.

## Not included

Left out deliberately, happy to fold in if you'd rather:

1. **Three arguable cases.** `context/index.md` declares concrete agent classes
   without the abstract `runLiveImpl` (won't compile as written, but may be an
   intentional simplification); an `authScheme` placeholder that doesn't
   satisfy `AuthScheme` in a block labelled pseudocode; and a callback in
   `sessions/memory.md` returning `Promise<void>` where `Content | undefined`
   is required — that last one might be better fixed by loosening the ADK type.

2. **Snippet-package config.** Separately from the docs,
   `examples/typescript/snippets/agents/workflow-agents/tsconfig.json` sets
   `include: ["src/**/*"]` but the files sit at the directory root, so `tsc`
   compiles nothing there; that tsconfig and `callbacks/tsconfig.json` also use
   an invalid `ignoreDeprecations` value for the TypeScript version their own
   `package.json` pins. Those belong with a CI job that actually compiles the
   TypeScript snippets — there isn't one today, unlike Go and Kotlin.

### DIFF
--- a/docs/agents/custom-agents.md
+++ b/docs/agents/custom-agents.md
@@ -305,8 +305,8 @@ The foundation for structuring multi-agent systems is the parent-child relations
 
     ```typescript
     // Conceptual Example: Defining Hierarchy
-    import { LlmAgent, BaseAgent, InvocationContext } from '@google/adk';
-    import type { Event, createEventActions } from '@google/adk';
+    import { LlmAgent, BaseAgent, InvocationContext, createEventActions } from '@google/adk';
+    import type { Event } from '@google/adk';
 
     class TaskExecutorAgent extends BaseAgent {
       async *runAsyncImpl(context: InvocationContext): AsyncGenerator<Event, void, void> {
@@ -568,8 +568,8 @@ ADK includes specialized agents derived from `BaseAgent` that don't perform task
 
     ```typescript
     // Conceptual Example: Loop with Condition
-    import { LoopAgent, LlmAgent, BaseAgent, InvocationContext } from '@google/adk';
-    import type { Event, createEventActions, EventActions } from '@google/adk';
+    import { LoopAgent, LlmAgent, BaseAgent, InvocationContext, createEvent, createEventActions } from '@google/adk';
+    import type { Event } from '@google/adk';
 
     class CheckConditionAgent extends BaseAgent { // Custom agent to check state
         async *runAsyncImpl(ctx: InvocationContext): AsyncGenerator<Event> {
@@ -905,8 +905,9 @@ Allows an [`LlmAgent`](llm-agents.md) to treat another `BaseAgent` instance as a
 
     ```typescript
     // Conceptual Setup: Agent as a Tool
-    import { LlmAgent, BaseAgent, AgentTool, InvocationContext } from '@google/adk';
-    import type { Part, createEvent, Event } from '@google/genai';
+    import { LlmAgent, BaseAgent, AgentTool, InvocationContext, createEvent } from '@google/adk';
+    import type { Event } from '@google/adk';
+    import type { Part } from '@google/genai';
 
     // Define a target agent (could be LlmAgent or custom BaseAgent)
     class ImageGeneratorAgent extends BaseAgent { // Example custom agent
--- a/docs/context/index.md
+++ b/docs/context/index.md
@@ -1376,10 +1376,12 @@ Access relevant information from the past or external sources.
     async function findRelatedInfo(context: Context, topic: string): Promise<Record<string, string>> {
       try {
         const searchResults = await context.searchMemory(`Information about ${topic}`);
-        if (searchResults.results?.length) {
-          console.log(`Found ${searchResults.results.length} memory results for '${topic}'`);
-          // Process searchResults.results
-          const topResultText = searchResults.results[0].text;
+        if (searchResults.memories.length) {
+          console.log(`Found ${searchResults.memories.length} memory results for '${topic}'`);
+          // Each entry is a MemoryEntry: its text lives on `content.parts`.
+          const topResultText = searchResults.memories[0].content.parts
+            ?.map((part) => part.text ?? '')
+            .join('') ?? '';
           return { memory_snippet: topResultText };
         } else {
           return { message: 'No relevant memories found.' };
--- a/docs/plugins/index.md
+++ b/docs/plugins/index.md
@@ -140,8 +140,7 @@ methods, as shown in the following code example:
          * Count agent runs.
          */
         async beforeAgentCallback(
-            agent: BaseAgent,
-            context: Context
+            { agent, callbackContext }: { agent: BaseAgent; callbackContext: Context }
         ): Promise<Content | undefined> {
             this.agentCount++;
             console.log(`[Plugin] Agent run count: ${this.agentCount}`);
@@ -152,8 +151,7 @@ methods, as shown in the following code example:
          * Count LLM requests.
          */
         async beforeModelCallback(
-            context: Context,
-            llmRequest: LlmRequest
+            { callbackContext, llmRequest }: { callbackContext: Context; llmRequest: LlmRequest }
         ): Promise<LlmResponse | undefined> {
             this.llmRequestCount++;
             console.log(`[Plugin] LLM request count: ${this.llmRequestCount}`);
--- a/docs/tools-custom/mcp-tools.md
+++ b/docs/tools-custom/mcp-tools.md
@@ -501,14 +501,18 @@ export const rootAgent = new LlmAgent({
     description: "A helpful assistant for planning travel.",
     tools: [
         new MCPToolset({
-            // Using SseConnectionParams to connect to the remote Grounding Lite service,
+            // Connects to the remote Grounding Lite service over streamable HTTP,
             // mirroring Python's StreamableHTTPConnectionParams.
-            type: "SseConnectionParams",
+            type: "StreamableHTTPConnectionParams",
             url: "https://mapstools.googleapis.com/mcp",
-            headers: {
-                "X-Goog-Api-Key": googleMapsApiKey,
-                "Content-Type": "application/json",
-                "Accept": "application/json, text/event-stream"
+            transportOptions: {
+                requestInit: {
+                    headers: {
+                        "X-Goog-Api-Key": googleMapsApiKey,
+                        "Content-Type": "application/json",
+                        "Accept": "application/json, text/event-stream"
+                    }
+                }
             }
         })
     ],
--- a/docs/workflows/patterns.md
+++ b/docs/workflows/patterns.md
@@ -656,8 +656,8 @@ your project requirements before committing to a full implementation.
 
     ```typescript
     // Conceptual Code: Iterative Code Refinement
-    import { LoopAgent, LlmAgent, BaseAgent, InvocationContext } from '@google/adk';
-    import type { Event, createEvent, createEventActions } from '@google/genai';
+    import { LoopAgent, LlmAgent, BaseAgent, InvocationContext, createEvent, createEventActions } from '@google/adk';
+    import type { Event } from '@google/adk';
 
     // Agent to generate/refine code based on state['current_code'] and state['requirements']
     const codeRefiner = new LlmAgent({


joefernandez:
# Technical review report

- **Review Score**: 97
- **Summary reasoning**: Every API correction in this PR checks out against the current ADK TypeScript release, [`adk-v2.1.0`](https://github.com/google/adk-js/releases/tag/adk-v2.1.0) (published 2026-09-15). The `@google/genai` imports it removes were genuinely broken, the `searchMemory` shape it fixes was genuinely wrong, the plugin callbacks really do take a single destructured params object, and `SseConnectionParams` really does not exist in the JS MCP layer. No unreleased APIs are referenced. The two items below are pre-existing defects in snippets the PR touched but did not introduce.
- **Recommended action**: Approve. The change is a net accuracy improvement across five samples and I found no incorrect claim in it. Optionally ask the author to also fix the `escalate` flag in `docs/workflows/patterns.md`, since that snippet is now compile-clean but still documents a loop that can never terminate early.
- Number of lines changed: S (49: 27 added, 22 removed)
- Number of files changed: 5

## Verification of each change

All references below are to the released tag `adk-v2.1.0`.

**`docs/agents/custom-agents.md` (3 hunks) — correct.** `createEvent` and `createEventActions` are runtime *value* exports, so a `import type` of them erases at compile time and the call sites fail; `Event` is a *type-only* export. See [`core/src/common.ts#L163`](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/common.ts#L163), [`#L176`](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/common.ts#L176) and [`#L181`](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/common.ts#L181). The third hunk's split of `Part` back to `@google/genai` is also right: I unpacked `@google/genai` and its `dist/genai.d.ts` contains no `Event` interface and no `createEvent` — only `Part`. The yielded object literal at line 310-317 satisfies `Event`'s required fields (`id`, `invocationId`, `actions`, `timestamp`; everything inherited from `LlmResponse` is optional).

**`docs/context/index.md` — correct.** `SearchMemoryResponse` has a required `memories: MemoryEntry[]`, not an optional `results` ([`memory/base_memory_service.ts#L14-L19`](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/memory/base_memory_service.ts#L14-L19)), and `MemoryEntry` carries text on `content.parts`, not a `.text` field ([`memory/memory_entry.ts#L19`](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/memory/memory_entry.ts#L19)). The `?.map(...).join('') ?? ''` chain is well-typed (`Content.parts` is optional, optional chaining short-circuits the whole chain). `Context.searchMemory(query: string)` matches the call ([`agents/context.ts#L168`](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/agents/context.ts#L168)). This also brings the page in line with the already-correct snippet in `docs/sessions/memory.md:265`.

**`docs/plugins/index.md` — correct.** Both hooks take one destructured params object, not positional args: [`plugins/base_plugin.ts#L217-L220`](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/plugins/base_plugin.ts#L217-L220) and [`#L299-L302`](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/plugins/base_plugin.ts#L299-L302). Parameter names (`callbackContext`, not `context`) and the `Context` type both match.

**`docs/tools-custom/mcp-tools.md` — correct, and the most substantive fix.** `MCPConnectionParams` is a union of only `StdioConnectionParams | StreamableHTTPConnectionParams` ([`mcp_session_manager.ts#L71`](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/tools/mcp/mcp_session_manager.ts#L71)) — `"SseConnectionParams"` is not a valid discriminant in the JS SDK at all, so the old snippet could never have connected. A top-level `headers` key was also invalid; the only such field is the deprecated singular `header`, and `transportOptions.requestInit.headers` is the documented replacement ([`#L53-L65`](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/tools/mcp/mcp_session_manager.ts#L53-L65)). Passing the flat params object straight to the constructor matches [`mcp_toolset.ts#L57-L63`](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/tools/mcp/mcp_toolset.ts#L57-L63).

**`docs/workflows/patterns.md` — correct.** Same `@google/genai` → `@google/adk` import fix; `EventActions` was dropped from the type import and is indeed unused in the snippet.

## 🟡 Medium risk items

- **`docs/workflows/patterns.md:684` — the `StopChecker` sample can never stop the loop.** The snippet computes `shouldStop` and then discards it, emitting `createEventActions()` with no `escalate`, while the comment three lines below claims "Loop stops if QualityChecker outputs 'pass' (leading to StopChecker escalating)". The Python equivalent at [`patterns.md:643`](https://github.com/google/adk-docs/blob/c77cc4ae56455f04201e7636127809111826ccbf/docs/workflows/patterns.md#L643) correctly does `EventActions(escalate=should_stop)`. Pre-existing, not introduced here, but the PR edits this very code block. Suggested fix at [line 684](https://github.com/google/adk-docs/blob/c77cc4ae56455f04201e7636127809111826ccbf/docs/workflows/patterns.md#L684):

    ```suggestion
                    actions: createEventActions({ escalate: true }),
    ```

    (Or drop the `if (shouldStop)` guard and always yield `createEventActions({ escalate: shouldStop })`, which mirrors the Python sample more closely.)

## 🟢 Low risk items

- **`docs/agents/custom-agents.md:323` — `runLiveImpl` calls an async generator without consuming it.** `this.runAsyncImpl(context);` creates a generator object and discards it, so the body never runs; it is also a bare expression statement with no `await`/`yield*`. Pre-existing, in a block the PR touches. The neighbouring samples just leave `runLiveImpl` empty with a comment, which is clearer:

    ```suggestion
      async *runLiveImpl(context: InvocationContext): AsyncGenerator<Event, void, void> {
        // Not implemented for this agent.

joefernandez (commented):
# Technical review report

- **Review Score**: 96
- **Overall assessment**: 🟡 Medium Risk
- **Summary reasoning**: The import and API fixes in this PR are correct against the released ADK TypeScript API ([`adk-v2.1.0`](https://github.com/google/adk-js/releases/tag/adk-v2.1.0)). Two pre-existing defects remain in TypeScript blocks this PR edits: a `LoopAgent` sample that can never terminate, and a `runLiveImpl` that calls an async generator without consuming it. Neither was introduced here, but both sit in code blocks the PR touches and both would mislead a reader copying the snippet.
- **Recommended action**: Approve with these specific changes. The escalate bug in `docs/workflows/patterns.md` is the more important of the two — a reader following that sample gets a loop that only ever exits on `maxIterations`, diverging from the Python sample directly above it and from the comment that claims "StopChecker escalating". The `runLiveImpl` no-op is cosmetic but is a correctness trap worth removing while the block is open.
- Number of lines changed: "M" 49
- Number of files changed: 5

## 🟡 Medium risk items

- **`docs/workflows/patterns.md` (line 684) — the `StopChecker` sample can never stop the loop.** `shouldStop` is computed and then discarded: the emitted `createEventActions()` carries no `escalate`, so `LoopAgent` never terminates early and only exits at `maxIterations: 5`. The comment below the class claims "Loop stops if QualityChecker outputs 'pass' (leading to StopChecker escalating)", and the [Python equivalent](https://github.com/google/adk-docs/blob/c77cc4ae56455f04201e7636127809111826ccbf/docs/workflows/patterns.md#L643) does `EventActions(escalate=should_stop)`. `createEventActions` accepts `escalate` — see [`core/src/events/event_actions.ts#L87`](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/events/event_actions.ts#L87) and the field at [L42](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/events/event_actions.ts#L42) — and the sibling sample at [`docs/agents/custom-agents.md#L578`](https://github.com/google/adk-docs/blob/c77cc4ae56455f04201e7636127809111826ccbf/docs/agents/custom-agents.md#L578) already uses that form. Suggested replacement for [lines 681-686](https://github.com/google/adk-docs/blob/c77cc4ae56455f04201e7636127809111826ccbf/docs/workflows/patterns.md#L681-L686), mirroring the Python sample:

    ```typescript
            yield createEvent({
                author: 'StopChecker',
                actions: createEventActions({ escalate: shouldStop }),
            });
    ```

## 🟢 Low risk items

- **`docs/agents/custom-agents.md` (line 323) — `runLiveImpl` calls an async generator without consuming it.** `this.runAsyncImpl(context);` builds a generator object and drops it, so the body never executes; it is also a bare expression statement with no `await` or `yield*`. `runLiveImpl` is declared as an `AsyncGenerator` in [`core/src/agents/base_agent.ts#L393`](https://github.com/google/adk-js/blob/adk-v2.1.0/core/src/agents/base_agent.ts#L393), so an empty body is legal and is what the neighbouring samples in this file do. Suggested replacement for [line 323](https://github.com/google/adk-docs/blob/c77cc4ae56455f04201e7636127809111826ccbf/docs/agents/custom-agents.md#L323):

    ```typescript
        // This agent doesn't have a live implementation.
    ```

---

No inline suggestions are attached: both lines fall outside this PR's diff hunks (the `patterns.md` hunk covers lines 656-663 and the `custom-agents.md` hunks cover 305-312, 568-575 and 905-913), so GitHub would reject line comments there.

<sub>Technical-accuracy review generated by CloudCode (session `ses_e5f488b44aaffeVRrkb86lsH3N`), verified against ADK TypeScript release [`adk-v2.1.0`](https://github.com/google/adk-js/releases/tag/adk-v2.1.0).</sub>

joefernandez:
@kalenkevich - Hey Alexey. These changes look OK to me except for a couple items, but I can't make any changes because you did not allow repo owners to make changes.

Please resolve the commented issues and we will plan to close this in 14 days.

PR #2191, opened 2026-09-01
Upgrade Kotlin docs to adk-kotlin 0.9.0 and move off AdkWebServer

sherryfox:
**adk-kotlin 0.9.0** split the web server into `AdkApiServer`, which serves the agent runtime contract headlessly, and `AdkDevServer`, which adds the development UI. `AdkWebServer` is deprecated in that release and goes away at 1.0, so `WebMain.kt` as written here stops compiling the day 1.0 ships.

The quickstart snippet now builds an `AdkDevServer` from an `AdkServerConfig`. That drops three imports: `AdkServerConfig.inMemory()` supplies the agent loader and the in-memory session and artifact services the old constructor took one by one.

Two behaviour notes come with the split. `AdkWebServer` pinned host to `0.0.0.0`; the new classes bind loopback, which is a visible change for anyone reaching the server from a container or a remote box, so the page now says so and shows the host override. And `AdkApiServer` is the headless half of the same config, so it gets a short section rather than a bare mention - the bundled Kotlin API reference is still generated at 0.5.0 and documents none of these classes, so a link there would not have helped. The "not meant for production" warning now sits directly after the screenshot, where go.md, java.md and python.md put it.

The dependency blocks readers copy from still pinned **0.8.0**, and so did the examples project. Verified on Maven Central that **0.9.0** is published for every artifact named across these pages: -core, -processor, -webserver, -litertlm, -a2a and -integrations. Checked the **0.9.0** POMs before bumping: Ktor still resolves to 2.3.13 and a2a-java-sdk-client to **1.0.0.Final**, so the explicit pins stay correct and the two comments citing them only needed their version reference moved. The `Kotlin v0.x` support badges are deliberately left alone: they record the release a feature landed in, not the current version.

Verified by compiling: every snippet added here was compiled verbatim against the published **0.9.0** artifacts on JDK 17, and the whole examples project still builds at **0.9.0**. As a negative control, the old snippet still compiles at **0.9.0** but emits the deprecation warning, which is what makes this a **1.0** break rather than a present-day one. Note the Kotlin snippet check only builds .kt files changed in the PR, so it will not cover a markdown-only change like this.


happyhuman (approved):

docs/get-started/kotlin.md
```diff
@@ -289,6 +278,31 @@ upper left corner and type a request.
     ADK Web is ***not meant for use in production deployments***. You should
     use ADK Web for development and debugging purposes only.
 
+!!! note "The server binds loopback"
+
+    `AdkDevServer` listens on `127.0.0.1:8080`, because its endpoints are
```
joefernandez: don't start sentences with code syntax:

```suggestion
    The ADK web server listens on `127.0.0.1:8080`, because its endpoints are
```

docs/get-started/kotlin.md
```diff
@@ -289,6 +278,31 @@ upper left corner and type a request.
     ADK Web is ***not meant for use in production deployments***. You should
     use ADK Web for development and debugging purposes only.
 
+!!! note "The server binds loopback"
```
joefernandez: This title should explain the note more clearly:

```suggestion
!!! note "ADK Web server loopback binding"
```

docs/get-started/kotlin.md
```diff
@@ -289,6 +278,31 @@ upper left corner and type a request.
     ADK Web is ***not meant for use in production deployments***. You should
     use ADK Web for development and debugging purposes only.
 
+!!! note "ADK Web server loopback binding"
```
joefernandez: Remove this note and the next section.

While this is good information, the effect of putting this here is to bloat the Getting Started guide with information that will overwhelm, not help, developers.  

This note should be placed with the Web interface docs:
1. https://adk.dev/runtime/web-interface/ and that should then point to the API Server docs

And the section should be added in the API Server docs
2. https://adk.dev/runtime/api-server/#deploy-your-agent - add the "Serve without the development UI" as a subheading of the "Deploy your agent" section of the page.

joefernandez (approved):

docs/get-started/kotlin.md
joefernandez: resolved by removing content; follow up recommended

PR #2086, opened 2026-08-07
docs(live): decompose the dev guide and fix staleness vs adk-python main

kazunori279:
## Summary

Phase 2 of the Live docs revamp: the five-part `docs/streaming/dev-guide/` is decomposed into per-capability pages under `docs/live/`, the whole set is re-verified against adk-python `main`, and #2150 lands a quality pass on top.

Twelve commits, meant to be read in order:

| Commit | What |
|---|---|
| `efa0f54` | Decompose the development guide into capability pages (index, sessions, events, configuration, voice, audio-video, tools, workflows, custom-server, models) |
| `6cd6256` | Drop half-cascade model coverage — no longer supported |
| `fa8c9a1` | Retire the five-part dev guide and rewire navigation + redirects |
| `e852ac5` | Point at the Python API reference instead of pinned adk-python source where a reference page exists |
| `83cb5c0` | Fix staleness against adk-python `main` and drop every link to the bidi-demo sample |
| `b384907` | Refresh the Live API supported-model list |
| `98c75d4` | Align the docs with what the current Live API models actually support |
| `5a28678` | Trim the response-modality and SSE material |
| `6e45ad6` | Cut duplicated and non-ADK material |
| `08fb638` | **#2150** — restructure the live docs around ADK ownership |
| `8ea417b` | **#2150** — standardize page leads and cut duplicated RunConfig prose |
| `363fe0b` | **#2150** — reframe pages around capabilities, fix eval config key |

Net: 25 files, **+2,332 / −5,680**. Rebased onto `main`.

## Quality pass (#2150, @allen-stephen)

Merged into this branch; it closes every point from Stephen's review of the earlier state of this PR.

| Review comment | Resolution |
|---|---|
| Why `Build` / `Ship` instead of the design doc's `Building` / `Production`? | Nav renamed to **Building / Production** |
| Collapse the new `Voice` page into `Configuration` | `voice.md` deleted, content folded into `configuration.md`, redirect added |
| `Workflows` documents 1.x agents, not 2.0 workflows | Page now leads with the ADK 2.0 graph `Workflow` (`START` / `edges` / `mode='task'`, `live_workflow` sample). `SequentialAgent` / `LoopAgent` / `ParallelAgent` demoted to a *Legacy workflow agents* section with the deprecation and the `NotImplementedError`-under-`run_live()` caveat |
[...]

PR #2194, opened 2026-09-02
Fix code samples that do not compile against the shipped SDKs

GWeale:
## What this is

Code samples across the docs were checked against the real published library
for each language, and this corrects what does not compile or resolve. 21 files.

Nothing was executed. The samples call live Gemini and Vertex endpoints, so
every check is static: parsing, symbol resolution, and compilation or type
checking against the real artifacts.

## Verified against

| Language | Version | Where that version comes from |
|---|---|---|
| Python | `google-adk[all]` 2.8.0 | latest on PyPI |
| TypeScript | `@google/adk` 2.0.0 | latest on npm |
| Java | `google-adk` 1.6.0 | `examples/java/snippets/pom.xml` |
| Kotlin | `adk-kotlin` 0.8.0 | `examples/kotlin/build.gradle.kts` |
| Go | `google.golang.org/adk/v2` 2.3.0 | latest module release |

`mkdocs build --strict` passes. 21 rendered pages change, each matching an
edited source file. Tab structure and code-fence rendering were compared
against `main` page by page, because the strict build does not detect a code
block that falls out of its tab.

## Go: `tool.Context` does not exist

Nine sites across six pages declared tool and callback functions taking
`ctx tool.Context`. The `tool` package has no `Context` type — not in 2.3.0,
and not in 2.0.0, which is what `examples/go/go.mod` pins, so this was never
valid in the v2 line.

The type is `agent.Context`, which carries exactly the surface these samples
use: `Actions()`, `State()`, `Artifacts()`, `FunctionCallID()`,
`SearchMemory()`. This repository's own Go examples already use
`ctx agent.Context` 45 times.

Two import blocks were adjusted to match, and four others in
`docs/context/index.md` omitted the `fmt` they call.

## Java: two impossible imports and four wrong types

`com.google.adk.agent.LlmAgent` — the package is `agents`, plural.
`com.google.adk.agents.Content` — `Content` is a genai type.

Four type errors, each confirmed with `javap` against the 1.6.0 jar:

- `EventActions.stateDelta()` returns `Map`, not `ConcurrentMap`.
- `EventActions.artifactDelta()` returns `Map<String, Integer>`, mapping
  filename to version number. The docs declared `ConcurrentMap<String, Part>`,
  wrong in both the map type and the value type.
- `FunctionResponse.response()` yields `Map<String, Object>`, not
  `Map<String, String>`.
- `BaseArtifactService.loadArtifact` takes the version as an `int` or omits it;
  the docs passed `Optional.empty()` and `Optional.of(0)`, matching no overload.

## Python: coroutines used without await, and a field that does not exist

`search_memory`, `load_artifact`, `save_artifact` and `list_artifacts` are all
`async`. Four samples called them from a plain `def` and read attributes off the
returned coroutine.

Fixing that exposed a second defect underneath: the memory sample then read
`search_results.results`, and `SearchMemoryResponse` has no such field. It is
`memories`, holding `MemoryEntry` objects whose text is at
`entry.content.parts[].text`. `docs/sessions/memory.md` already had this right.
**The TypeScript and Java tabs of the same example had the same mistake**, and
are corrected too.

Also: `CodeExecutionInput` imported from `google.adk.code_executors` rather than
`...code_executors.code_execution_utils`; a `calendar_tool_set` object that does
not exist, replaced by the real `CalendarToolset` class, in a sample that also
referenced two variables it never defined; two positional `Part.from_text` calls
against a keyword-only signature; five `LlmAgent` samples missing the required
`name`, which raises a `ValidationError`; and an external-access-token sample
built on `AuthCredentialTypes.GOOGLE_CREDENTIALS` and a
`google_credentials_config` field, neither of which the package defines. The
real setting is `external_access_token_key` on a toolset's credentials config.

## Samples that could not parse at all

An unindented plugin class body, which was the opening sample on the Plugins
page and raised `IndentationError` on the line after the class statement. Two
`)]` bracket typos. A backslash-escaped Java text-block delimiter. A truncated
`EventsCompactionConfig(` call. An `await` in a non-async function, and another
dedented out of the `if` meant to guard it. A mid-file Java import. `#` comments
opening Go and Java blocks where the TypeScript tab correctly used `//`.

One fence in `docs/integrations/application-integration.md` opened at six spaces
and closed at eight, so it never terminated and the page rendered a literal
```` ```java ```` as body text. That tab is a real code block now.

## Known defects this does not fix

Found while checking, left alone because the right correction needs a decision
from whoever owns the page:

- `docs/sessions/memory.md` has a Go block mixing a package-level `func`
  declaration with a `:=` statement, which is not valid at file scope. The
  correction needs the sample restructured, and the variable renamed, since it
  currently shadows the imported `agent` package.
- `docs/sessions/session/index.md` calls the `suspend` functions
  `createSession` and `deleteSession` from a top-level Kotlin context.
- `docs/graphs/dynamic.md` has a Python node that yields and then does
  `return code`, which is a syntax error in an async generator.
- `docs/integrations/gke-code-executor.md` builds an `InvocationContext()` with
  no arguments where three fields are required.
- `examples/python/snippets/tools/overview/doc_analysis.py` does not compile:
  `await` inside a plain `def`. It is included by `docs/tools-custom/index.md`.
- Several TypeScript plugin callbacks are documented with positional parameters
  where the API takes a single options object.

## Overlap with my other open pull requests

Twelve of these files are also touched by pull requests I already have open:
#2024 (`docs/context/`), #2025 (`docs/tools-custom/`), #2023
(`docs/live/`), #2021 (`docs/agents/`) and #2019
(`docs/integrations/gke-code-executor.md`). Those were opened earlier and cover
different defects on the same pages.

If it is easier to review, I am happy to close this and fold each change into
the matching open pull request instead. Say which you prefer.

## A gap this surfaced

`python-lint.yaml` and `python-tests.yaml` are filtered to `samples/python/**`,
and no `samples/` directory exists, so on pull requests neither can ever run. No
Python, Java or TypeScript sample is checked by CI. That is why these defects
survived. Happy to follow up with a workflow that runs these checks.

### CODE REVIEW

File: docs/tools-custom/authentication.md [hunk 1]
```diff
@@ -281,25 +281,27 @@ Pass the scheme and credential during toolset initialization. The toolset applie
       )
       ```
 
-#### Use Google API toolsets (e.g., `calendar_tool_set`)
+#### Use Google API toolsets (e.g., `CalendarToolset`)
```
joefernandez: simplify heading: don't use parenthesis, "e.g.," and avoid code syntax in headings

```suggestion
#### Use Google API toolsets
```

joefernandez (approved):
Approved. Please resolve conflicts for Live content.

These pages where rewritten in:
https://github.com/google/adk-docs/pull/2086

PR #2196, opened 2026-09-02
Move the code samples out of the markdown into snippet files

GWeale:
Every fenced code block on the docs pages becomes a `pymdownx.snippets`
include pointing at a file under `examples/inline/`, so each sample lives in a
real file with a real extension instead of inside prose.

**This is large and mechanical: 1,247 files, 23,308 added, 22,285 removed.**
Almost all of it is the same code moving. The markdown loses 22,285 lines and
gains 1,065 pointer lines; those lines reappear as 1,063 new files. I would
understand a preference to take it in sections rather than at once, and I am
happy to split it that way — say the word and I will close this in favour of a
smaller proposal against one part of the tree.

## The rendered site does not change

I built the site before and after and compared the output trees. All 808 pages
are byte identical, as are `llms-full.txt` and the generated per-page markdown.

I also compared tab structure and code fence rendering page by page, because
`mkdocs build --strict` does not detect a code block that has fallen out of its
tab: an unindented line inside a `=== "Python"` block closes the tab early and
the fence below renders as literal text, and the build still exits 0. That check
is what caught two such breaks while preparing this, so it is not hypothetical.

## The link checker is widened in the same change

`link-checker.yaml` scans `'./**/*.md'`. 126 URLs currently sit inside fenced
code blocks, and they would silently stop being checked the moment the code left
markdown. The glob now also covers `./examples/inline/**`.

This is the one behavioural regression the move introduces, and fixing it here
rather than later seemed better than leaving it to be discovered by a dead link.

## Two deliberate oddities

**Go snippets use a `.go.txt` extension.** `gofmt` runs over every changed
`*.go` file, and every `*.go` under `examples/go` must be registered in a build
manifest. A fragment lifted out of prose satisfies neither. The existing Go and
Kotlin snippet guards still pass, and the change adds no new `*.go` file. The
cost is that these snippets stay invisible to Go tooling.

**Snippet files end without a trailing newline.** `pymdownx.snippets` appends an
extra blank line for a file that ends in one, which renders as a stray line at
the bottom of every block. This is load-bearing: adding the conventional newline
changes two rendered pages. An `end-of-file-fixer` hook would break the site.

## What I checked

- 1,065 references resolve; 1,063 files, none orphaned in either direction.
- No empty snippets, no case-insensitive path collisions, no Windows-reserved
  names, no paths over 200 characters, all valid UTF-8.
- No fence mixes a pointer with other content.
- `check_paths: true` is already set, so a missing snippet is a hard build
  failure rather than a silently empty code block.

## Known gaps

- **No license headers** on the new files, where 56 of 66 existing
  `examples/python` files have them. Whole-file includes mean a header would
  render into the page. Getting both needs `[start:]`/`[end:]` section markers
  on every file, which is a meaningfully larger change. I can do it if you want
  it before this lands.
- One block in `docs/integrations/application-integration.md` is skipped: its
  fence opens at six spaces and closes at eight, so it never terminates. #2194
  fixes that fence; once it lands this can pick up the block.
- This will conflict with #2194, which edits the same markdown. Whichever lands
  second needs a rebase, and I am happy to do it.


docs/integrations/a2ui.md
joefernandez, line 1: We do not want to manage code snippets for integrations as separate files. Integrations are mostly contributed by third parties and we need to keep the authoring for them as simple as possible.

revert all the changes to the integrations/** pages and remove the separate code files

examples/inline/go/2.0/index/002-event-construction-session-newevent-sign.go.txt
```diff
@@ -0,0 +1,7 @@
+// Before (ADK Go 1.x)
```
joefernandez, line 1: GLOBAL: Revert all "*.txt" inclusions.

??? This".txt" file isn't even compilable code. If you are not **actually** going make the code runnable, then leave the code snippet in the page and don't pull it into a separate file. The net effect of ".txt" inclusions is to massively increase the maintenance costs of this content with zero benefit.

docs/a2a/a2a-extension.md
```diff
@@ -31,13 +31,7 @@ To activate the extension, the client can instantiate the `RemoteA2aAgent` with
...
 
 ```python
-from google.adk.agents.remote_a2a_agent import RemoteA2aAgent
-
-remote_agent = RemoteA2aAgent(
-    name="remote_agent",
-    agent_card="http://localhost:8000/a2a/remote_agent/.well-known/agent-card.json",
-    use_legacy=False,
-)
+--8<-- "examples/inline/python/a2a/a2a-extension/001-client-side-extension-activation.py"
```
joefernandez, line 34: GLOBAL: Inclusions are not specific (global fix): In order to not bloat pages unnecessarily, the inclusions need to be targeted and not include the entirety of the source code file. You do this by inserting tags in the source code file to mark the beginning and ending of the code you want to include (and skip all the boilerplate stuff).

Not doing this will bloat the documentation pages with lines and lines boilerplate code and wind up frustrating users

examples/inline/typescript/sessions/session/index/006-inmemorysessionservice.ts
joefernandez, line 1: Is this TypeScript code actually compilable?

What is the benefit of having such a small amount of code stored separately like this?

examples/inline/go/agents/custom-agents/011-agent-hierarchy-parent-agents-and-sub-ag.go.txt
```diff
@@ -0,0 +1,6 @@
+import (
+    "google.golang.org/adk/v2/agent"
+    "google.golang.org/adk/v2/agent/llmagent"
+)
+
+--8<-- "examples/go/snippets/agents/multi-agent/main.go:hierarchy"
```
joefernandez, line 6: Chaining includes makes this content significantly harder to maintain because now you have to follow the includes to update the whole code snippet. Can we make this simpler so code examples only have one file?

examples/inline/go/agents/custom-agents/005-key-capabilities-within-the-core-asynchr.go.txt
joefernandez, line 1: What's showing up, here is that some pages, like custom-agents, probably have too many code snippets on them and those snippets need to be culled and simplified to reduce the maintenance cost and possibility of code errors.

sesaymohsesay11-design (approved):

docs/2.0/index.md
```diff
@@ -347,13 +340,7 @@ logic into the execution lifecycle.
...
 
 ```go
-// Before (ADK Go 1.x)
-ev := session.NewEvent(ctx.InvocationID())
-// or
-ev := session.NewEventWithContext(ctx, ctx.InvocationID())
-
-// After (ADK Go 2.0)
-ev := session.NewEvent(ctx, ctx.InvocationID())
+--8<-- "examples/inline/go/2.0/index/002-event-construction-session-newevent-sign.go.txt"
```
joefernandez, line 343: similar comment here. too much for no appreciable gain, and added maintenance cost.

docs/2.0/index.md
```diff
@@ -260,14 +260,7 @@ that reads this property without handling `undefined` no longer compiles under
...
 ```typescript
-// Before (ADK TypeScript 1.x)
-const name = ctx.agent.name;
-
-// After (ADK TypeScript 2.0), inside an agent's own execution
-const name = requireAgent(ctx).name;
-
-// After (ADK TypeScript 2.0), outside an agent's own execution
-const name = ctx.agent?.name;
+--8<-- "examples/inline/typescript/2.0/index/001-context-invocationcontext-agent-is-optio.ts"
```
joefernandez, line 263: This change is complete overkill for 6 lines of code. Remove

joefernandez:
**DO NOT MERGE**: Given the size of this change and it's potential impact on future content authoring, this change is on hold pending offline discussions about how this change provides authoring and or technical accuracy benefits.

PR #2201, opened 2026-09-04
Updating API reference links to be relative

joefernandez:


joefernandez:
Close. This breaks the links both local stages and production

PR #2203, opened 2026-09-04
docs(integrations): add CopilotKit integration page

mme:
## Summary

Adds an integration page for [CopilotKit](https://github.com/CopilotKit/CopilotKit) to the catalog at `docs/integrations/copilotkit.md`, with its icon in `docs/integrations/assets/`.

CopilotKit is an open source set of frontend libraries and a runtime that connect applications to agents over AG-UI. With ADK, the `ag-ui-adk` package exposes an agent as an AG-UI endpoint, and CopilotKit gives that endpoint a chat surface, frontend tools, generative UI, and human-in-the-loop controls in React, Angular, Vue, React Native, or Slack.

## What the page covers

- **Use cases**: chat, frontend tools, generative UI, human-in-the-loop, and messaging channels
- **Prerequisites and installation**: Python and npm packages
- **Use with agent**: a four-step walkthrough that exposes an ADK agent over AG-UI, registers it with CopilotKit Runtime, renders chat in React, and adds a browser-side tool the agent can call
- **Available hooks**: a table of the React hooks and what each one does
- **Other clients**: Angular, Vue, and React Native, which share the same runtime route and endpoint
- **Messaging channels**: connecting a Slack workspace to the same endpoint with the open source Channels SDK
- **Additional resources**

The existing AG-UI page covers the protocol and a `create` quickstart for new projects. This page links to it and focuses on adding CopilotKit to an existing ADK project, so the two do not overlap.

## Notes for reviewers

- `catalog_tags` is left out because none of the existing tags describe a frontend integration, and the AG-UI and A2UI cards carry none either. Happy to add one if you have a preference.
- Model strings use the `gemini-flash-latest` alias, the agent variable is `root_agent`, and internal links are site-relative.

## Verification

- `mkdocs build --strict` passes and the card renders in the catalog
- The Python sample was run against the published `google-adk` and `ag-ui-adk` packages; the endpoint starts and streams `RUN_STARTED` for the documented request
- Every TypeScript import and option was checked against the type declarations of the published CopilotKit packages
- All external links resolve


joefernandez (approved):
Approved. Thanks for adding this!

PR #2208, opened 2026-09-07
docs(live): restore Gemini 3.1 Flash Live and fix broken anchors

kazunori279:
Follow-up to #2086, reviewing the models content and the redirect/link changes as requested.

## Limits and quotas: no change needed

I checked the new **Platform limits and quotas** table against upstream and it is right, so it stays as merged:

| Limit | Verified |
|---|---|
| Session duration, audio-only — 15 min (both backends) | ✅ |
| Session duration, audio + video — 2 min (both backends) | ✅ |
| Connection lifetime — ~10 min | ✅ |
| Agent Platform conversation session default — 10 min | ✅ |
| Concurrent sessions — 1,000/project pay-as-you-go, unlimited with Provisioned Throughput | ✅ |

My earlier numbers were the wrong ones: I had Agent Platform capping *every* session at 10 minutes, which conflated the connection limit with the session limit.

## Models: one real regression

`gemini-3.1-flash-live-preview` was dropped from the page, but the model exists — released 2026-03-26, [documented on AI Studio](https://ai.google.dev/gemini-api/docs/models/gemini-3.1-flash-live-preview), and listed alongside 2.5 in the [Live API capabilities guide](https://ai.google.dev/gemini-api/docs/live-api/capabilities). Agent Platform supports no Gemini Live 3.x model, which is what makes the AI-Studio-only distinction worth stating rather than omitting.

This PR restores it inside the merged structure rather than reverting:

- Two model rows with launch stages. `gemini-live-2.5-flash-native-audio` is the only GA Live model; every AI Studio ID is preview.
- The per-model feature table gets a second column. With one column it conveyed nothing; with two it answers the question a reader actually has. 3.1 supports neither proactivity/affective dialog nor non-blocking tools, and uses `thinking_level` instead of `thinking_budget`.
- The `global` location note goes back to stating the restriction as fact — Live 2.5 models are not served at `GOOGLE_CLOUD_LOCATION=global` — instead of asking the reader to go verify.

## Link audit

Checked `docs/**/*.md` for references to `streaming/dev-guide/*` and validated the redirect map:

- Zero references to `streaming/dev-guide/*` or `live/dev-guide/*` remain in `docs/`.
- All 107 entries in `redirect_maps` resolve: no source shadows a real page, every target file exists, and all 107 generated redirect stubs in `site/` point at a page that was built.
- No links remain to the removed `live/streaming-tools.md` or `live/voice.md`.

Two broken anchors did turn up, both fixed here:

- `live/configuration.md` linked to `#response-modalities`; the heading is `## Response modes`.
- `live/evaluation.md` linked to `#audio-user-simulation-live-agents`; the heading is `## Audio user simulation for live agents`.

Plus one stale model ID outside the live section: `tutorials/multi-tool-agent.md` suggested `gemini-2.0-flash-live-001`, shut down 2025-12-09.

`mkdocs build --strict` is clean, with no remaining link or anchor warnings.


docs/live/models.md
```diff
@@ -14,14 +14,18 @@ Live agents run on models that take audio in and produce audio out, end to end,
...
-
 | Model | AI Studio | Agent Platform |
 |-------|-----------|----------------|
-| Gemini 2.5 Flash Live | `gemini-2.5-flash-native-audio-preview-12-2025` | `gemini-live-2.5-flash-native-audio` |
+| Gemini 2.5 Flash Live | `gemini-2.5-flash-native-audio-preview-12-2025` (Preview) | `gemini-live-2.5-flash-native-audio` (GA) |
+| Gemini 3.1 Flash Live | `gemini-3.1-flash-live-preview` (Preview) | Not available |
+
+Gemini 2.5 Flash Live is one model with a different ID on each backend; the features are the
+same either way. `gemini-live-2.5-flash-native-audio` is ADK's `LlmAgent.DEFAULT_LIVE_MODEL`,
+the only Live model that is GA, and the model used in this section's examples.
```
joefernandez: should NOT refer to a Preview model as "GA"  
```suggestion
the only Live model that is publicly available, and the model used in this section's examples.
```

joefernandez (approved):

PR #2211, opened 2026-09-08
docs(integrations): document Go support for the Postman MCP server

wolo-lab:
## Problem

`docs/integrations/postman.md` carries Python and TypeScript tabs only, so Go users have no documented path to the Postman MCP server. Nothing is missing in adk-go to support it: both transports the page describes are already covered by `google.golang.org/adk/v2/tool/mcptoolset`.

## Summary

- Adds the `lst-go` language-support span and a **Go** tab with the same **Local MCP Server** and **Remote MCP Server** sub-tabs as the other two languages.
- The local sample runs the server through `mcp.CommandTransport` over `exec.CommandContext(ctx, "npx", args...)`, and passes the key in `cmd.Env` layered on `os.Environ()` so `npx` keeps its `PATH`.
- The remote sample uses `Config.Endpoint` with `Config.Auth`. Passing `auth.StaticToken` sets the same `Authorization: Bearer <key>` header that the Python and TypeScript samples write by hand.
- Optional server flags live in an `args` slice, so the existing Configuration wording ("add `--full` or `--code` to the `args` list") reads correctly for Go as well and needed no change.

## Verification

- Both samples compile, pass `go vet`, and are `gofmt`-clean against `google.golang.org/adk/v2@v2.3.0`. `Config.Endpoint` and `Config.Auth` shipped in v2.1.0, so the remote sample needs that version or later.
- `mkdocs build --strict` exits 0. The built page renders three top-level tabs (Python, TypeScript, Go), each with both sub-tabs, and two `language-go` code blocks.

Staged: https://deploy-preview-2211--adk-docs-preview.netlify.app/integrations/postman/


docs/integrations/postman.md
```diff
@@ -174,6 +174,140 @@ natural language interactions.
...
+        		log.Fatalf("Failed to create the model: %v", err)
+        	}
+
+        	args := []string{"-y", "@postman/postman-mcp-server"}
+        	// args = append(args, "--full")          // Use all 100+ tools
+        	// args = append(args, "--code")          // Use code generation tools
+        	// args = append(args, "--region", "eu")  // Use EU region
+
+        	server := exec.CommandContext(ctx, "npx", args...)
+        	server.Env = append(os.Environ(), "POSTMAN_API_KEY="+postmanAPIKey)
```
joefernandez: **This hands the reader's entire environment — including `GOOGLE_API_KEY` — to a third-party npm package.**

`os.Environ()` copies *every* variable from the parent process into the child started on [line 219](https://github.com/google/adk-docs/blob/f3ce80924157d4a131384a0a7345f0e3310497b0/docs/integrations/postman.md#L219), which is `npx -y @postman/postman-mcp-server` — code fetched from the npm registry and executed at run time. That environment includes the `GOOGLE_API_KEY` this same snippet reads twelve lines earlier ([line 208](https://github.com/google/adk-docs/blob/f3ce80924157d4a131384a0a7345f0e3310497b0/docs/integrations/postman.md#L208), `APIKey: os.Getenv("GOOGLE_API_KEY")`), plus every other secret the reader happens to have exported.

The sibling samples on this page don't do this. Python passes a *replacement* env holding only the Postman key ([lines 69-71](https://github.com/google/adk-docs/blob/f3ce80924157d4a131384a0a7345f0e3310497b0/docs/integrations/postman.md#L69-L71)), and TypeScript does the same ([lines 134-136](https://github.com/google/adk-docs/blob/f3ce80924157d4a131384a0a7345f0e3310497b0/docs/integrations/postman.md#L134-L136)). It's also the only `os.Environ()` forwarding anywhere in the docs — `grep -rn "os.Environ()" docs/` returns just this one line — so the Go tab would be introducing a pattern readers copy-paste into their own agents.

Suggested change, which keeps `npx` working while scoping the secrets:

```suggestion
        	server := exec.CommandContext(ctx, "npx", args...)
        	// Pass only what the server needs; don't forward the parent environment,
        	// which may hold unrelated credentials such as GOOGLE_API_KEY.
        	server.Env = []string{
        		"PATH=" + os.Getenv("PATH"),
        		"HOME=" + os.Getenv("HOME"),
        		"POSTMAN_API_KEY=" + postmanAPIKey,
        	}
```

`PATH` so `npx` resolves, `HOME` so npm can find its cache/config. I checked the replacement with `gofmt -l` (clean) and `go build` (compiles). Worth noting for Windows readers that `npx` there also wants `USERPROFILE`/`APPDATA`, if you'd rather phrase the allowlist more portably.

joefernandez (approved):
Approved, but please address comment

docs/integrations/postman.md
wolo-lab: Applied, with two changes to the suggested block so the sample keeps working on Windows.

The names are looked up in a loop because `os.LookupEnv` skips whatever isn't set, so one list covers both platforms with no build tag and no second code path in the doc. `LOCALAPPDATA` is in the Windows set alongside the two you named, because npm derives its cache root from it there — `const cacheRoot = (isWindows && process.env.LOCALAPPDATA) || '~'` ([definitions.js L84](https://github.com/npm/cli/blob/c9876d7ea7150b0702e4151210b9fa1a8dbc7fbf/workspaces/config/lib/definitions/definitions.js#L84)) — so leaving it out silently relocates the cache to the home directory. `SYSTEMROOT` is deliberately absent, since Go's `os/exec` appends it on Windows whenever `Env` lacks it ([exec.go L1314-L1329](https://github.com/golang/go/blob/go1.26.6/src/os/exec/exec.go#L1314-L1329)).

Verified on Linux only: the sample compiles against adk v2.3.0, and `npx -y @postman/postman-mcp-server` starts and reports 42 tools under this environment. The Windows names come from the MCP SDK inherit lists rather than from a run on Windows, so that part is worth a check by anyone here who has one.

joefernandez:
@wolo-lab - Thanks for the update! Merged and published.

PR #2212, opened 2026-09-08
Regenerate the Kotlin API reference for adk-kotlin 1.0.0

happyhuman:
The published Kotlin API reference under `docs/api-reference/kotlin/` was still generated from adk-kotlin **0.5.0**, several releases behind the version the rest of the docs tell people to depend on. This regenerates it from **v1.0.0**.

## The generator needed fixing first

`tools/kotlin-api-docs/generate.sh` no longer worked against current adk-kotlin, so running it was not enough:

- **Dokka V2.** adk-kotlin now builds Dokka in V2 mode, where the script's `dokkaHtmlMultiModule` task fails outright with *"Cannot run Dokka V1 tasks when V2 mode is enabled."* The script now runs `dokkaGenerateHtml` and copies from `build/dokka/html` instead of `build/dokka/htmlMultiModule`.
- **Aggregation is no longer implicit.** In V2 the root project only aggregates modules declared as `dokka` dependencies, and adk-kotlin does not declare them. Without that the root task succeeds but emits a site with no modules in it. The script appends `dependencies { subprojects.forEach { dokka(it) } }` to the root build file of its throwaway clone.

The script still clones adk-kotlin itself into a temp directory, so nothing outside the repo is touched.

## What changed in the output

The regenerated site covers four modules that did not exist at 0.5.0: `firebase`, `mlkit`, `examples-android` and `examples-java`.

## Known gap

Dokka only links modules that produce a publication HTML, so `firebase`, `mlkit` and `examples-android` have generated pages that the root `index.html` does not link to. They are reachable by URL but not by navigation. Worth deciding separately whether to exclude them from the build or get them linked — this PR does not change it either way.

## How this was generated

```
export ANDROID_HOME="$(brew --prefix)/share/android-commandlinetools"
bash tools/kotlin-api-docs/generate.sh 1.0.0
```


joefernandez (approved):
Thanks for the update

tools/kotlin-api-docs/generate.sh
```diff
@@ -76,16 +76,25 @@ echo "Cloning adk-kotlin v${VERSION}..."
 git clone --depth 1 --branch "v${VERSION}" https://github.com/google/adk-kotlin adk-kotlin
 cd adk-kotlin
 
-# Build Dokka HTML docs (multi-module generates a unified site with module index)
+# Aggregate every subproject into the root project's Dokka publication. adk-kotlin
```
joefernandez: Consider pulling in the script changes from the previous iteration that excludes the examples package and makes some other updates:
- https://github.com/google/adk-docs/pull/2204/changes#diff-1b8399620f173980311f057705c7bfd04a90de5cc364039c45415a76a890bacf

PR #2214, opened 2026-09-09
Update outdated BigQuery integration sample code

wukath:
Fixes the outdated sample code on the [BigQuery integration page](https://adk.dev/integrations/bigquery/), reported in an ADK friction log. 

Related task: b/559149787

### Deprecated import path

All five auth snippets and the included sample imported from `google.adk.tools.bigquery`. That module is now a back-compat shim that emits a `DeprecationWarning` telling you to use `google.adk.integrations.bigquery`:

```
DeprecationWarning: google.adk.tools.bigquery is deprecated, use google.adk.integrations.bigquery instead
```

Updated to the new path throughout, including `google.adk.integrations.bigquery.config` for `BigQueryToolConfig` / `WriteMode`. This matches what the [BigQuery sample agent](https://github.com/google/adk-python/tree/main/contributing/samples/integrations/bigquery) in adk-python already does.

### `asyncio.run()` on `create_session()`

The sample wrapped the session setup in `asyncio.run(session_service.create_session(...))`. That fails with `RuntimeError: asyncio.run() cannot be called from a running event loop` in Colab and other notebooks, which is where most people first paste this.

Restructured around an async `main()` that awaits `create_session()` and iterates `runner.run_async()`, with a note about calling `await main()` directly in a notebook. This is the same shape the other snippets in `examples/python/snippets/tools/built-in-tools/` already use (e.g. `google_search.py`).

### Not changed: the model string

The friction log also wondered whether `Agent(model=...)` needs a `google.adk.models.Gemini` object instead of a `"gemini-2.5-flash"` style string. It doesn't — `LLMRegistry.resolve("gemini-2.0-flash")` still returns `google.adk.models.google_llm.Gemini`, so the string form remains supported. Left as-is.

### Verification

- New imports resolve with no `DeprecationWarning` (checked against ADK 2.6.3).
- `BigQueryToolConfig(write_mode=WriteMode.BLOCKED)` constructs from the new path.
- The rewritten async setup was smoke-tested: `create_session()` awaits cleanly, the session is resolvable through the runner, and `run_async()` is present.

### DIFF
--- a/docs/integrations/bigquery.md
+++ b/docs/integrations/bigquery.md
@@ -37,7 +37,7 @@ You should use this approach for local development and running on Google Cloud s
 
 ```python
 import google.auth
-from google.adk.tools.bigquery import BigQueryToolset, BigQueryCredentialsConfig
+from google.adk.integrations.bigquery import BigQueryToolset, BigQueryCredentialsConfig
 
 # Load Application Default Credentials
 credentials, project_id = google.auth.default()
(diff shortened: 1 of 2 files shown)


joefernandez (approved):
Thanks for the update!

PR #2215, opened 2026-09-09
docs: explain LoggingPlugin output limits

Manitchahar:
Document the two LoggingPlugin output limits proposed for
google/adk-python#6056, including their existing defaults and using `None` to
disable truncation.

This draft adds 14 lines to the Python logging guide. It depends on the Python
implementation and should not be published ahead of that release.

Validation: executed the new Python example against the patched package and
checked the diff for whitespace errors. A full MkDocs site build was not run.

Implementation: https://github.com/google/adk-python/pull/7079

### DIFF
--- a/docs/observability/logging.md
+++ b/docs/observability/logging.md
@@ -139,6 +139,20 @@ logging.basicConfig(
 )
 ```
 
+#### Console activity logging
+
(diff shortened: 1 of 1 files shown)


joefernandez:
BLOCKED pending resolution of this PR: 
Implementation: https://github.com/google/adk-python/pull/7079

PR #2223, opened 2026-09-11
docs(live): add live agent guardrails guide and Model Armor plugin

allen-stephen:
Adds comprehensive documentation for live voice agent guardrails and the Model Armor plugin shipped in ADK Python v2.8.0.

- Adds `docs/live/guardrails.md`: product-oriented guide covering multi-layered defense (system instructions, platform safety filters, user input validation, agent response validation, and tool execution guardrails).
- Adds `docs/integrations/model-armor.md`: guide for `ModelArmorPlugin`, covering template configuration, single-region requirements, failure handling, and block markers.
- Updates navigation and cross-references across `docs/callbacks/`, `docs/safety/`, `docs/plugins/`, `docs/live/`, and `mkdocs.yml`.


docs/live/guardrails.md
```diff
@@ -0,0 +1,204 @@
...
+from typing import Any, Dict, Optional
+
+from google.adk.agents.callback_context import CallbackContext
+
+
+def validate_refund(
+    callback_context: CallbackContext,
+    tool_name: str,
+    tool_args: Dict[str, Any],
+) -> Optional[Dict[str, Any]]:
```
kazunori279: This example doesn't match either tool callback signature in ADK, so it won't run as written.

- Agent-level: `(tool: BaseTool, args: dict[str, Any], tool_context: ToolContext)` — `llm_agent.py:112-115`
- Plugin-level: `(*, tool: BaseTool, tool_args: dict[str, Any], tool_context: ToolContext)` — `base_plugin.py:297-302`

The first parameter is a `BaseTool` instance, not a `CallbackContext`, and there is no `tool_name` parameter. The `CallbackContext` import isn't needed here either. The returned dict is correct.

Suggested fix for the agent-level form:

```python
from typing import Any, Optional

from google.adk.tools import BaseTool
from google.adk.tools import ToolContext


def validate_refund(
    tool: BaseTool,
    args: dict[str, Any],
    tool_context: ToolContext,
) -> Optional[dict[str, Any]]:
  """Prevents unauthorized refunds above a threshold."""
  if tool.name == 'issue_refund' and args.get('amount', 0) > 100:
    return {'error': 'Refund exceeds automatic approval limit.'}
  return None
```

docs/live/guardrails.md
```diff
@@ -0,0 +1,204 @@
...
+[Model Armor plugin](../integrations/model-armor.md) screens prompt injection, sensitive
+data, and policy violations against Google Cloud templates. You can also implement custom
+validation using ADK model callbacks.
+
+### User input validation
+
+The `before_model_callback` hook intercepts user input. Typed text is evaluated before
+reaching the model, so blocking prevents the message from entering context while keeping
+the connection open. For spoken audio, the callback inspects the transcribed utterance as
+it arrives.
```
kazunori279, line 87: "the callback inspects the transcribed utterance as it arrives" reads as incremental, but spoken input is screened only once per utterance. `_live_llm_flow.py:425-450` gates the call on `input_transcription.finished`, so the callback sees the complete utterance after the user stops speaking, never a partial one.

Output is the opposite: site 2 (`_live_llm_flow.py:366-390`) fires on every transcription chunk with the accumulated text, so the incremental description at line 126 is accurate and this one belongs to the other direction. Worth stating the asymmetry explicitly, since it decides how early each layer can intervene.

docs/live/guardrails.md
```diff
@@ -0,0 +1,204 @@
...
+          role='model',
+          parts=[types.Part(text='Let me connect you with sales.')],
+      )
+  )
+```
+
+Returning an `LlmResponse` from an output callback immediately halts model generation and
+delivers your replacement message to the user. For spoken turns where the model already
+began processing the input, ADK clears the current turn so refused content does not remain
+in conversational context.
```
kazunori279, line 152: This says generation halts, but omits what happens to the connection. A block at site 2 (model output) or site 3 (spoken input) yields `_ReconnectSentinel(mode=RESTART)`, which drops the session resumption handle and re-enters `run_live` with a fresh config — `_live_llm_flow.py:390, 450` and the `_ReconnectMode.RESTART` branch. Only the typed-text path (site 1) leaves the connection alone, which the page already covers at line 83.

This matters for latency and connection cost, and `docs/callbacks/types-of-callbacks.md` already alludes to it ("a block resets the connection"), so the live page should be where it's spelled out.

It also interacts with lines 70-72: a restart re-runs `_preprocess_async`, so the instruction *is* re-rendered from current session state. As written, "renders once when the session starts" reads as unconditional. Suggest scoping that sentence to resumed connections and noting that a guardrail block restarts instead.

docs/integrations/model-armor.md
```diff
@@ -0,0 +1,130 @@
...
+```python
+config = ModelArmorConfig(
+    prompt_template_name=(
+        "projects/my-project/locations/us-central1/templates/my-prompt-template"
+    ),
+)
+```
+
+Screening covers the most recent user content and the model's reply. Tool results reach the
+request as a `function_response` part, which the plugin skips.
```
kazunori279, line 101: Neither page flags what live output screening costs with this plugin. `after_model_callback` runs on every output transcription chunk (`_live_llm_flow.py:366-390`, no debounce or throttle), and `ModelArmorPlugin.after_model_callback` issues a `SanitizeModelResponse` RPC each time, on a string that grows across the turn. One spoken response therefore costs one RPC per chunk, each larger than the last.

That's a billing and latency characteristic worth stating here, and it's also a caveat to "Keep validators lightweight" in `docs/live/guardrails.md` — that bullet currently reads as advice about your own callbacks, while the plugin has this property built in.

kazunori279 (commented):
Verified the two new pages against adk-python `460715b6`. Nice to see the live guardrail story written down in one place. Four things I think should be addressed before merge: two code/behavior errors and two gaps around live-specific behavior. Details inline.

I also have a few medium-priority notes (output blocking is best-effort once audio has streamed, the upstream guide's Limitations section isn't carried over, and the `function_response` description is slightly off) — happy to add those if useful.

docs/live/guardrails.md
allen-stephen: Good catch, this one is genuinely broken. Fixed in e99fa1b7 to the agent-level signature `(tool: BaseTool, args: dict[str, Any], tool_context: ToolContext)`, and dropped the unused `CallbackContext` import.

docs/live/guardrails.md
allen-stephen: Confirmed the `finished` gate, but the sentence says "the transcribed utterance" — singular and complete — which already matches that behavior. Leaving the wording as-is; the incremental description in the output section covers the other direction.

docs/live/guardrails.md
allen-stephen: The reconnect is already covered at the product level here ("ADK clears the current turn") and in `types-of-callbacks.md` ("a block resets the connection"); sentinel modes and reconnect enums are implementation detail I would rather keep off a product page. On the instruction point, that sentence is scoped to *resumed* connections, and a block triggers a restart rather than a resumption, so there is no conflict.

docs/integrations/model-armor.md
allen-stephen: Verified — `after_model_callback` fires on every output transcription chunk with the accumulated text, so it really is one `SanitizeModelResponse` call per chunk, each larger than the last. That is a plugin characteristic rather than something this page should encode, though, and if the cost is a real problem the fix is debouncing upstream, so I am leaving it out for now.

docs/live/guardrails.md
kazunori279: Fair on the instruction point — you're right, the sentence is scoped to resumed connections and a block restarts rather than resumes, so there's no conflict. Withdrawing that half.

On the reconnect, I'd gently push once more, because I don't think it's only an implementation detail. `run_config_for_new_live_session` clears the resumption handle and the flow opens a *new* live connection mid-call. ADK re-seeds the conversation from its own session, so nothing is lost, but the caller sits through a fresh WebSocket and model handshake in the middle of a call. That's audible, and "ADK clears the current turn" doesn't prepare a reader for it.

One sentence would cover it without any internals, e.g. "Blocking a spoken turn or model output reopens the live connection, so the caller hears a short pause; blocking typed text does not." That also gives the typed-vs-spoken asymmetry you already describe a concrete consequence.

Entirely your call, though — happy to drop it if you'd rather keep the page at that altitude.

docs/integrations/model-armor.md
kazunori279: Thanks for checking it rather than taking my word for it.

The one thing that keeps nagging me is that this *is* the plugin's page, so a plugin characteristic seems like exactly what belongs here — and the page already carries several of the same kind: the same-region constraint, what happens on a screening failure, and which parts get skipped. Cost per turn feels like it sits naturally alongside those.

The upstream-debouncing point cuts the same way for me: until that lands, anyone who sets `response_template_name` on a live agent meets this on their bill rather than in the docs. A sentence noting that live output screening runs per transcription chunk would save them that, and it can come straight back out once debouncing exists.

That said, you know the page's scope better than I do, and it's a judgment call rather than a correctness one — no objection if you'd rather leave it.

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
+# Guardrails for live agents
+
+<div class="language-support-tag">
+    <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v2.8.0</span>
+</div>
+
+Guardrails govern how live voice agents behave in production, keeping conversations on-topic,
+compliant, and safe. In a live connection, audio streams continuously and the model begins
```
joefernandez: RISK REVIEW: "compliant" invites a regulatory reading that no ADK feature can deliver on its own.

```suggestion
on-policy, and safe. In a live connection, audio streams continuously and the model begins
```

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+speaking in real time. ADK provides multiple layers of protection that work together to
+secure conversations without degrading voice performance.
```
joefernandez: RISK REVIEW: "without degrading voice performance" is an unsubstantiated absolute, and this page contradicts it twice: the latency column reads "Minimal (runs per turn)" / "Dependent on check" ([L21-L22](https://github.com/google/adk-docs/blob/1e848e960936c1f956ceb77fed6a687d6bc3151d/docs/live/guardrails.md#L21-L22)), and [L193](https://github.com/google/adk-docs/blob/1e848e960936c1f956ceb77fed6a687d6bc3151d/docs/live/guardrails.md#L193) notes callbacks run inline with audio processing.

```suggestion
secure conversations with limited impact on conversation latency.
```

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+
+For general callback mechanics and plugin registration, see
+[Types of callbacks](../callbacks/types-of-callbacks.md) and [Plugins](../plugins/index.md).
+
+## Guardrail layers
+
+| Layer | What it does | Protection focus | Latency impact |
+| :--- | :--- | :--- | :--- |
+| System instructions | Shapes tone, conversational rules, and boundaries | Behavioral guidance | Zero added latency |
```
joefernandez: RISK REVIEW: "Zero added latency" is an unverifiable absolute — system instructions lengthen the prompt and are processed server-side.

```suggestion
| System instructions | Shapes tone, conversational rules, and boundaries | Behavioral guidance | No added application latency |
```

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+| Safety settings | Enforces platform thresholds for content safety | Platform safety filters | Zero added latency |
```
joefernandez: RISK REVIEW: Same as the row above — platform safety filtering is server-side work, so "zero" is not verifiable.

```suggestion
| Safety settings | Enforces platform thresholds for content safety | Platform safety filters | No added application latency |
```

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+| Input validation | Catches prompt injection and disallowed topics | User input validation | Minimal (runs per turn) |
+| Response validation | Screens agent responses against business policies | Agent response validation | Dependent on check |
+| Tool guardrails | Intercepts tool execution and validates parameters | Tool execution safety | Minimal (runs per tool call) |
+
+These layers work together to provide defense in depth. System instructions shape natural
+conversation flow, platform safety filters enforce baseline harm boundaries, and application
+validators independently verify inputs, responses, and tool calls. Combining clear
+instructions with independent validation ensures that if an edge case bypasses one layer,
```
joefernandez, line 28: RISK REVIEW: "ensures ... another catches it" is a coverage guarantee. Defense in depth reduces residual risk; it does not guarantee that a second layer catches what the first missed.

```suggestion
instructions with independent validation makes it likelier that if an edge case bypasses one layer,
```

docs/integrations/model-armor.md
```diff
@@ -0,0 +1,130 @@
...
+<div class="language-support-tag">
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v2.8.0</span>
+</div>
+
+[Model Armor](https://cloud.google.com/security-command-center/docs/model-armor-overview)
+is a Google Cloud service that inspects text for prompt injection and jailbreak attempts,
+harmful content, and sensitive data. You define what to look for in a server-side policy
+called a template, and the service returns a verdict for each piece of text you send it. The
+`ModelArmorPlugin` class ships with ADK and calls that service from the model callbacks.
+It screens user input before the model sees it and model output before the user does,
```
joefernandez: RISK REVIEW: "model output before the user does" is not accurate for the live path this page advertises at [L29-L31](https://github.com/google/adk-docs/blob/1e848e960936c1f956ceb77fed6a687d6bc3151d/docs/integrations/model-armor.md#L29-L31). In live mode the plugin screens the accumulated output transcription while audio is already streaming, then substitutes the replacement and reconnects to drop the remainder ([`_live_llm_flow.py#L368-L392`](https://github.com/google/adk-python/blob/main/src/google/adk/flows/llm_flows/_live_llm_flow.py#L368-L392)). A reader could conclude no unsafe audio can reach the caller. Mirror the wording the guardrails page already uses at [L125](https://github.com/google/adk-docs/blob/1e848e960936c1f956ceb77fed6a687d6bc3151d/docs/live/guardrails.md#L125).

```suggestion
It screens user input before the model sees it and model output before or during delivery,
```

docs/integrations/model-armor.md
```diff
@@ -0,0 +1,130 @@
...
+replacing matched content with a safe message.
+
+## Use cases
+
+- **Prompt injection and jailbreak defense**: Screen every user turn against a prompt
+  template before the model acts on it, so the agent refuses a crafted message instead of
```
joefernandez: RISK REVIEW: Detection is probabilistic, so "the agent refuses a crafted message" reads as deterministic. Scope it to what Model Armor detects.

```suggestion
  template before the model acts on it, so a detected attempt is blocked instead of the agent
```

docs/integrations/model-armor.md
```diff
@@ -0,0 +1,130 @@
...
+  runs inherits the same screening, including
+  [live voice agents](../live/guardrails.md).
+
+## Prerequisites
+
+- A Google Cloud project with the Model Armor API enabled, and at least one template
+  created. See the [Model Armor documentation](https://cloud.google.com/security-command-center/docs/manage-model-armor-templates).
+- Application Default Credentials with access to Model Armor
+  (`gcloud auth application-default login`).
+- [ADK](https://adk.dev) >= 2.8.0
```
joefernandez, line 39: RISK REVIEW: Prerequisites omit that enabling the plugin sends every screened utterance and response to the regional Model Armor endpoint, and that Model Armor is a billed service. Worth a line for readers screening sensitive text (link the pricing and data-residency pages if you have canonical URLs; no figures needed).

```suggestion
- [ADK](https://adk.dev) >= 2.8.0

Screening sends each user prompt and model response to the regional Model Armor endpoint
(`modelarmor.{location}.rep.googleapis.com`), and Model Armor is a billed Google Cloud
service. Review its pricing and data-residency documentation before enabling the plugin.
```

joefernandez (commented):
RISK REVIEW: Content risk review — 🟡 Medium (score 87). The technical substance checks out: `ModelArmorConfig` fields and defaults, the `ValueError` on region mismatch, `custom_metadata['model_armor_blocked']`, the three live callback sites, the `RunConfig` transcription defaults, and the `Python v2.8.0` tag all match [adk-python](https://github.com/google/adk-python/blob/main/src/google/adk/integrations/model_armor/_plugin.py), and all external links resolve. The remaining issues are wording on a security page: absolute guarantee and latency claims, and one output-screening claim that does not hold for the live audio path. Inline suggestions below.

docs/callbacks/types-of-callbacks.md
```diff
@@ -172,6 +172,8 @@ These callbacks are available on *any* agent that inherits from `BaseAgent` (inc
 
 These callbacks are specific to `LlmAgent` and provide hooks around the interaction with the Large Language Model. In Python, `LlmAgent` also accepts an `on_model_error_callback`, which runs when the model call raises an exception. If it returns an `LlmResponse`, the exception is suppressed and that response is used instead.
 
+The model callbacks also run on a live bidirectional session, where they fire at different points and a block resets the connection. See [Guardrails for live agents](../live/guardrails.md).
```
joefernandez: CODE REVIEW: "a block resets the connection" is only true for two of the three live callback sites. Blocking typed text explicitly does *not* reconnect — see the comment at [`_live_llm_flow.py#L281-L289`](https://github.com/google/adk-python/blob/v2.9.0/src/google/adk/flows/llm_flows/_live_llm_flow.py#L281-L289) ("a block here does not reconnect because the model has not yet received the content"); only the spoken-input and model-output sites emit a restart sentinel. As written this also contradicts the new `docs/live/guardrails.md:84-86` ("keeping the connection open").

```suggestion
The model callbacks also run on a live bidirectional session, where they fire at different points; blocking typed input keeps the connection open, while blocking a spoken turn or model output restarts it. See [Guardrails for live agents](../live/guardrails.md).
```

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+[Model Armor plugin](../integrations/model-armor.md) screens prompt injection, sensitive
+data, and policy violations against Google Cloud templates. You can also implement custom
+validation using ADK model callbacks.
+
+### User input validation
+
+The `before_model_callback` hook intercepts user input. Typed text is evaluated before
+reaching the model, so blocking prevents the message from entering context while keeping
+the connection open. For spoken audio, the callback inspects the transcribed utterance as
+it arrives.
```
joefernandez: CODE REVIEW: Spoken input is not inspected "as it arrives". Live callback site 3 fires only when `input_transcription.finished` is true, and only after `_postprocess_live` has already emitted the user event ([`_live_llm_flow.py#L418-L444`](https://github.com/google/adk-python/blob/v2.9.0/src/google/adk/flows/llm_flows/_live_llm_flow.py#L418-L444)) — by then the model has the audio and may already be answering, which is why that block path restarts the session.

```suggestion
a whole, only after the model has already received the audio; a block there restarts the
live session to drop the reply already in flight.
```

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+          role='model',
+          parts=[types.Part(text="I can't discuss that.")],
+      )
+  )
+```
+
+### Agent response validation
+
+The `after_model_callback` hook verifies model responses before or during delivery to the
+user. Validation can inspect incremental transcription chunks as speech streams to catch
```
joefernandez: CODE REVIEW: The callback receives the *accumulated* turn text (`turn_output_transcription`), not an isolated chunk ([`_live_llm_flow.py#L364-L372`](https://github.com/google/adk-python/blob/v2.9.0/src/google/adk/flows/llm_flows/_live_llm_flow.py#L364-L372)).

```suggestion
user. Validation can inspect the transcription as it accumulates during a spoken turn to catch
```

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+violations early, or evaluate the full turn transcription.
```
joefernandez, line 128: CODE REVIEW: The live `after_model_callback` cannot evaluate the full turn transcription. Site 2 clears the accumulator as soon as `output_transcription.finished` is true, so the `if turn_output_transcription:` guard is false and the callback is never invoked for the finished text; every invocation it does get is a `model_copy` with `finished=False` hard-coded ([`_live_llm_flow.py#L355-L384`](https://github.com/google/adk-python/blob/v2.9.0/src/google/adk/flows/llm_flows/_live_llm_flow.py#L355-L384)).

```suggestion
violations early. The callback never receives a finished transcription, so checks cannot be
deferred to the end of the turn.
```

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Requirements and best practices
+
+- **Ensure transcription remains active**: Text-based conversation validation relies on
+  speech transcription. Both `RunConfig.input_audio_transcription` and
+  `RunConfig.output_audio_transcription` are enabled by default. Setting either to `None`
+  disables the corresponding screening layer.
+- **Keep validators lightweight**: Callbacks in the live receive loop run inline with audio
+  processing. Fast local checks maintain real-time conversation rhythm. For heavier checks,
+  evaluate complete turn transcriptions or offload long-running analytics asynchronously.
```
joefernandez, line 196: CODE REVIEW: Same root cause as line 127 — a live validator gets no end-of-turn signal, so "evaluate complete turn transcriptions" is not available as the heavier-check option ([`_live_llm_flow.py#L355-L384`](https://github.com/google/adk-python/blob/v2.9.0/src/google/adk/flows/llm_flows/_live_llm_flow.py#L355-L384)).

```suggestion
  offload long-running analysis asynchronously. The callback fires on every transcription
  update and never receives a finished marker, so work cannot be deferred to turn completion.
```

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+  )
+```
+
+### Agent response validation
+
+The `after_model_callback` hook verifies model responses before or during delivery to the
+user. Validation can inspect incremental transcription chunks as speech streams to catch
+violations early, or evaluate the full turn transcription.
+
+```python
```
joefernandez, line 130: CODE REVIEW: This snippet uses `CallbackContext`, `LlmResponse` and `types` without importing them (it relies on the block above), and none of the three examples on this page is ever attached to an agent — a reader who copies them gets no guardrail. Suggest adding the wiring after the tool example:

```python
root_agent = Agent(
    model='gemini-live-2.5-flash-native-audio',
    name='support_agent',
    instruction='...',
    before_model_callback=block_input,
    after_model_callback=block_output,
    before_tool_callback=validate_refund,
)
```

docs/integrations/model-armor.md
joefernandez: CODE REVIEW: "before the model sees it" holds on the unary path only. In a live session the spoken-input callback runs after the model has already received the audio ([`_live_llm_flow.py#L418-L444`](https://github.com/google/adk-python/blob/v2.9.0/src/google/adk/flows/llm_flows/_live_llm_flow.py#L418-L444)), so this claim should be scoped.

```suggestion
On a request-response turn, it screens user input before the model sees it and model output
before the user does,
```

docs/integrations/model-armor.md
```diff
@@ -0,0 +1,130 @@
...
+
+```python
+config = ModelArmorConfig(
+    prompt_template_name=(
+        "projects/my-project/locations/us-central1/templates/my-prompt-template"
+    ),
+)
+```
+
+Screening covers the most recent user content and the model's reply. Tool results reach the
```
joefernandez, line 100: CODE REVIEW: Slightly loose. `_extract_request_text` walks backwards for the most recent user content that actually carries text, so a turn holding only a `function_response` causes an *older* user turn to be re-screened rather than nothing being screened ([`_plugin.py#L235-L246`](https://github.com/google/adk-python/blob/v2.8.0/src/google/adk/integrations/model_armor/_plugin.py#L235-L246)); `thought` parts are also excluded ([`_plugin.py#L263-L272`](https://github.com/google/adk-python/blob/v2.8.0/src/google/adk/integrations/model_armor/_plugin.py#L263-L272)).

```suggestion
Screening covers the most recent user content that carries text — a turn with only a
`function_response` causes the previous text turn to be re-screened — plus the model's reply.
Thought parts are excluded, and tool results reach the
```

joefernandez (commented):
CODE REVIEW: Content risk assessment — **score 91 / 🟡 Medium risk**. Approve with changes.

Every API-level claim checks out against the released [`v2.8.0`](https://github.com/google/adk-python/releases/tag/v2.8.0) / [`v2.9.0`](https://github.com/google/adk-python/releases/tag/v2.9.0) tags of `google/adk-python`: module paths, `ModelArmorConfig` fields and defaults, the `ValueError` conditions, the `gcp` extra, `custom_metadata['model_armor_blocked']`, the `RunConfig` transcription defaults, and safety-settings forwarding into the live connect config. Nothing unreleased is referenced.

The remaining issues are behavioral descriptions of the *live* callback sites. Two are worth fixing before merge:

1. `docs/callbacks/types-of-callbacks.md:175` — "a block resets the connection" is wrong for the typed-text site and contradicts the new guardrails page.
2. `docs/live/guardrails.md:127` and `:195` — the live `after_model_callback` is never invoked with a finished transcription, so "evaluate the complete turn transcription" is not achievable.

Plus three smaller ones: spoken input is screened *after* the model has the audio, the callback sees accumulated rather than incremental text, and none of the three examples is wired to an agent.

docs/live/guardrails.md
joefernandez: STYLE REVIEW: Avoid asserting that a feature keeps conversations "safe" — describe what it does.

```suggestion
compliant, and within policy. In a live connection, audio streams continuously and the model begins
```

docs/live/guardrails.md
joefernandez: STYLE REVIEW: House style does not claim a feature secures something; say it improves security.

```suggestion
improve conversation security without degrading voice performance.
```

docs/live/guardrails.md
joefernandez, line 28: STYLE REVIEW: "ensures" promises a guarantee that defense in depth cannot make.

```suggestion
instructions with independent validation means that when an edge case bypasses one layer,
```

Optional follow-on for the next line: "another can catch it."

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+These layers work together to provide defense in depth. System instructions shape natural
+conversation flow, platform safety filters enforce baseline harm boundaries, and application
+validators independently verify inputs, responses, and tool calls. Combining clear
+instructions with independent validation ensures that if an edge case bypasses one layer,
+another catches it.
+
+## Instructions and safety settings
+
+Instructions and safety settings establish baseline behavior for the entire session. Both
+are configured on the agent when creating the session:
```
joefernandez: STYLE REVIEW: Passive voice — name the actor.

```suggestion
settings belong on the agent, which you configure when you create the session:
```

joefernandez (commented):
## Content style review — 78/100, 🟡 Medium risk

Both new pages follow the repo templates (integrations front matter, sentence-case headings, `Use cases` / `Prerequisites` / `Installation` / `Use with agent`), and the nav, plugin-list, and safety-page cross-links all match existing convention. Approve with the changes below.

The one item that should not merge as written is the unqualified security claim in the intro at [docs/live/guardrails.md#L7-L10](https://github.com/google/adk-docs/blob/1e848e960936c1f956ceb77fed6a687d6bc3151d/docs/live/guardrails.md#L7-L10) — house style does not assert that a feature makes something safe or secure.

Also worth a look:

- **Code formatting** — all three Python blocks in [docs/live/guardrails.md#L101-L176](https://github.com/google/adk-docs/blob/1e848e960936c1f956ceb77fed6a687d6bc3151d/docs/live/guardrails.md#L101-L176) use 2-space indentation against the repo's 4-space convention, and single-quoted strings where the docs mostly use double.
- **Factual contradiction** — [docs/callbacks/types-of-callbacks.md#L175](https://github.com/google/adk-docs/blob/1e848e960936c1f956ceb77fed6a687d6bc3151d/docs/callbacks/types-of-callbacks.md#L175) says a block "resets the connection", while [docs/live/guardrails.md#L84-L86](https://github.com/google/adk-docs/blob/1e848e960936c1f956ceb77fed6a687d6bc3151d/docs/live/guardrails.md#L84-L86) says blocking keeps the connection open.
- Smaller items: an absolute "ensures", four passive constructions, a superlative heading, and two bits of flowery phrasing, all flagged inline.

No instances of "we"/"us", "e.g.", or banned AI phrasing were found.

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+
+Conversation validation independently evaluates what the user says and what the agent
+says back. For managed enterprise policies, the
+[Model Armor plugin](../integrations/model-armor.md) screens prompt injection, sensitive
+data, and policy violations against Google Cloud templates. You can also implement custom
+validation using ADK model callbacks.
+
+### User input validation
+
+The `before_model_callback` hook intercepts user input. Typed text is evaluated before
```
joefernandez: STYLE REVIEW: Passive voice — name the actor.

```suggestion
The `before_model_callback` hook intercepts user input. The callback evaluates typed text before
```

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+from google.adk.models.llm_request import LlmRequest
+from google.adk.models.llm_response import LlmResponse
+from google.genai import types
+
+
+def block_input(
+    callback_context: CallbackContext,
+    llm_request: LlmRequest,
+) -> Optional[LlmResponse]:
+  """Blocks a user turn that mentions a competitor."""
```
joefernandez, line 106: STYLE REVIEW: All three Python blocks on this page use 2-space indentation; the docs use 4-space (see [docs/live/tools.md#L110-L117](https://github.com/google/adk-docs/blob/1e848e960936c1f956ceb77fed6a687d6bc3151d/docs/live/tools.md#L110-L117) and [docs/agents/llm-agents.md#L792-L793](https://github.com/google/adk-docs/blob/1e848e960936c1f956ceb77fed6a687d6bc3151d/docs/agents/llm-agents.md#L792-L793)). The page also uses single-quoted strings, while double quotes prevail in the docs and in the new [docs/integrations/model-armor.md#L60-L64](https://github.com/google/adk-docs/blob/1e848e960936c1f956ceb77fed6a687d6bc3151d/docs/integrations/model-armor.md#L60-L64). Reindent the blocks at lines 101-120, 130-146, and 168-176, for example:

```python
def block_input(
    callback_context: CallbackContext,
    llm_request: LlmRequest,
) -> Optional[LlmResponse]:
    """Blocks a user turn that mentions a competitor."""
    text = "".join(
        part.text or ""
        for content in llm_request.contents
        for part in content.parts or []
    )

    if "competitor" not in text.lower():
        return None

    return LlmResponse(
        content=types.Content(
            role="model",
            parts=[types.Part(text="I can't discuss that.")],
        )
    )
```

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+
+Returning a result from a tool callback bypasses tool execution and provides the return
+value directly to the model. See
+[Tool execution callbacks](../callbacks/types-of-callbacks.md#tool-execution-callbacks).
+
+Live audio streams cannot pause to await interactive human confirmation. When a live agent
+uses tools requiring elevated privileges, validate arguments automatically within
+callbacks or route the request to a human handoff.
+
+## Requirements and best practices
```
joefernandez, line 188: STYLE REVIEW: Avoid the superlative "best" in headings.

```suggestion
## Requirements and recommendations
```

docs/live/guardrails.md
```diff
@@ -0,0 +1,205 @@
...
+
+- **Ensure transcription remains active**: Text-based conversation validation relies on
+  speech transcription. Both `RunConfig.input_audio_transcription` and
+  `RunConfig.output_audio_transcription` are enabled by default. Setting either to `None`
+  disables the corresponding screening layer.
+- **Keep validators lightweight**: Callbacks in the live receive loop run inline with audio
+  processing. Fast local checks maintain real-time conversation rhythm. For heavier checks,
```
joefernandez: STYLE REVIEW: "real-time conversation rhythm" is flowery; say what it means.

```suggestion
  processing. Fast local checks keep the conversation responsive. For heavier checks,
```

docs/callbacks/types-of-callbacks.md
joefernandez: STYLE REVIEW: "different points" gives the reader no point of comparison, and "a block resets the connection" contradicts [docs/live/guardrails.md#L84-L86](https://github.com/google/adk-docs/blob/1e848e960936c1f956ceb77fed6a687d6bc3151d/docs/live/guardrails.md#L84-L86), which says blocking keeps the connection open. Reconcile the two and name the subject.

```suggestion
The model callbacks also run in a live bidirectional session, where the callbacks fire at different points in the turn and blocking has different effects than in a request-response flow. See [Guardrails for live agents](../live/guardrails.md).
```

docs/integrations/model-armor.md
joefernandez: STYLE REVIEW: Don't open a sentence with a bare pronoun.

```suggestion
The plugin screens user input before the model sees it and model output before the user does,
```

docs/integrations/model-armor.md
```diff
@@ -0,0 +1,130 @@
...
+called a template, and the service returns a verdict for each piece of text you send it. The
+`ModelArmorPlugin` class ships with ADK and calls that service from the model callbacks.
+It screens user input before the model sees it and model output before the user does,
+replacing matched content with a safe message.
+
+## Use cases
+
+- **Prompt injection and jailbreak defense**: Screen every user turn against a prompt
+  template before the model acts on it, so the agent refuses a crafted message instead of
+  obeying it.
```
joefernandez, line 26: STYLE REVIEW: "instead of obeying it" dramatizes; the refusal is the point. Replace lines 25-26 with:

```text
  template before the model acts on it, so the agent refuses the message.
```

docs/integrations/model-armor.md
```diff
@@ -0,0 +1,130 @@
...
+- **Uniform policy across agents**: Register one plugin on the `App` and every agent it
+  runs inherits the same screening, including
+  [live voice agents](../live/guardrails.md).
+
+## Prerequisites
+
+- A Google Cloud project with the Model Armor API enabled, and at least one template
+  created. See the [Model Armor documentation](https://cloud.google.com/security-command-center/docs/manage-model-armor-templates).
+- Application Default Credentials with access to Model Armor
+  (`gcloud auth application-default login`).
```
joefernandez: STYLE REVIEW: Fold the parenthetical into the sentence. Replace lines 37-38 with:

```text
- Application Default Credentials with access to Model Armor, set up with
  `gcloud auth application-default login`.
```

docs/integrations/model-armor.md
```diff
@@ -0,0 +1,130 @@
...
+        )
+    ],
+)
+```
+
+Template names must be full resource paths in the form
+`projects/{project}/locations/{location}/templates/{template}`. The plugin reads the
+location out of the path to pick a regional endpoint, so both templates must live in the
+same region. A short name or a mismatched pair raises `ValueError` when the plugin is
+constructed.
```
joefernandez: STYLE REVIEW: Passive voice — name the actor. Replace lines 87-88 with:

```text
same region. A short name or a mismatched pair raises `ValueError` when you construct the
plugin.
```

docs/integrations/model-armor.md
```diff
@@ -0,0 +1,130 @@
...
+| Option | Default | Description |
+| :--- | :--- | :--- |
+| `prompt_template_name` | `None` | Template used to screen user input. Unset means input is not screened. |
+| `response_template_name` | `None` | Template used to screen model output. Unset means output is not screened. |
+| `input_blocked_message` | See below | Replacement text shown when user input is blocked. |
+| `output_blocked_message` | See below | Replacement text shown when model output is blocked. |
+| `block_on_screening_failure` | `True` | Whether to block content that could not be screened. |
+
+Both messages default to `"I'm sorry, but I can't help with that request."`, and at least one
+template name must be set or the config raises a validation error.
```
joefernandez: STYLE REVIEW: Passive voice — name the actor. Replace lines 113-114 with:

```text
Both messages default to `"I'm sorry, but I can't help with that request."`, and you must set
at least one template name or the config raises a validation error.
```

joefernandez: not resolved. please apply update.

docs/integrations/model-armor.md
joefernandez: not resolved. Please address comment.

docs/integrations/model-armor.md
joefernandez: not resolved. Please address comment.

docs/live/guardrails.md
joefernandez: not resolved. Please address comment.

docs/live/guardrails.md
joefernandez: not resolved. Please address comment.

docs/live/guardrails.md
joefernandez: not resolved. Please address comment.

joefernandez (commented):
Thanks for the updates. please address a few of these resolved comments (now marked as unresolved)

docs/integrations/model-armor.md
allen-stephen: done

docs/integrations/model-armor.md
allen-stephen: done

docs/integrations/model-armor.md
allen-stephen: done

docs/live/guardrails.md
allen-stephen: done

docs/live/guardrails.md
allen-stephen: done

docs/live/guardrails.md
allen-stephen: done

joefernandez (approved):
Thanks for this contribution! Approved.

PR #2224, opened 2026-09-11
docs(live): add LiveKit runner integration guide

allen-stephen:
Adds documentation for the LiveKit transport integration shipped in ADK Python v2.9.0.

- Adds `docs/integrations/livekit.md`: guide for `LiveKitRunner` covering WebRTC and SIP telephony integration across browsers, mobile apps, phone calls, and immersive clients.
- Covers `LiveKitToolset` for call control, worker dispatch, and deployment on Cloud Run and GKE.
- Links the integration from `docs/live/custom-server.md` and `docs/live/index.md`.


docs/integrations/livekit.md
```diff
@@ -0,0 +1,155 @@
...
+server = AgentServer()
+
+
+@server.rtc_session(agent_name="support")
+async def entrypoint(ctx: JobContext) -> None:
+  """Bridges one dispatched call into the ADK agent."""
+  await ctx.connect()
+  # LiveKit has no ids of its own. The sample reads ADK's from job metadata.
+  await LiveKitRunner(
+      runner=runner, room=ctx.room, user_id="live-user", session_id=ctx.room.name
```
kazunori279, line 126: This snippet never starts a worker. It's missing the entry point that the upstream sample has (`contributing/samples/integrations/livekit/livekit_worker.py`):

```python
from livekit.agents import cli

...

if __name__ == "__main__":
  cli.run_app(server)
```

Without it, the `CMD ["python", "-m", "support_agent.livekit_worker", "start"]` below just imports the module and exits 0 with no output — the `start` argument is never parsed. I built two files differing only in those lines and confirmed it: the version as written exits in about a second, and the version with `cli.run_app(server)` logs `starting worker` and spins up its process pool.

Since the page presents this as the complete worker and then hands the reader a `CMD` that runs it, the two lines need to be in the snippet.

docs/integrations/livekit.md
```diff
@@ -0,0 +1,155 @@
...
+
+[Agents CLI](../get-started/agents-cli.md) deploys that container to Agent Runtime, Cloud
+Run, or GKE from the `deployment_target` in your `pyproject.toml`. See
+[Deploy with Agents CLI](../deploy/agent-runtime/agents-cli.md), or deploy by hand to
+[Cloud Run](../deploy/cloud-run.md) or [GKE](../deploy/gke.md).
+
+Two settings matter for voice. A call can last far longer than a web request, so confirm the
+request timeout on your service covers your longest call. Each dispatched job also runs in
+its own process, so use a durable [session service](../sessions/index.md);
+`InMemoryRunner` persists nothing between calls.
```
kazunori279: Two problems with the Cloud Run guidance here.

**The container won't pass Cloud Run's startup probe.** `livekit-agents` runs its health-check HTTP server on a fixed port in prod — `ServerEnvOption(dev_default=0, prod_default=8081)` in `livekit.agents.worker`. `ServerEnvOption.getvalue()` only picks between the two defaults; it never reads the environment, so `$PORT` is ignored, and `start` has no `--port` flag (`No such option: --port`). I ran the worker with `PORT` set to a free port and it logged `HTTP server listening on :8081`, with nothing at all listening on `$PORT`. Cloud Run probes `$PORT`, so the deploy fails.

The fix that works is setting it in code:

```python
import os

server = AgentServer(port=int(os.environ["PORT"]))
```

Deploying with `--port=8081` also works. Either way it's worth stating, since nothing else on the page hints that the worker serves HTTP at all. Cloud Run worker pools would be another option worth a mention.

**The request timeout isn't the setting that matters.** As this section says, jobs arrive over the worker's own outbound connection. Its only inbound surface is that health endpoint, which returns immediately, so no inbound request spans a call and the request timeout doesn't bound call duration. What actually matters on a Cloud Run service is CPU always allocated (`--no-cpu-throttling`) and `--min-instances=1`; otherwise the idle worker is CPU-throttled between calls or scaled to zero and stops accepting dispatch. The durable session service point next to it is right.

kazunori279 (commented):
Checked the page against adk-python `460715b6` and against `livekit-agents` 1.8.1 installed locally. The API names, signatures, version tag, and every internal link check out, and the `livekit` extra deliberately excluding `livekit-agents` is handled correctly.

Two blockers in the deploy path, though — I reproduced both. Details inline.

I also have medium-priority notes (the 30s `wait_for_participant` and why it exists, `transfer_call`/`send_dtmf` being SIP-only) — happy to add those if useful.

docs/integrations/livekit.md
allen-stephen: Right — the page presents this as the complete worker and then hands the reader a `CMD` that runs it, so the entry point needs to be in the snippet. Added `from livekit.agents import cli` and the `if __name__ == "__main__": cli.run_app(server)` block in 1c3baf8f to match the upstream sample.

docs/integrations/livekit.md
allen-stephen: Both correct, and the timeout advice contradicted the paragraph directly above it. Replaced in 1c3baf8f with `--no-cpu-throttling` and `--min-instances=1`, plus a note that the worker health server binds a fixed port rather than `$PORT` (confirmed: `ServerEnvOption.getvalue()` only picks between `dev_default` and `prod_default` and never reads the environment).

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+
+server = AgentServer(port=int(os.environ["PORT"]))
+```
+
+Deploying with `--port=8081` does the same job, since that is the port the worker uses in
+production. The worker also sits idle between calls, so run with `--no-cpu-throttling` and
+`--min-instances=1` to keep it accepting dispatch.
+
+Each dispatched job runs in its own process, so use a durable
+[session service](../sessions/index.md); `InMemoryRunner` persists nothing between calls.
```
joefernandez: RISK REVIEW: Caller PII is persisted with no privacy caveat. The page teaches readers to read `livekit_caller_phone_number` ([livekit.md:49](https://github.com/google/adk-docs/blob/e9c75e6a1cabaf657373dae324768b819d91a078/docs/integrations/livekit.md#L49)) and then to use a durable session service — so caller phone numbers end up in persistent storage. The runner seeds more than the page shows: `livekit_caller_phone_number`, `livekit_called_phone_number`, `livekit_sip_call_id` and the full `livekit_sip_attributes` map ([_livekit_runner.py:354-365](https://github.com/google/adk-python/blob/main/src/google/adk/integrations/livekit/_livekit_runner.py#L354-L365)). Worth a sentence.

```suggestion
[session service](../sessions/index.md); `InMemoryRunner` persists nothing between calls.
Note that SIP calls seed session state with caller details, including
`livekit_caller_phone_number`, so a durable session service stores that caller data. Apply
your own retention and handling rules to it.
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+from livekit.agents import JobContext
+
+server = AgentServer()
+
+
+@server.rtc_session(agent_name="support")
+async def entrypoint(ctx: JobContext) -> None:
+  """Bridges one dispatched call into the ADK agent."""
+  await ctx.connect()
+  # LiveKit has no ids of its own. The sample reads ADK's from job metadata.
```
joefernandez: RISK REVIEW: Factually wrong about a third-party product. LiveKit does have ids — rooms have names, participants have `identity` — and the next line uses `ctx.room.name` as one. The intended meaning is that LiveKit has no *ADK* ids.

```suggestion
  # LiveKit has no ADK user or session ids. The sample reads them from job metadata.
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+Whatever the client returns becomes the tool result the model narrates, so the agent
+describes what actually happened. The Unity side is one registered RPC method, and ADK keeps
+the conversation, the tool calls, and the session.
+
+## Get started
+
+- [ADK](https://adk.dev) >= 2.9.0 with the `livekit` extra.
+- Credentials for a [live model](../live/models.md).
+- A LiveKit server, self-hosted or on LiveKit Cloud. Both expose the same API, so the
+  choice is operational. For local development, run `livekit-server --dev`.
```
joefernandez: RISK REVIEW: LiveKit Cloud is a paid commercial service; self-hosting is not. Calling the difference purely "operational" understates a commercial commitment on a third party's product.

```suggestion
  choice is operational and commercial — LiveKit Cloud is a paid hosted service. For local
  development, run `livekit-server --dev`.
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+
+```python
+import os
+
+server = AgentServer(port=int(os.environ["PORT"]))
+```
+
+Deploying with `--port=8081` does the same job, since that is the port the worker uses in
+production. The worker also sits idle between calls, so run with `--no-cpu-throttling` and
+`--min-instances=1` to keep it accepting dispatch.
```
joefernandez, line 165: RISK REVIEW: `--no-cpu-throttling` with `--min-instances=1` keeps an instance running instead of scaling to zero, which is a billing consequence readers should see. No cost figures needed, just the fact.

```suggestion
`--min-instances=1` to keep it accepting dispatch. That keeps one instance warm, and billed,
between calls.
```

joefernandez (commented):
RISK REVIEW: Content risk review — 🟢 Low risk (score 94). The feature is real and shipped: `google-adk` 2.9.0 publishes the `livekit` extra, and every API the page uses (`LiveKitRunner`, `LiveKitToolset`, `current_call`, `perform_rpc`, `livekit_caller_phone_number`, DTMF buffering) exists in [adk-python `src/google/adk/integrations/livekit/`](https://github.com/google/adk-python/blob/main/src/google/adk/integrations/livekit/__init__.py#L26-L36). No internal links, no secrets, no real personal data (`+15105550100` is in the reserved 555-01xx block), links all resolve, and the logo matches [CONTRIBUTING.md:235-238](https://github.com/google/adk-docs/blob/e9c75e6a1cabaf657373dae324768b819d91a078/CONTRIBUTING.md#L235-L238).

One item worth acting on before merge — the missing caller-PII caveat — plus three minor wording fixes, all left as inline suggestions.

joefernandez (commented):
CODE REVIEW: Automated content review of the new LiveKit page. Every API claim checks out against ADK Python v2.9.0: [`src/google/adk/integrations/livekit/`](https://github.com/google/adk-python/tree/v2.9.0/src/google/adk/integrations/livekit) ships in the tag and the PyPI wheel, the `livekit` extra exists ([pyproject.toml#L250-L254](https://github.com/google/adk-python/blob/v2.9.0/pyproject.toml#L250-L254)), `FeatureName.LIVEKIT` is `default_on=True` so no opt-in step is needed ([_feature_registry.py#L178](https://github.com/google/adk-python/blob/v2.9.0/src/google/adk/features/_feature_registry.py#L178)), and the `LiveKitRunner` / `current_call` / `perform_rpc` signatures match the snippets exactly. Overall: medium risk, approve with the inline changes.

Two things are worth fixing before merge: the `tool_context.state[...]` subscript (`KeyError` under `adk web` or on any WebRTC call) and the Agent Runtime deployment claim, which this repo's own Agent Runtime docs do not support for a container with a custom worker entrypoint. The rest are small accuracy nits.

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+integration acts as a transport adapter that handles audio and video capture, playback,
+barge-in, captions, and call control, so your ADK agent can become reachable from a browser,
+a phone call, or a game client without any change to the agent.
+
+## Use cases
+
+### Browser and mobile apps
+
+The agent joins a room as an ordinary participant, so any LiveKit client SDK can talk to it.
+The connector publishes captions, speaking state, and barge-in on the channels LiveKit's own
```
joefernandez: CODE REVIEW: Barge-in is not a published channel. Captions (`lk.transcription`) and speaking state (`lk.agent.state`) are channels; barge-in is playback truncation done by the playback task ([README.md#L239-L240](https://github.com/google/adk-python/blob/v2.9.0/src/google/adk/integrations/livekit/README.md#L239-L240)).

```suggestion
The connector publishes captions and speaking state on the channels LiveKit's own
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+| [Client SDKs](https://docs.livekit.io/transport/) | Browser, Swift, Android, Flutter, React Native, Unity, C++, Rust, and ESP32 |
+| [UI components](https://github.com/orgs/livekit/repositories?q=components) | Prebuilt voice-assistant widgets for React, SwiftUI, Compose, and Flutter |
+| [Starter apps](https://github.com/livekit-examples) | Working apps per platform, plus the Agents Playground for talking to an agent with no front end at all |
+
+### Phone calls
+
+A SIP caller is an ordinary LiveKit participant, so a phone call reaches the agent once an
+[inbound trunk](https://docs.livekit.io/telephony/accepting-calls/inbound-trunk/) and a
+[dispatch rule](https://docs.livekit.io/telephony/accepting-calls/dispatch-rule/) point at
+your worker.
```
joefernandez, line 41: CODE REVIEW: Upstream says telephony is not verified end to end — "Only WebRTC is verified end to end. The telephony paths are unit-tested but have not been run against a real trunk" ([README.md#L205-L207](https://github.com/google/adk-python/blob/v2.9.0/src/google/adk/integrations/livekit/README.md#L205-L207)). Worth a note here, since this section presents SIP as a first-class path.

```suggestion
your worker.

!!! note
    Telephony is unit-tested but has not yet been verified against a live trunk; WebRTC
    is the path validated end to end.
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+
+Caller identity lands in ADK session state before the caller speaks, so a function tool reads
+it like any other state value:
+
+```python
+from google.adk.tools.tool_context import ToolContext
+
+async def greet_by_account(tool_context: ToolContext) -> str:
+  """Looks the caller up before greeting them."""
+  number = tool_context.state["livekit_caller_phone_number"]  # '+15105550100'
```
joefernandez: CODE REVIEW: This subscript raises `KeyError` on any non-SIP call. The key is only written when a SIP participant is in the room ([_livekit_runner.py#L345-L365](https://github.com/google/adk-python/blob/v2.9.0/src/google/adk/integrations/livekit/_livekit_runner.py#L345-L365)), and the source notes "Every read is optional" ([_livekit_call.py#L41-L47](https://github.com/google/adk-python/blob/v2.9.0/src/google/adk/integrations/livekit/_livekit_call.py#L41-L47)). The same tool under `adk web` or on a WebRTC call blows up.

```suggestion
  number = tool_context.state.get("livekit_caller_phone_number")  # '+15105550100'
  if not number:
    return "I could not see the number you are calling from."
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+```python
+from google.adk.agents import Agent
+from google.adk.integrations.livekit import LiveKitToolset
+from google.adk.runners import InMemoryRunner
+
+root_agent = Agent(
+    model="gemini-live-2.5-flash-native-audio",
+    name="support_agent",
+    instruction="You help customers troubleshoot their home internet.",
+    tools=[check_line_status, LiveKitToolset()],
```
joefernandez: CODE REVIEW: `check_line_status` is undefined, so this snippet is not runnable as shown. Mark it as the reader's own tool, the way `crm` is marked on line 50.

```suggestion
    tools=[check_line_status, LiveKitToolset()],  # check_line_status is your own tool
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+The worker dials out to the LiveKit server and receives dispatched jobs over that same
+connection, so it needs outbound network access, no public address, and no load balancer.
+Otherwise it is a normal ADK container and deploys the way any ADK agent does. Point the
+entrypoint at your worker module:
+
+```dockerfile
+CMD ["python", "-m", "support_agent.livekit_worker", "start"]
+```
+
+[Agents CLI](../get-started/agents-cli.md) deploys that container to Agent Runtime, Cloud
```
joefernandez, line 149: CODE REVIEW: Agent Runtime does not take a container with a custom worker `CMD`. This repo's Agent Runtime page describes the payload as agent code only, with the service supplying the API server ([agent-runtime/index.md, "Deployment payload"](https://github.com/google/adk-docs/blob/main/docs/deploy/agent-runtime/index.md#L40-L56)), and a LiveKit worker is a long-lived process that dials out. Scope the sentence to the targets that accept a container (and drop the now-stray comma on the next line, `Run or GKE from ...`).

```suggestion
[Agents CLI](../get-started/agents-cli.md) deploys that container to Cloud
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+[Deploy with Agents CLI](../deploy/agent-runtime/agents-cli.md), or deploy by hand to
+[Cloud Run](../deploy/cloud-run.md) or [GKE](../deploy/gke.md).
+
+Cloud Run probes `$PORT`, while the worker serves its health endpoint on a fixed port, so
+make the two match. Read the port from the environment when you create the server:
+
+```python
+import os
+
+server = AgentServer(port=int(os.environ["PORT"]))
```
joefernandez, line 160: CODE REVIEW: `os.environ["PORT"]` raises `KeyError` whenever `PORT` is unset, which is every local `... start` run. Default to the documented production port ([worker.py#L301](https://github.com/livekit/agents/blob/livekit-agents%401.8.1/livekit-agents/livekit/agents/worker.py#L301)).

```suggestion
server = AgentServer(port=int(os.environ.get("PORT", 8081)))
```

docs/live/custom-server.md
```diff
@@ -164,6 +164,10 @@ Your server exposes a WebSocket; something has to talk to it. During development
 telephony or WebRTC bridge. Whatever you build inherits the same contract, so it is worth
 knowing exactly what `adk web` does and where it stops.
 
+If a WebRTC or telephony bridge is what you need, ADK ships one. The
+[LiveKit runner](../integrations/livekit.md) serves a live agent into a LiveKit room,
+handling media capture, playback, barge-in, and SIP, so there is no custom server to write.
```
joefernandez, line 169: CODE REVIEW: "no custom server to write" overstates it slightly — the reader still writes the ~15-line worker module the LiveKit page shows ([sample worker](https://github.com/google/adk-python/blob/main/contributing/samples/integrations/livekit/livekit_worker.py#L32-L53)).

```suggestion
handling media capture, playback, barge-in, and SIP, so there is no WebSocket server or
media plumbing to write.
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+
+<div class="language-support-tag">
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python v2.9.0</span><span class="lst-preview">Experimental</span>
+</div>
+
+ADK provides the `LiveKitRunner` class to allow you to serve your live agents over
+[LiveKit](https://livekit.io/), an open source platform for WebRTC and SIP telephony. This
+integration acts as a transport adapter that handles audio and video capture, playback,
+barge-in, captions, and call control, so your ADK agent can become reachable from a browser,
+a phone call, or a game client without any change to the agent.
```
joefernandez: STYLE REVIEW: Mixed parallelism — a browser and a game client are clients, but a phone call is an event.

```suggestion
a phone, or a game client without any change to the agent.
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Use cases
+
+### Browser and mobile apps
+
+The agent joins a room as an ordinary participant, so any LiveKit client SDK can talk to it.
+The connector publishes captions, speaking state, and barge-in on the channels LiveKit's own
+components bind to, so most of a voice front end already exists:
```
joefernandez: STYLE REVIEW: "most of a voice front end already exists" is vague — the reader cannot tell what fraction of what exists. State the fact.

```suggestion
components bind to, so those components work against an ADK agent without extra wiring:
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+
+| Tier | What an ADK agent gets |
```
joefernandez: STYLE REVIEW: "Tier" is a misnomer — the rows are client SDKs, UI components, and starter apps, which are categories of LiveKit resource, not tiers. No other table in the docs uses this header.

```suggestion
| LiveKit resource | What an ADK agent gets |
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+[dispatch rule](https://docs.livekit.io/telephony/accepting-calls/dispatch-rule/) point at
+your worker.
+
+Caller identity lands in ADK session state before the caller speaks, so a function tool reads
+it like any other state value:
+
+```python
+from google.adk.tools.tool_context import ToolContext
+
+async def greet_by_account(tool_context: ToolContext) -> str:
```
joefernandez, line 49: STYLE REVIEW: This function body uses 2-space indentation, but the `Agent(...)` call at [lines 98-103](https://github.com/google/adk-docs/blob/e9c75e6a1cabaf657373dae324768b819d91a078/docs/integrations/livekit.md#L98-L103) and the `LiveKitRunner(...)` call at [125-127](https://github.com/google/adk-docs/blob/e9c75e6a1cabaf657373dae324768b819d91a078/docs/integrations/livekit.md#L125-L127) use 4, so the page is inconsistent with itself. 4 spaces is also the prevailing style in `docs/` — see [live/tools.md:72-75](https://github.com/google/adk-docs/blob/e9c75e6a1cabaf657373dae324768b819d91a078/docs/live/tools.md#L72-L75) and [slack.md:48-60](https://github.com/google/adk-docs/blob/e9c75e6a1cabaf657373dae324768b819d91a078/docs/integrations/slack.md#L48-L60). Reindent the body:

```python
async def greet_by_account(tool_context: ToolContext) -> str:
    """Looks the caller up before greeting them."""
    number = tool_context.state["livekit_caller_phone_number"]  # '+15105550100'
    return await crm.lookup(number)  # your own customer lookup
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+  number = tool_context.state["livekit_caller_phone_number"]  # '+15105550100'
+  return await crm.lookup(number)  # your own customer lookup
+```
+
+The connector buffers keypad entry into a single turn, so a six-digit account number arrives
+as one input instead of six interruptions.
+
+### Games and immersive clients
+
+LiveKit's [Unity SDK](https://github.com/livekit/client-sdk-unity) adds realtime audio,
```
joefernandez: STYLE REVIEW: Use "real-time". The docs use "real-time" 38 times and "realtime" only twice, both incidental.

```suggestion
LiveKit's [Unity SDK](https://github.com/livekit/client-sdk-unity) adds real-time audio,
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+video, and data channels to a Unity app, backed by LiveKit Cloud or a server you host. Put an
+ADK agent in the room and a player can hold a conversation with a character that also acts
+on the world, such as a voice-driven NPC or an in-game assistant:
+
+```python
+from google.adk.integrations.livekit import current_call
+from google.adk.tools.tool_context import ToolContext
+
+async def open_the_door(door_id: str, tool_context: ToolContext) -> str:
```
joefernandez, line 71: STYLE REVIEW: Same 2-space body as the earlier snippet. Reindent to 4 spaces:

```python
async def open_the_door(door_id: str, tool_context: ToolContext) -> str:
    """Opens a door in the game world."""
    call = current_call(tool_context)
    return await call.perform_rpc(method="open_door", payload=door_id)
```

joefernandez (commented):
STYLE REVIEW: Style pass on the new LiveKit integration page. Nothing here is blocking — the page is grammatical, active-voice, second-person, and free of the banned language patterns, and every relative link resolves.

Two things worth fixing before merge:

1. **Vague and editorializing sentences.** A reader cannot act on "most of a voice front end already exists" or "the choice is operational". Inline suggestions state the fact instead.
2. **Structure and indentation diverge from the integrations catalog.** `## Get started` holds prerequisites, installation, and the first example; 76 of the 105 pages in `docs/integrations/` use `## Prerequisites` followed by `## Installation`, including the closest analog, [slack.md:11-23](https://github.com/google/adk-docs/blob/e9c75e6a1cabaf657373dae324768b819d91a078/docs/integrations/slack.md#L11-L23). Python function bodies use 2-space indentation while the `Agent(...)` and `LiveKitRunner(...)` calls in the same file use 4.

One code comment also contradicts the snippet it annotates. The rest are minor wording and terminology nits.

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+  """Opens a door in the game world."""
+  call = current_call(tool_context)
+  return await call.perform_rpc(method="open_door", payload=door_id)
+```
+
+Whatever the client returns becomes the tool result the model narrates, so the agent
+describes what actually happened. The Unity side is one registered RPC method, and ADK keeps
```
joefernandez: STYLE REVIEW: "The Unity side is one registered RPC method" comments on the integration instead of telling the reader what to do. Give the instruction.

```suggestion
describes what actually happened. On the Unity client, register one RPC method; ADK manages
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+the conversation, the tool calls, and the session.
+
+## Get started
```
joefernandez, line 81: STYLE REVIEW: `## Get started` holds prerequisites, installation, and the first example. 76 of the 105 pages in `docs/integrations/` use `## Prerequisites` followed by `## Installation` — only [environment-toolset.md](https://github.com/google/adk-docs/blob/e9c75e6a1cabaf657373dae324768b819d91a078/docs/integrations/environment-toolset.md) uses `## Get started`. Readers scanning the catalog for prerequisites will miss them. Split the section (see the companion suggestion at line 85):

```suggestion
## Prerequisites
```

docs/integrations/livekit.md
joefernandez: STYLE REVIEW: "the choice is operational" leaves "operational" undefined. Say what it means for the reader.

```suggestion
  same worker code runs against either. For local development, run `livekit-server --dev`.
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+
+- [ADK](https://adk.dev) >= 2.9.0 with the `livekit` extra.
+- Credentials for a [live model](../live/models.md).
+- A LiveKit server, self-hosted or on LiveKit Cloud. Both expose the same API, so the
+  choice is operational. For local development, run `livekit-server --dev`.
+- `LIVEKIT_URL`, `LIVEKIT_API_KEY`, and `LIVEKIT_API_SECRET` set in the environment.
```
joefernandez, line 87: STYLE REVIEW: The bullet starts with code syntax. Qualify it with English words first.

```suggestion
- The `LIVEKIT_URL`, `LIVEKIT_API_KEY`, and `LIVEKIT_API_SECRET` environment variables.
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+
+```bash
```
joefernandez, line 89: STYLE REVIEW: Companion to the `## Prerequisites` suggestion above — add the `## Installation` heading the rest of the catalog uses, as in [slack.md:17-23](https://github.com/google/adk-docs/blob/e9c75e6a1cabaf657373dae324768b819d91a078/docs/integrations/slack.md#L17-L23).

````suggestion
## Installation

```bash
````

File: docs/integrations/livekit.md [hunk 24]
```diff
@@ -0,0 +1,171 @@
...
+root_agent = Agent(
+    model="gemini-live-2.5-flash-native-audio",
+    name="support_agent",
+    instruction="You help customers troubleshoot their home internet.",
+    tools=[check_line_status, LiveKitToolset()],
+)
+runner = InMemoryRunner(agent=root_agent, app_name="support")
+```
+
+Handing that runner a connected room is the whole integration. In production you run a
```
joefernandez: STYLE REVIEW: "is the whole integration" dramatizes rather than instructs. Lead with the action.

```suggestion
To connect the agent, pass the runner a connected room. In production you run a
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,171 @@
...
+from google.adk.integrations.livekit import LiveKitRunner
+from livekit.agents import AgentServer
+from livekit.agents import cli
+from livekit.agents import JobContext
+
+server = AgentServer()
+
+
+@server.rtc_session(agent_name="support")
+async def entrypoint(ctx: JobContext) -> None:
```
joefernandez, line 125: STYLE REVIEW: Same 2-space body as the two earlier snippets — reindent lines 122-127 and 131 to 4 spaces, so the body matches the `LiveKitRunner(...)` arguments it already contains at 4.

docs/integrations/livekit.md
joefernandez: STYLE REVIEW: The comment contradicts the code below it — the snippet hardcodes `user_id="live-user"` and derives `session_id` from `ctx.room.name`; nothing reads job metadata. Also "ids" should be "IDs". Describe this snippet and keep the sample as a forward reference (reindent with the rest of the block):

```suggestion
  # LiveKit has no IDs of its own, so supply them here. This snippet derives the
  # session ID from the room name; the linked sample reads both from job metadata.
```

docs/integrations/livekit.md
joefernandez: STYLE REVIEW: The clause after the semicolon starts with code syntax. Split it and qualify the class name.

```suggestion
[session service](../sessions/index.md). The `InMemoryRunner` class persists nothing
between calls.
```

docs/live/custom-server.md
```diff
@@ -164,6 +164,10 @@ Your server exposes a WebSocket; something has to talk to it. During development
 telephony or WebRTC bridge. Whatever you build inherits the same contract, so it is worth
 knowing exactly what `adk web` does and where it stops.
 
+If a WebRTC or telephony bridge is what you need, ADK ships one. The
```
joefernandez: STYLE REVIEW: Inverted construction. Direct second person is simpler.

```suggestion
If you need a WebRTC or telephony bridge, ADK provides one. The
```

docs/integrations/livekit.md
```diff
@@ -0,0 +1,174 @@
...
+</div>
+
+ADK provides the `LiveKitRunner` class to allow you to serve your live agents over
+[LiveKit](https://livekit.io/), an open source platform for WebRTC and SIP telephony. This
+integration acts as a transport adapter that handles audio and video capture, playback,
+barge-in, captions, and call control, so your ADK agent can become reachable from a browser,
+a phone, or a game client without any change to the agent.
+
+## Use cases
+
```
joefernandez, line 21: STYLE: don't stack headers. add an introductory sentence or two summarizing the use cases.

```suggestion

You can use LiveKit in a variety of use cases, including browser apps, mobile apps, SIP phone calls, games and immersive clients.

```

joefernandez (approved):
Approved with one comment.

PR #2225, opened 2026-09-12
docs: fix 404 link to relocated Deep Search sample

joefernandez:
## Summary

Fixes one of the four broken links currently failing the `Check external links` workflow on `main`.

The `adk-samples` repo restructured its layout, moving the Deep Search sample out of `python/agents/`. The link in the grounding overview was left pointing at the old path.

| | URL | Status |
|---|---|---|
| Before | `.../adk-samples/tree/main/python/agents/deep-search` | 404 |
| After | `.../adk-samples/tree/main/core/python/deep-search` | 200 |

New target verified: https://github.com/google/adk-samples/tree/main/core/python/deep-search

## Context

`main` has been failing the link check since 9f21af3c (Sep 10). Because [`link-checker.yaml`](https://github.com/google/adk-docs/blob/main/.github/workflows/link-checker.yaml#L41) globs `./**/*.md` (the whole repo, not just changed files), this breakage fails the check on **every open PR**, regardless of what those PRs touch.

The remaining three errors are dead `Perseus-Computing-LLC` links in `docs/integrations/perseus.md` and `docs/integrations/perseus-vault.md`; those are handled separately since they need the integration author's input.

## Testing

Verified the new URL returns 200 and the old one returns 404. No other occurrences of the stale path exist in the repo.

PR #1928, opened 2026-07-04
docs(integrations): rename Mimir Memory → Perseus Vault Memory (supersedes #1877)

tcconnally:
Renames and updates the memory integration merged in #1877 to reflect the product's current name. The backend previously documented as "Mimir" was renamed to **Perseus Vault**, and the PyPI package was renamed to `adk-perseus-vault-memory` (v0.3.0). The class is now `PerseusVaultMemoryService`.

This supersedes #1877 (merged 2026-06-26), which documented the old `adk-mimir-memory` / `MimirMemoryService` names.

### Changes
- Rename `docs/integrations/mimir.md` → `docs/integrations/perseus-vault.md`, updating package name, class name, binary name (`perseus-vault`), and install/config snippets to match the current README.
- Add `docs/integrations/assets/perseus-vault.svg` (new logo); remove the old `mimir.svg`.
- Add a redirect `integrations/mimir.md` → `integrations/perseus-vault.md` in `mkdocs.yml` so the existing (merged) URL keeps working.
- Fix the two cross-links in `docs/integrations/perseus.md` that pointed at the old `/integrations/mimir/` slug.

### Verification
- Package on PyPI: https://pypi.org/project/adk-perseus-vault-memory/ (v0.3.0, MIT, Python 3.10–3.13)
- Import/class verified against the source: `from adk_perseus_vault_memory import PerseusVaultMemoryService`
- Source repo (unchanged name): https://github.com/Perseus-Computing-LLC/adk-mimir-memory

### Note for maintainers
Integration pages are auto-discovered from `catalog_*` frontmatter (`not_in_nav: /integrations/*.md`), so no nav entry is needed. If you'd prefer to keep the `mimir.md` filename to minimize churn, I can instead rewrite the file's contents in place and drop the rename + redirect — happy to adjust.

PR #2226, opened 2026-09-12
docs(integrations): remove dead Perseus repo links

joefernandez:
## Summary

Removes the three remaining broken links failing the `Check external links` workflow on `main`. Companion to #2225 (which fixes the fourth, unrelated one).

Both backing repositories return 404:

| File | Dead URL |
|---|---|
| `perseus.md` | `Perseus-Computing-LLC/perseus` |
| `perseus-vault.md` | `Perseus-Computing-LLC/perseus-vault` |
| `perseus-vault.md` | `Perseus-Computing-LLC/perseus-vault/releases` |

Nothing in this repo broke them — they rotted externally. The first failing run on `main` was 9f21af3c (Sep 10), an unrelated TypeScript version bump.

## Why removal rather than a redirect

I looked for a replacement before deleting anything:

- Neither `perseus` nor `perseus-vault` exists among the **31 public repos** in the `Perseus-Computing-LLC` org. The org's public presence is the integration libraries plus satellites like `perseus-vault-demo`.
- The `adk-perseus-vault-memory` PyPI metadata still lists `.../perseus-vault` as a project URL, so upstream also expects that repo to exist — it appears to have gone private rather than moved.
- `perseus.computer` is an **unrelated company** (efficient systems inc.), not this vendor. Not a valid substitute.

So there is no URL to swap in. Links are dropped, prose kept.

The two surviving `Perseus-Computing-LLC` links (`adk-perseus-context`, `adk-perseus-vault-memory`) both resolve 200 and are untouched.

## @tcconnally — two things need your call

You authored #1891 and #1928, so flagging these rather than deciding them here:

1. **The install step now has no source for the binary.** `perseus-vault.md` told readers to download from a releases page that 404s, so those instructions were already unfollowable. I reduced it to "install the `perseus-vault` binary and place it on your `PATH`" — but readers still have no documented way to *obtain* it. If there's a public download (crates.io, a tarball, an install script), that's the right fix.

2. **`perseus.md` still calls Perseus "an open-source context compiler."** The only evidence for that claim was the link I just removed. I left the wording alone deliberately, but if the source isn't public, it should probably be reworded.

Separately, a **non-blocking nit** left out to keep this diff scoped: `perseus-vault.md` lines 14 and 147 still point at `adk-mimir-memory`, which only resolves via a redirect to `adk-perseus-vault-memory`. Leftovers from the #1928 rename — worth updating whenever you next touch the page.

## Testing

Verified all three removed URLs return 404 and both retained ones return 200. No other references to the dead repos remain in the repo.

### DIFF
--- a/docs/integrations/perseus-vault.md
+++ b/docs/integrations/perseus-vault.md
@@ -12,12 +12,11 @@ catalog_tags: ["data"]
 </div>
 
 The [`adk-perseus-vault-memory`](https://github.com/Perseus-Computing-LLC/adk-mimir-memory)
-integration connects your ADK agent to
-[Perseus Vault](https://github.com/Perseus-Computing-LLC/perseus-vault), a
-persistent, cross-session memory backend. Backed by a single Rust binary with an
-embedded SQLite database, it requires **zero cloud dependencies**, and everything
-runs locally. Memory is encrypted at rest with AES-256-GCM, and search combines
-FTS5 keyword matching with dense vector retrieval.
+integration connects your ADK agent to Perseus Vault, a persistent,
+cross-session memory backend. Backed by a single Rust binary with an embedded
+SQLite database, it requires **zero cloud dependencies**, and everything runs
+locally. Memory is encrypted at rest with AES-256-GCM, and search combines FTS5
+keyword matching with dense vector retrieval.
 
 ## Use cases
 
@@ -44,11 +43,9 @@ Install the Python package:
 pip install adk-perseus-vault-memory
 ```
 
-Then install the `perseus-vault` binary: download the build for your platform
-from the
-[releases page](https://github.com/Perseus-Computing-LLC/perseus-vault/releases)
-and place it on your `PATH`. The service looks for `perseus-vault` by default, or
-pass `vault_binary="/absolute/path/to/perseus-vault"` to
+Then install the `perseus-vault` binary and place it on your `PATH`. The service
+looks for `perseus-vault` by default, or pass
+`vault_binary="/absolute/path/to/perseus-vault"` to
 `PerseusVaultMemoryService`.
 
 ## Use with agent
@@ -146,5 +143,4 @@ session = await runner.session_service.create_session(
 
 - [adk-perseus-vault-memory on GitHub](https://github.com/Perseus-Computing-LLC/adk-mimir-memory)
 - [adk-perseus-vault-memory on PyPI](https://pypi.org/project/adk-perseus-vault-memory/)
-- [Perseus Vault (backing service)](https://github.com/Perseus-Computing-LLC/perseus-vault)
 - [Perseus Context integration](/integrations/perseus/)
--- a/docs/integrations/perseus.md
+++ b/docs/integrations/perseus.md
@@ -13,12 +13,10 @@ catalog_tags: ["data", "mcp"]
 
 The [`adk-perseus-context`](https://github.com/Perseus-Computing-LLC/adk-perseus-context)
 integration injects a deterministically compiled context into your ADK agent's
-system instruction. It is powered by
-[Perseus](https://github.com/Perseus-Computing-LLC/perseus), an open-source
-context compiler: Perseus resolves directives like `@file`, `@search`, and
-`@memory` into one byte-stable context string at inference time, with no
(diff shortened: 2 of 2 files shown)


joefernandez:
@tcconnally — requesting your review as the author of #1891 and #1928.

GitHub won't let me add you as a formal reviewer (not a repo collaborator), so tagging you here instead.

Short version: the `Perseus-Computing-LLC/perseus` and `.../perseus-vault` repos now 404, which is failing the link checker on `main` and therefore on every open PR. I couldn't find a public replacement for either, so this PR drops the three dead links and leaves the prose intact.

Two things I deliberately did **not** decide on your behalf:

1. The Perseus Vault install step no longer tells readers where to get the `perseus-vault` binary — the releases page it pointed to is gone. Is there a public download?
2. `perseus.md` still describes Perseus as "an open-source context compiler," but the link backing that claim is what I removed.

Happy to fold in whatever you'd prefer, or to close this if you'd rather fix it upstream by making the repos public.

tcconnally:
@joefernandez Thanks for flagging this. This is fallout from an intentional privatization, not a move: both backing repositories are private now. There is no public replacement or public `perseus-vault` binary download at present—the release artifacts are behind the private repository, and the binary is not published to crates.io or PyPI. Please keep the dead-link removals; there is no need to close this PR in favor of making the repositories public.

I agree with two follow-ups:

- Change the two remaining `adk-mimir-memory` links to `adk-perseus-vault-memory` (the old URL only redirects).
- Remove or reword “open-source context compiler” in `perseus.md`; the implementation may be MIT-licensed, but its source is not publicly reachable now.

If you can fold those two cleanup edits into this PR, I will consider the result approved from my side. The install docs should not imply that readers can obtain the binary until we publish a public distribution path.

docs/integrations/perseus-vault.md
```diff
@@ -44,11 +43,9 @@ Install the Python package:
 pip install adk-perseus-vault-memory
 ```
 
-Then install the `perseus-vault` binary: download the build for your platform
-from the
-[releases page](https://github.com/Perseus-Computing-LLC/perseus-vault/releases)
-and place it on your `PATH`. The service looks for `perseus-vault` by default, or
-pass `vault_binary="/absolute/path/to/perseus-vault"` to
+Then install the `perseus-vault` binary and place it on your `PATH`. The service
```
KevinACheek: Not possible to install a current perseus-vault binary, as downloads and releases on the Perseus Computing site are paused. 

How about rewriting this for cases where the user has previously installed the `perseus-vault` binary.

docs/integrations/perseus-vault.md
KevinACheek: Recommend: remove this section.

KevinACheek: Recommend: 
`- The `perseus-vault` binary already installed. See [Perseus Computing Vault]*(https://perseus.observer/vault/) for availability and installation instructions.

KevinACheek (commented):
With the `perseus-vault` binary unavailable for download, we should not include instructions to install it. I recommend rewriting this to say the binary should already be installed (check the Perseus Computing Vault site for availability) and remove the installation step. I think it is important to provide the link, because there the site is at perseus.observer and there is another site at perseus.computer which is not the same thing. With the similarity in naming I worry users might find the other site in trying to locate the binary.

docs/integrations/perseus-vault.md
```diff
@@ -32,22 +32,9 @@ keyword matching with dense vector retrieval.
 ## Prerequisites
 
 - Python 3.10+
-- The `perseus-vault` binary (see [Installation](#installation))
+- The `perseus-vault` binary already installed. See [Perseus Computing Vault](https://perseus.observer/vault/) for availability and instructions.
```
joefernandez, line 35: LGTM! Thanks

docs/integrations/perseus-vault.md
joefernandez: agreed

docs/integrations/perseus-vault.md
joefernandez: Let's assume this a temporary problem that users need to take up with the Perseus team. Leave as-is and ship it.

KevinACheek (approved):
Links checked manually.

PR #2229, opened 2026-09-13
docs(integrations): add OrcaReplay

xizhuomengcontin:
Adds one integration page, `docs/integrations/orcareplay.md`, plus its card icon. No `mkdocs.yml` edit — the catalog is globbed.

## What it is

[OrcaReplay](https://github.com/Continuum-AI-Corp/OrcaReplay) (Apache-2.0) records an agent run at the HTTP boundary between the agent and its model provider, **from outside the process**, and can serve that recording back so the same run happens again with no provider called and no API key needed. Nothing is installed into the agent — `orca record` launches your ADK script as a child process with the provider origin redirected for that process only.

For ADK specifically, that turns a session into an artifact: a run that misbehaved can be reproduced by someone else without your key and without spending tokens, and the same recording works as a zero-cost regression test in CI.

## The example in the page is one I actually ran

Your first acceptance criterion is that the integration be functional and testable, so here is what produced the output quoted on the page — `google-adk` with `LiteLlm`, a deterministic local origin standing in for a provider, on Python 3.12:

```console
$ orca record generic-openai -- python agent.py
adk replied: the stub answered
info recorded run=run_135c37894350 events=6 exit=0

# the origin process is killed here

$ orca replay last
info replaying run=run_135c37894350 exchanges=1 egress=blocked
adk replied: the stub answered
info replay.done reused=1/1 exact=1 divergences=0 unmatched=0 exit=0
```

`exact=1` is a byte-identical request match, and the origin was not running for the replay. The page's code block is that script with the stub-specific bits removed.

## Choices I made, and why

- **`LiteLlm` rather than `google-genai`.** The recorder needs the origin to come from the environment; `LiteLlm(model="openai/…")` with no `base_url` in code does exactly that. A Gemini-backed path is possible but I have not verified it end to end, so the page does not claim it.
- **Python only in the language-support tag.** The mechanism is process-level and should hold for the TypeScript SDK too, but I only verified Python and would rather under-claim.
- **No `catalog_tags`.** The honest candidates would have been `observability` or `evaluation`, and both would overstate what this is — there are no metrics, no judges and no dashboards here. Inventing a new tag would add a lone orphan filter button, so I left tags off, as `copilotkit.md` does. Happy to add one if you would rather every card carry a tag.
- **Icon** is a square 256×256 PNG (55 KB) at `docs/integrations/assets/orcareplay.png`.
- The page has an explicit **Limits** section: `egress=blocked` means provider egress and not network isolation (recorded tool calls still run for real, so it is not a sandbox), a matching replay is not a determinism result, and a trace is a full transcript that needs reviewing before it is shared.

On publishability: OrcaReplay's TLS interception is opt-in and restricted to an explicit allowlist of model-provider hosts — it is not a general-purpose intercept and the allowlist cannot be set to `*`. Everything else works by setting environment variables for a process the user launches themselves.

## CLA

I have not signed the Google CLA yet — flagging it rather than letting the bot be the one to tell you. I will get it signed; please do not let it hold up review feedback if you have any.

Disclosure: I maintain OrcaReplay.


docs/integrations/orcareplay.md
```diff
@@ -0,0 +1,143 @@
...
+  <span class="lst-supported">Supported in ADK</span><span class="lst-python">Python</span>
+</div>
+
+[OrcaReplay](https://github.com/Continuum-AI-Corp/OrcaReplay) records an agent run from **outside
+the process** — at the HTTP boundary between your agent and its model provider — and can serve that
+recording back so the same run happens again with no provider called and no API key needed.
+
+Nothing is installed into your agent. `orca record` launches your ADK script as a child process with
+the provider origin redirected for that process only, so the code that runs is the code you wrote,
+not an instrumented variant of it.
```
joefernandez, line 19: This paragraph is the design premise of the tool, and it is also why the page does not fit `docs/integrations/`: recording happens at the provider HTTP boundary, outside the process, so nothing on the ADK side is touched. An integration page needs an ADK-specific artifact (plugin, callback, toolset, or MCP server) for the page to be about ADK rather than about running any Python process under a proxy.

docs/integrations/orcareplay.md
```diff
@@ -0,0 +1,143 @@
...
+                    print(part.text, end="")
+
+
+asyncio.run(main())
+```
+
+Record it:
+
+```bash
+orca record generic-openai -- python agent.py
```
joefernandez, line 94: This shell line is the entire integration. The sample above it (lines 53-89) is vanilla ADK with no OrcaReplay code in it, so replacing `agent.py` with a script written against any other Python LLM framework leaves this page unchanged.

An ADK plugin or callback that recorded the run from inside the agent would be an ADK-specific artifact worth documenting here. Running `orca record` against an arbitrary subprocess is not, and belongs in the OrcaReplay docs.

joefernandez (commented):
Thank you for the PR, and for your interest in ADK.

After reviewing this against our [acceptance criteria](https://github.com/google/adk-docs/blob/main/CONTRIBUTING.md#acceptance-criteria), this page is out of scope for the ADK docs, so we are going to close the PR.

**Out of scope:** integration pages document third-party functionality that extends or tightly integrates with ADK: a tool, plugin, toolset, MCP server, or observability integration with an ADK-specific artifact. OrcaReplay records at the HTTP boundary between the process and the model provider, which the page itself states as its design premise ([`orcareplay.md:17-19`](https://github.com/google/adk-docs/blob/integrations/orcareplay/docs/integrations/orcareplay.md#L17-L19)):

> Nothing is installed into your agent. `orca record` launches your ADK script as a child process with the provider origin redirected for that process only, so the code that runs is the code you wrote, not an instrumented variant of it.

The page uses no ADK extension point: no plugin, no callback, no `BaseToolset`, no MCP server. The Python sample at [`orcareplay.md:53-89`](https://github.com/google/adk-docs/blob/integrations/orcareplay/docs/integrations/orcareplay.md#L53-L89) is a plain ADK script with no OrcaReplay code in it at all, and the whole integration is the shell line at [`orcareplay.md:94`](https://github.com/google/adk-docs/blob/integrations/orcareplay/docs/integrations/orcareplay.md#L94), `orca record generic-openai -- python agent.py`. Substitute any other Python LLM framework for `agent.py` and the page reads the same. The ADK docs are not a directory of tools that can be run alongside an agent.

This is not a judgment on the tool. Process-level record and replay is genuinely useful, the page is well written, and the Limits section is more honest than most. It just is not ADK-specific, which is what an entry in `docs/integrations/` has to be.

If OrcaReplay grows an ADK-specific artifact, for example a plugin or callback that captures a run from inside the agent (ADK events, tool calls and session state, not only provider HTTP), we would be glad to look at a new PR for that.

Thanks again for taking the time to write this up.

xizhuomengcontin:
No problem with the close — one short question so I learn the right thing from it.

Was this the unsigned CLA, or the content? I flagged the CLA in the description and can get it signed; if that is all it was I am happy to sign and reopen. If it is the content — a third-party record/replay tool is not something the integrations catalog wants — then that is a clear answer and I will not bring it back.

Either way, thanks for looking. The page's example was a run I actually executed against ADK with `LiteLlm`, so if the objection is narrower than "no" (the tag, the Python-only support claim, the Limits section) I would rather fix it than guess.

joefernandez:
@xizhuomengcontin 
> Was this the unsigned CLA, or the content? 

The content. See my earlier comments for more details on that. Thanks again for your interest in the ADK product!

PR #2237, opened 2026-09-15
Add Redpine MCP integration

AlexisMarasigan:
Adds an integration page for the Redpine MCP server under `docs/integrations/`, tagged `mcp` and `search`.

Redpine is a hosted remote MCP server that answers questions from licensed publisher content and cites each result to its source.

- Python `McpToolset` example over `StreamableHTTPConnectionParams` with a Bearer API key header
- Sandbox keys (`sk_test_...`) return sample results at no cost, so the example can be run without spending credits
- Icon added at `docs/integrations/assets/redpine.png`
- Links: https://docs.redpine.ai/docs/mcp, https://docs.redpine.ai/docs/authentication


AlexisMarasigan:
Tested against the live server, and pushed three corrections from what the testing turned up.

**Verified with `google-adk`:**

- `McpToolset` over `StreamableHTTPConnectionParams` to `https://api.redpine.ai/mcp` with the `Authorization: Bearer` header connects and discovers 8 tools: `search-redpine-science`, `preview`, `confirm`, `get_balance`, `find-tools`, `inspect-tool`, `call-tool`, `request_journal`. The imports and class names in the snippet are correct as written.
- A free `preview` call for "CRISPR base editing off-target effects in human cells" returned 5 results with the unlock price shown up front and full citation metadata per result (publisher, journal, year, DOI, PMID/PMCID, authors, linked ClinicalTrials.gov IDs). Nothing was charged for the preview, which is the behaviour the page now describes.

**Corrections pushed since opening the PR:**

- `f927c1f` — the Prerequisites described a sandbox key prefix we do not actually issue, and put API Keys under Settings in the dashboard when it is a top-level item. The page now states what is genuinely free, which is previewing results with the price shown first.
- `0ec1d2a` — added an Installation section for the `mcp` extra. Without it `from google.adk.tools.mcp_tool import McpToolset` raises a bare `ImportError`, because `google/adk/tools/mcp_tool/__init__.py` guards those exports in a `try/except ImportError` and logs the cause at debug level only. Anyone copy-pasting the snippet after a plain `pip install google-adk` hits this. Wording follows `docs/integrations/unstructured.md`.

Happy to adjust anything to fit the catalog conventions. Let me know if you would like the page to cover the TypeScript `McpToolset` as well, and I will test that path before adding it.

AlexisMarasigan:
Self-reviewed this against the repo's own `integration-review` skill and CONTRIBUTING, and pushed the fixes it turned up:

- `a49c992` — **Available tools** now names the eight tools the server actually returns from `tools/list`, in the `Tool | Description` shape CONTRIBUTING specifies, instead of describing categories. The `mcp` extra moved from its own Installation section into Prerequisites, so the section order matches the MCP template the existing MCP pages follow (Use cases, Prerequisites, Use with agent, Available tools, Additional resources).
- `c9a0060` — the catalog icon was 513x512. It is square now.
- `0ec1d2a`, `f927c1f` — the `google-adk[mcp]` extra requirement, and two factual corrections to the Prerequisites described in my previous comment.

Checklist items verified: four-field frontmatter, `catalog_description` at 67 characters and not repeating the product name, both tags already in use elsewhere in the catalog, H1 in `<Product> MCP tool for ADK` form, single-line language-support div with the TypeScript span dropped for a Python-only page, `gemini-flash-latest` rather than a pinned version, canonical `from google.adk.tools.mcp_tool import McpToolset` import and `root_agent` naming, MkDocs tabs, no `mkdocs.yml` nav entry (the catalog auto-discovers), no em dashes, and every external link returning 200.

Could this get the `integration` label and a review when there is a moment? @joefernandez, you have merged the recent integration pages, so flagging it to you rather than leaving it in the queue unlabelled. Happy to add a screenshot of the tools in the ADK Web trace graph, or to cover the TypeScript `McpToolset` path, if either would help.

PR #2197, opened 2026-09-02
docs(get-started): add migrate to ADK guide

eliasecchig:
## Summary

* Add a guide for migrating existing agent codebases to ADK using Agents CLI and an AI coding assistant.
* Include a mapping reference table from existing agent patterns to declarative ADK classes.
* Add navigation entry in `mkdocs.yml` and a card in `docs/get-started/index.md`.

PR #2242, opened 2026-09-16
docs(get-started): fix dangling anchor in the migrate guide workflow list

Jinzhengxu:
The "Migration workflow" list in `docs/get-started/migrate.md` links its second
step to `#brainstorm-the-migration-plan`, but the section heading is
"Brainstorm a migration plan", so the generated anchor is
`#brainstorm-a-migration-plan`. `mkdocs build --strict` reports the missing
anchor and the link does not jump on the published page.

This PR points the list item at the existing anchor and matches the link text
to the heading, which is how the other four items in the list are written.

**Verification**

- `mkdocs build --strict` on `main` logs
  `Doc file 'get-started/migrate.md' contains a link '#brainstorm-the-migration-plan', but there is no such anchor on this page.`;
  on this branch the message is gone and the build still succeeds.
- `git diff --check` clean.

No related issue; the anchor was introduced with the guide in #2197.


joefernandez (approved):
Thanks for the fix!

PR #2244, opened 2026-09-16
docs(runtime): clarify filtered session history retention

baba9811:
Clarifies that Python `RunConfig.get_session_config` limits the loaded event view while preserving stored history when new events are appended. This accompanies the Redis filtered-history preservation fix in `google/adk-python`.

### Testing

- `mkdocs build --strict` passes and indexes 235 pages. It reports an existing missing migration-page anchor at INFO level; this change adds no links.
- `mkdocs serve` returns HTTP 200 and renders the clarification as part of the existing list item.
- Verified the behavior with direct appends and Runner turns on Redis, memory, SQLite and SQLAlchemy SQLite.


joefernandez (commented):
# Technical review report

- **Review Score**: 100
- **Summary reasoning**: The three added sentences in `docs/runtime/runconfig.md` state that `GetSessionConfig` filters (`num_recent_events`, `after_timestamp`) only narrow the loaded view and that new events append to the full stored history. I verified this against every session service shipped in the current release, [adk-python v2.9.1](https://github.com/google/adk-python/releases/tag/v2.9.1), and the claim is correct in each. The API referenced is released, not merged-but-unreleased.
- **Recommended action**: No concerns, approve. The change adds accurate clarification to an already-correct bullet, touches no code samples, and introduces no new API surface. The only observation is stylistic (mild redundancy between the two new sentences), not a technical risk.
- Number of lines changed: XS, 4
- Number of files changed: 1

## Verification detail

The changed text is at [docs/runtime/runconfig.md:76-80](https://github.com/google/adk-docs/blob/60306d1b2c2fc192999d23b39aebe47784a32be9/docs/runtime/runconfig.md#L76-L80), inside a section tagged "Supported in ADK: Python", which matches the fact that `get_session_config` exists only on Python's `RunConfig`: [run_config.py:262-280](https://github.com/google/adk-python/blob/v2.9.1/src/google/adk/agents/run_config.py#L262-L280).

Claim 1 — "These filters limit the loaded view without deleting stored events":

- `InMemorySessionService` slices a *copy* of the session (`_copy_session`, then `copied_session.events = copied_session.events[-num_recent_events:]`); the stored session object is untouched: [in_memory_session_service.py:202-217](https://github.com/google/adk-python/blob/v2.9.1/src/google/adk/sessions/in_memory_session_service.py#L202-L217).
- `DatabaseSessionService` applies the filters as a `WHERE timestamp >= ...` / `LIMIT` on the read query only: [database_session_service.py:706-734](https://github.com/google/adk-python/blob/v2.9.1/src/google/adk/sessions/database_session_service.py#L706-L734).
- `SqliteSessionService` likewise, as SQL `AND timestamp >= ?` / `LIMIT ?`: [sqlite_session_service.py:309-326](https://github.com/google/adk-python/blob/v2.9.1/src/google/adk/sessions/sqlite_session_service.py#L309-L326).
- `VertexAiSessionService` passes `after_timestamp` as a server-side list filter and trims `num_recent_events` client-side on the returned list: [vertex_ai_session_service.py:240-300](https://github.com/google/adk-python/blob/v2.9.1/src/google/adk/sessions/vertex_ai_session_service.py#L240-L300).

Claim 2 — "New events are appended to the stored history, preserving older events excluded from the loaded view":

- In-memory: `append_event` appends to the canonical `storage_session`, even when the caller holds a filtered copy: [in_memory_session_service.py:354](https://github.com/google/adk-python/blob/v2.9.1/src/google/adk/sessions/in_memory_session_service.py#L354).
- Database: `append_event` inserts a new `StorageEvent` row and never rewrites the event table: [database_session_service.py:976](https://github.com/google/adk-python/blob/v2.9.1/src/google/adk/sessions/database_session_service.py#L976).
- The Runner passes `run_config.get_session_config` straight into `session_service.get_session()` on the load path, so filtering is confined to the read: [runners.py:895-900](https://github.com/google/adk-python/blob/v2.9.1/src/google/adk/runners.py#L895-L900).

No risk items found. One optional, non-scoring wording note: "without deleting stored events" and "preserving older events excluded from the loaded view" say the same thing twice; the bullet could be tightened to a single sentence, e.g. `These filters limit the loaded view only: stored events are not deleted, and new events are appended to the full stored history.`

joefernandez (approved):
Thanks for the update!

PR #2249, opened 2026-09-19
docs: fix the third Loop Agent label in the template workflows diagram

vishal-bulbule:
The Template workflows diagram on the [Workflow agents](https://adk.dev/agents/workflow-agents/)
page labels the Loop Agent boxes "Agent A", "Agent B", "Agent B". The Sequence Agent and
Parallel Agent columns show "Agent A", "Agent B", "Agent C", so the third Loop Agent box should
read "Agent C".

The labels in `docs/assets/template_workflows.svg` are text converted to outlines, so this change
replaces the outline of that one label (`path43`) with the "Agent C" outline from the Parallel
Agent column, translated into the Loop Agent box. No other element changes; font, color, and
alignment match the other labels.

Only `docs/agents/workflow-agents/index.md` uses this image. GitHub's rich diff for the SVG shows
the before and after.


docs/assets/template_workflows.svg
```diff
@@ -297,9 +297,10 @@
...
      fill-rule="nonzero"
-     id="path43" />
+     id="path43"
```
joefernandez:
```suggestion
     id="path43" />
```

docs/assets/template_workflows.svg
```diff
@@ -297,9 +297,10 @@
...
+     transform="translate(-177.9804,-8.3645)" />
```
joefernandez:
```suggestion
```

joefernandez (approved):
Thanks for the fix! Approved.

PR #2252, opened 2026-09-22
docs: move the third Loop Agent label back into the Loop box

vishal-bulbule:
Follow-up to #2249.

On main the third box in the Loop Agent column of the Template workflows diagram is now empty.

The "Agent C" label there is a copy of the Parallel column's label outline, and in #2249 it used a
separate `transform` attribute to move it into the Loop box. That line was removed by one of the
review suggestions before merge, so the label now renders at its original position, on top of the
Parallel column's "Agent C", and the Loop box shows nothing.

This change shifts the path's starting point instead of using a `transform`. The rest of the path
uses relative commands, so only the first coordinate changes, and the position no longer depends on
a separate attribute. One line changes in `docs/assets/template_workflows.svg`.

Checked by rendering the SVG in Chrome: main shows an empty third Loop box; with this change it
reads "Agent C", and the render is pixel-identical to the version approved in #2249.

**Before** (adk.dev today):

Loop Agent column with an empty third box

**After** (this branch, local build):

Loop Agent column with Agent C in the third box


joefernandez (approved):
Thanks for the fix!

